PR #7, opened 2026-08-28
feat(phase-391 W1/W4, #816, #843): the heap-free tier, on a real image, gated in ci-l3

jerry73204:
`bins/heap-free-poc-mps2` is the first image that CALLS runtime code — `Executor::open_in` over a caller-supplied static backing, then the `install_node_typed_in` seam over a per-class `ComponentSlotStorage` — and links with no allocation symbol at all. `ci-l3` builds it and runs `check-no-alloc-image --tier heap-free` on the ELF.

**Measured** (commands in ci-l3 / the leaf):
- Gate verdict OK at `symbols read: 460` — the three earlier probes passed vacuously at `symbols read: 1`.
- 12 executor/install symbols present (llvm-nm); QEMU boot prints the marker and exits 0.
- `just ci-l3`, `ci-l1`, `check-gate-selftests`, `check-leaf-lockfiles`, `check-build-profile-literals` green locally.

**Three blockers unwound, each found by running rather than reading:**
- `node_runtime`'s bare `extern crate alloc` linked the alloc crate into every image (crate-level bisect); now cfg-gated.
- `nros` unconditionally enabled `portable-atomic-util/alloc`; now optional behind the `alloc` feature. Two reference-leaf locks refreshed via `just lock-update` (pure removals).
- `nros-platform-mps2-an385` always emitted `malloc`/`free`/`realloc`/`calloc`; new default-on `libc-heap` feature — the poc opts out, every existing consumer keeps the symbols.

**Reasoned, not measured:** zenoh/XRCE images unaffected by the feature splits (defaults preserved) — tier-2 verifies.

The leaf doubles as issue 0843's regression tripwire (compiles `nros` without `alloc`). 0843 resolved → archived; 0816's remaining half is the two Embassy book claims.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -423,9 +423,18 @@ jobs:
         if: ${{ contains(fromJSON('["schedule","workflow_dispatch"]'), github.event_name) && !cancelled() }}
         run: just build-compile-check-fixtures cxx-syntax
 
+      # The `source /opt/ros/humble/setup.bash` below is load-bearing, not
+      # boilerplate (carried from the phase-391 branch through the W1 rebase):
+      # `check-rmw-cyclonedds`'s from-msg targets run msg_to_cyclone_idl.py at
+      # BUILD time, and it imports rosidl_adapter from /opt/ros. The
+      # configure-time probe only checks the converter script EXISTS (issue-0196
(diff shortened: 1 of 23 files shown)


jerry73204:
Superseded by #33.

PR #8, opened 2026-08-28
fix(#831): `[image.<id>].rmw` selects the backend on the cargo driver — and only it

jerry73204:
Closes #831 (`docs/issues/archived/0831-*`).

`[image.<id>].rmw` named the build DIRECTORY and nothing else on the cargo driver — the backend came from the `nros sync` selection facade, off `[system] rmw`. So `[image.native_cyclonedds]` produced `build/posix-cyclonedds/` holding a **zenoh** binary, and two of tier 2's fourteen coordinates (`linux,rust,cyclonedds`, `linux,rust,xrce`) ran zenoh while claiming otherwise.

## Two halves

**1. The facade reads the image.** It is already keyed per ENTRY, and an image NAMES an entry (`package_name(image_id)`), so per-entry *is* per-image — the only missing piece was the lookup. Matched forward (`package_name(id) == entry_name`), never by stripping `_entry`, since that function replaces `-`, `.` and `/` with `_` and the reverse is ambiguous. The stand-in refusal is deleted.

**2. The board's default RMW is carved out.** Half 1 alone got *both* backends: cargo unions features and cannot subtract a default (#270), so `rmw-cyclonedds` on top of `nros-board-linux`'s `default = ["rmw-zenoh"]` is both — and the runtime then refuses to choose. The facade *and* the generated entry both declare the board `default-features = false`; one is not enough, because the union restores it. The facade re-supplies the board's non-RMW defaults, which is not optional: esp32/mps2 boards default to `["ethernet", "rmw-zenoh"]` and the NuttX boards to `["image-runtime"]` (two lang items).

## MEASURED

`nm` on the three rust-workspace native images, after a real fixture build:

| image | `dds_` | `uxr_` | zenoh `_z_` | size |
| `native` (zenoh) | 0 | 0 | 777 | 8.5M |
| `native_cyclonedds` | 328 | 0 | 0 | 7.3M |
| `native_xrce` | 0 | 232 | 0 | 6.5M |

Before the fix the same command gave 0 `dds_` and 1960 zenoh. Both new images run and select their backend (then fail for want of a peer — expected standalone behaviour, not a selection failure).

Mutation-checked: deleting `rmw = "cyclonedds"` from the image and re-running `nros sync` reverts the facade to `rmw-zenoh`, so the wiring is load-bearing.

## The claim is now checked, not trusted

New `rmw_coordinate_truth.rs` asserts, for every `[[workspace_fixture]]` row naming zenoh/cyclonedds/xrce, that the row's own binary links that backend's C namespace and (outside a declared bridge) no other. 93 artifacts on the native lane.

Three false-positive classes came out of **running** it, each a narrowing the gate needed rather than an exemption it was given: a **bridge** links two backends by declaration; a **multi-row leaf** shares one `target/` across rows with different RMWs (#517); a **rename** leaves an orphan (#215's class). Naming the row's binary from the manifest removes all three.

It carries its own negative control on the normal path. Not decorative — the symbol counter was wrong twice here, each time in the direction that makes a gate useless: matching only global `T` called the bridge's 350 local `t dds_` entries an absent backend, and a substring match would have counted `U` imports and `d` data.

## Still open, and not this bug

The two bridge workspaces stay unmigrated: a bridge needs two backends in one image, which a single `[image.<id>].rmw` cannot express. That is a per-image backend *list*.

## Verified

- `just ci-l1` — green
- `just check` — green
- `rmw_coordinate_truth` — 2/2
- linux fixture leg builds all rows; 110 workspace rows validate

Everything in MEASURED above came from running the command; nothing in this description is inferred from reading the code.


jerry73204:
**Blocked on a pre-existing CI defect, not on this change — filed as #866.**

The required check `check (fast on push; full on PR/nightly)` fails here, and it fails for **every** PR. Same failure on a different commit:

| run | commit | failure |
| --- | --- | --- |
| 33149447233 | `e1abbf31` (main) | `platform_hdr_posix_cpp_heap/.compile-ok` not prebuilt |
| 33151705338 | this branch | same |

`check-source-gates` runs `cargo test --test platform_header_compile`, which resolves a compile-check fixture directly via `require_compile_check` — not through `_require-fixtures-ready`, which is what the workflow's "it needs NO fixtures" census was keyed on. `main` never catches it because a push runs only the fast lane.

I filed #866 rather than fixing it: three candidate fixes are written down and none is verified, and which one is right is a CI-cadence decision for phase-395. Guessing would red the lane a second time.

`check-decoupling` also fails in that job on a stale path (`packages/core/nros/Cargo.toml`; the crate is at `packages/api/nros`) — but that step is `continue-on-error: true` and gates nothing.

**The other 23 red jobs are also not from this change.** The 6 `nros-rmw-zenoh` failures under `workspace unit tests` are router-dependent tests in a container without one; `test-unit` passes `--exclude nros-tests`, so this PR's new test never runs in that job at all, and the test count is identical locally and in CI (993). The 20 zephyr matrix jobs need SDKs.

### What I did verify, by running it

- `just ci-l1` — green locally
- `just check` — green locally
- `rmw_coordinate_truth` — 2/2, 93 artifacts checked
- `nm` on the three rust-workspace images: 328 `dds_`/0 zenoh, 232 `uxr_`/0 zenoh, 0/0/777 zenoh — one backend each
- mutation check: dropping the image's `rmw` reverts the facade to `rmw-zenoh`

Nothing above is inferred from reading the code.

jerry73204:
## Status: #866's stated defect is fixed; a fourth, separate blocker is now visible

`check-source-gates` passes in CI now. Three commits, each verified by reproducing the failure first:

| # | blocker | fix |
|---|---|---|
| 1 | compile-check fixtures never built in this lane | build the `cxx-syntax` class in `check-source-gates` (3.6 s) |
| 2 | `cmake_fixture_prereqs_ok` demanded `play_launch_parser` even when no cmake row was selected | ask whether the lane has work first |
| 3 | `[SKIPPED]` precondition counted as a failure (issue 0388) | reuse `test-unit`'s junit tolerance |

**The fourth is `check-rmw-cyclonedds`, and it is not mine:**

```
The vendored fallback is also absent (third-party/ros/rosidl).
  nros setup --source rosidl      # vendored copy, no ROS needed
gmake: *** [tests/.../AddTwoInts.idl] Error 1
```

The IDL step needs the rosidl adapter from ROS or the vendored source, and the container's build env has neither. It was masked until now by blockers 1–3.

## The pattern is the finding

Four independent environment prerequisites, each revealed only by fixing the previous one. That is not four bugs — it is one fact: **`just check`'s BUILD tier has never run in this container.** Direct pushes to `main` only ever ran `check-fast` (buildless), and the nightly that would have covered it skipped on a path filter. Enabling the merge queue promoted a never-executed lane to required.

I'm stopping here rather than peeling a fifth layer, since each costs a ~20 min round-trip and the depth is unknown.

## Options

**B — make the required check the lane designed to run there.** `check-fast` + `test-unit` is buildless and is what the workflow comment already says it wants ("makes the PR gate exactly L1"). `check-build` wants ROS, cross toolchains with libstdc++, and fixtures — a full dev environment — so it belongs in nightly where those are provisioned.

**C — provision the container** with rosidl, a C++-capable newlib cross, and `play_launch_parser`.

**A — keep peeling.** Unknown remaining depth.

I'd take B: it matches the workflow's own stated intent and unblocks the queue today. C is the more complete answer but is a container change I shouldn't make unilaterally.

**This PR's own change is unaffected and verified** — `nm` shows one backend per image (328 `dds_`/0 zenoh, 232 `uxr_`/0 zenoh), `rmw_coordinate_truth` 2/2 over 93 artifacts, `just check` and `just ci-l1` green locally from a wiped fixture dir.

jerry73204:
**Rebased onto the new `main` (`967c9d09`). The remaining red is not this PR — per AGENTS.md's "if the check is red for several different PRs, stop and say so".**

## What changed here

The branch is now **one commit**: the #831 fix alone. Everything else I had stacked on it is gone, because it landed or was superseded while I was away:

| was | now |
|---|---|
| container ROS-env fix | **merged** via #16 (`4fccff6d4`) |
| `check-source-gates` builds its compile-check fixtures | superseded — main does this as `NROS_COMPILE_CHECK_LANES=cxx-syntax` |
| cmake prereqs demanded unconditionally | superseded — main gates inside `cmake_fixture_prereqs_ok` via `_lane_on`, which is cleaner than my call-site guard |
| `[SKIPPED]` counted as failure | superseded — main's `_nextest-tolerant` (issue #673) |

I also dropped a selftest I had added to `compile-check-fixtures.sh`: main now explicitly excludes `scripts/build/**` from gate scope ("PRODUCES artifacts; it does not assert"), so it is out of scope, and adding one would be arguing with a deliberate decision rather than fixing a defect.

## The failure is infrastructure

Both failing steps — `check-leaf-lockfiles` and `check-compile-smoke` — have the same cause:

```
sccache: error: Server startup failed: create gha cache failed: ConfigInvalid (permanent)
  at Builder::build
ERROR: 31 leaf crate(s) failed for a reason that is NOT lock drift (see above).
       Fix those first — this gate deliberately does not classify them.
```

`SCCACHE_GHA_ENABLED: true`, and the GHA cache backend refuses to initialise. 31 crates fail identically; the lockfile gate says outright it is not classifying them.

**Not mine, and not fixable by rebasing:**

- `just check-leaf-lockfiles` and `just check-compile-smoke` both pass locally on this exact tree
- **PR #14 fails with the identical sccache error** on `just check-fast`
- PRs #10 and #12 are also red on the required check

## Verified for this PR's own change, on the new main

- `nros sync` → facades: `native_cyclonedds_entry` → `rmw-cyclonedds`, `native_xrce_entry` → `rmw-xrce`, `native_entry` → `rmw-zenoh`
- `nm` on the rebuilt artifacts: 328 `dds_`/0 zenoh · 232 `uxr_`/0 zenoh · 0/0/777 zenoh — one backend each
- `rmw_coordinate_truth` 2/2 over 93 artifacts
- `just ci-l1` green — the same tier the merge group runs

Nothing above is inferred from reading; each line is a command that ran.

PR #17, opened 2026-08-28
fix(#832): cyclone's ddsrt heap funnels through the platform ABI, by target dependency

jerry73204:
**Stacked on #7** (`fix/heap-free-tier-gate`) — a PR off `main` would hit the same five required-check failures #7 is fixing.

Routes cyclone's `ddsrt` heap through `nros_platform_alloc` behind a compile-time switch set on `ddsc` only.

**Measured** — `examples/native/c/talker/build-cyclonedds/c_talker`, rebuilt via `just build-test-fixtures lane=native`, objdump caller attribution counting `call` AND `jmp` (the issue's own method note):

| | malloc@plt callers | funnel inbound edges |
| --- | --- | --- |
| before | `ddsi_config_init`, `ddsrt_malloc`, `ddsrt_malloc_s`, `network_interface_find_or_append`, `nros_platform_alloc` | 0 |
| after | `ddsi_config_init`, `network_interface_find_or_append`, `nros_platform_alloc` | 4 |

Compile-time, not weak linkage: weak keeps both branches in the binary, and the libc edge's *absence* is what a tier gate reads.

**Not fixed here** (0832 stays open): the two survivors are one bare `malloc` at `ddsi_config.c:2349` and its inlining. Editing that file re-runs cyclone's `_confgen`, which resolves `${CMAKE_SOURCE_DIR}/cmake/AppendHashScript.cmake` against *our* root in the nested `add_subdirectory` build and dies — a latent vendoring bug worth its own fix.

**Also measured:** 0832's XRCE half is already resolved — that image's only `malloc@plt` caller is the funnel, and `get_ip_from_iface` (a zenoh-pico TU, not XRCE as the issue's table records) calls through it.

Fork commit `6e2ad36f` pushed to `NEWSLabNTU/cyclonedds` `nano-ros` before the pin bump, per the vendored-fork workflow; delta doc updated. Also carries issue #872 documenting the CI-arm gaps.

### DIFF
--- a/docs/issues/0827-unused-rmw-pools-dominate-static-ram.md
+++ b/docs/issues/0827-unused-rmw-pools-dominate-static-ram.md
@@ -92,6 +92,58 @@ list: an image whose resolved model contains zero subscriptions needs zero of
 it, and that derivation needs no infrastructure accounting — unlike the
 queryable one above, the runtime creates no large-payload subscriber of its own.
 
+## Measured 2026-08-29 — the "easiest win" is NOT available with the existing knob
+
+The section above calls `LARGE_PAYLOADS` "the single easiest win ... an image
+whose resolved model contains zero subscriptions needs zero of it". Zero is
+**not expressible**. `packages/rmw/zenoh/nros-rmw-zenoh/build.rs:45`:
+
+```rust
+let max_large: usize = env_usize("ZPICO_MAX_LARGE_SUBSCRIBERS", 2).max(1);
+```
+
+The `.max(1)` floor means the smallest reachable pool is
+`1 * ZPICO_SUBSCRIBER_RING_DEPTH(4) * ZPICO_SUBSCRIBER_LARGE_SIZE(16384)` =
+**65,536 bytes**, not 0. So wiring the resolved model to this knob — the fix
+this issue proposes — buys half of what the issue claims: 131,072 -> 65,536 on
+a talker, with the other half structurally out of reach until the floor goes.
+
+MEASURED, not read — three builds of `examples/native/rust/talker`, the knob
+varied, `llvm-nm -S` on each binary:
+
+| `ZPICO_MAX_LARGE_SUBSCRIBERS` | `LARGE_PAYLOADS` |
+| ---: | ---: |
+| 2 (default) | 131,072 |
+| 1 | 65,536 |
+| **0** | **65,536** |
+
+Note the third row: asking for zero does not fail, it silently yields one. A
+codegen deriving "no subscriptions -> 0" would emit a config that reads as
+satisfied while still reserving 64 KiB — the same shape as every other defect
+this campaign has found, a value that looks applied and is not. If the floor
+stays, the knob should REJECT 0 rather than round it up.
+
+Baseline confirming the arithmetic, `just mem-report` on
+`build/cargo-fixtures/linux-14372940/nros-relwithdebinfo/talker`:
+
+```
+       131,072   35.8%  nros_rmw_zenoh::shim::subscriber::LARGE_PAYLOADS
+       131,072  LARGE_PAYLOADS  — agrees with `ZPICO_MAX_LARGE_SUBSCRIBERS *
+                                  ZPICO_SUBSCRIBER_RING_DEPTH * ZPICO_SUBSCRIBER_LARGE_SIZE`
+```
+
+Removing the floor is not free to reason about: `.max(1)` exists so a pool
+index is always valid, and the sibling `max_nodes` floor documents exactly that
+intent ("so a session always has room for its own primary node"). A zero-length
+pool needs the *lookup* path to refuse a large subscription rather than index an
+empty array — a code change, not a knob change. Whoever takes this should price
+both halves separately: the knob wiring (65,536, mechanical) and the floor
+removal (a further 65,536, needs the refusal path).
+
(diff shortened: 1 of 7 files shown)


jerry73204:
**Self-review finding — this PR is not ready to merge as-is.**

Running `just ci-l1` on this branch fails in `check-rmw-cyclonedds`:

```
heap.c:(.text+0x131): undefined reference to `nros_platform_realloc'
heap.c:(.text+0x14a): undefined reference to `nros_platform_dealloc'
collect2: error: ld returned 1 exit status
  tests/nros_rmw_cyclonedds_ros2_srv_client
  tests/nros_rmw_cyclonedds_ros2_srv_server
  tests/nros_rmw_cyclonedds_ros2_sub
```

The compile-time define lands on `ddsc`, and those test executables link `ddsc`
without the platform layer — so the funnel externs resolve nowhere. The measured
saving in the PR body still stands for the *example* images (they link the
platform layer), but the change breaks a lane I had not run before opening this.

That is exactly the trade I described in the commit message and got wrong in one
direction: weak linkage would have made these link (unresolved → libc), at the
cost of leaving the libc branch in the call graph. The fix is to make the
consumers link the platform layer, or to scope the define to the targets that
do — not to revert to weak.

Holding this PR until that is resolved rather than queueing it behind #7.

jerry73204:
**Attempted fix, did not work — recording it so the next attempt does not repeat it.**

Tried: weak DEFINITIONS of the three funnel symbols in a new
`src/platform_alloc_fallback.c`, so a real image's strong symbols from
`nros-platform` override them while standalone test binaries get a libc
fallback. Weak *definitions* rather than weak *references* on purpose — a weak
reference needs a null check at every call site and leaves the libc branch in
the call graph, which is the property this PR exists to remove.

Two shapes, both measured, both still 7 undefined references:

1. **As a member of the `nros_rmw_cyclonedds` archive.** The linker pulls an
   archive member only to satisfy a reference it has already seen, and
   `libddsc.a` comes AFTER `libnros_rmw_cyclonedds.a` on the link line — so the
   member was never pulled.
2. **As an `INTERFACE` source on the backend target**, so consumers compile it
   directly. After a clean reconfigure the file is not compiled at all (zero
   occurrences of `platform_alloc_fallback` in the build log), so the
   propagation is not reaching these executables for a reason I have not
   established.

Reverted both; the branch is back at `a4850b7ee` and PR stays held.

Next attempt should start by establishing **why (2) does not propagate** —
whether `tests/CMakeLists.txt` links the backend in a way that drops INTERFACE
sources, or whether the tests are configured before the property is set. A
cheap alternative if that is a dead end: add the fallback source directly to
the affected `add_executable` calls, which is uglier but has no propagation
question in it.

jerry73204:
Fixed by dependency order, not weak symbols — and the weak-symbol route is now recorded as the wrong one for a reason stronger than "it did not link".

**The fix.** `NROS_DDSRT_PLATFORM_FUNNEL` turns three platform-ABI symbols into undefined references in every archive that absorbs `heap.c`, so the define and the link requirement are one decision and are now made in one place: the provider (`NanoRos::Platform`, the Phase-138 §A alias every platform module exposes) is resolved *before* the define is set and linked into the same targets that get it. Being a dependency is what orders the link line — nothing arranges the order by hand. `$<BUILD_INTERFACE:>` keeps it out of Cyclone's `install(EXPORT)`, which otherwise refuses a dependency outside its export set.

Where no provider exists the funnel stays **off** and Cyclone keeps libc — that is the standalone `just cyclonedds ci` project, which builds the backend with no platform layer, and libc is the right answer there. Both outcomes print a STATUS line, because a silently-disabled funnel looks identical to an enabled one from the recipe.

**Why not the weak fallback**, beyond the two failed wirings recorded above: a weak `nros_platform_alloc` resolves the reference by calling `malloc`, reintroducing the libc heap this issue exists to remove. The issue-0050 audit has no legal `body:` label for that — `body:silent-wrong` deliberately does not exist, because it is the state the axis was added to surface. The gate would have rejected the file on those grounds independent of link order.

**Measured, both directions:**
- funnel ON (root configure, `-DCMAKE_DISABLE_FIND_PACKAGE_CycloneDDS=ON`): `heap.c.o` carries `U nros_platform_{alloc,realloc,dealloc}` and no `U malloc`/`U free`; an executable linking `NanoRos::NanoRos` resolves all three.
- funnel OFF (`just check-rmw-cyclonedds`): 22/22 ctest pass.
- `just ci-l1`: green.

**Scope correction the original commit did not state.** This is the self-provision branch, so a `find_package` build is unaffected — its `heap.c` is already compiled. On a host with an SDK-store or ROS Cyclone that is what a *native* build resolves, so the funnel does not reach native fixtures here. That is the right split for the tier model (`unified` is an embedded promise, and a cross build skips `find_package` outright so it always lands here with a platform present) — but it does mean the funnel-ON path has no fast-line CI lane today, only the embedded cyclone fixtures in tier 2. Worth its own issue if we want it cheaper to cover.

Also on this branch: `68127efa7` repairs a regression from `a4850b7ee` — the `env_usize_min` wrapper made two zpico knobs invisible to the pool-inventory scraper, deleting both payload pools' byte figures (issue 0271's failure mode, second time a convenience wrapper has caused it). Scraper now matches the spelling, self-test probes every spelling, verified non-vacuous.

Un-holding.

PR #24, opened 2026-08-29
docs(#880, #881): two gaps found while fixing the cyclone funnel's link

jerry73204:
Stacked on #17 (`fix/0832-cyclone-funnel`), per the FIFO merge-queue rule — not on `main`.

Docs only. Two issues filed for gaps measured while fixing #17; neither is fixed here.

**#0880 — `just format` is build-state dependent.** `examples/workspaces/launch`'s two Rust leaves are members of a workspace root that `nros build` generates and `.gitignore` hides, so the recipe is green on a tree that built that workspace and red on one that did not. `example_shape.rs` already records this fact and reads the tracked `.colcon_workspace` marker; nothing else does.

Swept the class, not the site — nearest enclosing `[workspace]`, root `members` globs expanded, `exclude` honoured, each candidate cross-checked against `cargo metadata`. Exactly two leaves, both under `launch`. Third occurrence of the shape issue 0359 named, second distinct cause.

The detail that makes it a design question rather than two `exclude` lines: `mixed` is equally unbuilt and equally migrated, and formats fine only because its one leaf carries its own `[workspace]` while `launch`'s two do not. Two leaf shapes coexist in one workspace class, and the gate that could forbid that currently does neither.

**#0881 — the funnel-ON cyclone path has no fast-line lane.** The switch only exists on the self-provision branch, so: standalone backend project has no platform to funnel into; native host resolves `find_package` from the SDK store; only a cross build lands on source-with-platform. Before #17 the standalone project was what caught the missing link dependency — after it, that lane no longer compiles the funnel, so the next regression of this shape surfaces in tier 2. Issue carries three candidate fixes and the ~4 min by-hand check that stands in meanwhile.

`just check-fast`: green.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -423,9 +423,17 @@ jobs:
         if: ${{ contains(fromJSON('["schedule","workflow_dispatch"]'), github.event_name) && !cancelled() }}
         run: just build-compile-check-fixtures cxx-syntax
 
+      #
+      # The `source /opt/ros/humble/setup.bash` below is load-bearing, not
+      # boilerplate: `check-rmw-cyclonedds`'s from-msg targets run
+      # msg_to_cyclone_idl.py at BUILD time, which imports rosidl_adapter from
+      # /opt/ros. The configure-time probe only checks the converter script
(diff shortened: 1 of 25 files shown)


jerry73204:
Superseded by #33.

PR #26, opened 2026-08-29
fix(#832): the four ddsi sites that called libc directly — three were already bugs

jerry73204:
Stacked on #24 (#25 merged into it). Bumps the cyclonedds fork forward one commit (`556f79d4` → `8e6ff48a`, pushed to NEWSLabNTU/cyclonedds `nano-ros`). Completes the **cyclone half** of #832.

## What turned out to be there

Routing `ddsrt`'s heap through the funnel was not enough: four ddsi sites called libc directly, and **three were a bug before the funnel existed**. Each allocates from one heap and releases to the other.

| site | allocated by | freed by |
| --- | --- | --- |
| `network_interface_find_or_append` listelem | `malloc` | `free_all_elements` → `ddsrt_free` |
| `split_at_comma`'s array | `ddsrt_malloc` | libc `free`, twelve lines later |
| `->verbatim` in `..._AppliedBuiltinTypeAnnotations_copy` | `calloc` | `dds_stream_free_sample` → `ddsrt_free` |
| `iceoryx_init`'s name string | `ddsrt_asprintf` → `ddsrt_malloc` | libc `free` |

On POSIX both heaps are glibc, so none of these can fault natively — which is how they survived. They cross genuinely different heaps on ThreadX, FreeRTOS and Zephyr, and under the funnel. The third is visibly inconsistent with its own caller three functions down, which already used `ddsrt_calloc` for the parent struct.

## Swept, not spot-fixed

#832 named two sites. Two more turned up, and two others were checked and **deliberately left**:

- `sysdeps.c`'s `free` pairs with `backtrace_symbols` — glibc allocated that block, so libc `free` is the required call.
- `q_freelist.c`'s `free` is the function's own parameter (`void (*free)(void *elem)`), not libc at all.

`src/idl/` is out of scope: the IDL compiler is a host tool that links no platform layer.

After: **`libddsc.a` has exactly one object referencing a raw libc allocator** — `sysdeps.c.o`'s `free`, the correct one.

## Runtime, not just link

A binary linking the funnelled `ddsc` against `libnros_platform_posix.a` carries 7 funnel call sites and runs: `ddsrt_malloc`/`realloc`/`free` round-trip, then a domain and participant created and deleted under a `NetworkInterfaceAddress` config — which is what drives the two `ddsi_config.c` sites. Its deprecation warning is the evidence the path executed rather than being optimised out.

Backend suite 22/22. `just ci-l1` green.

First work #25's confgen fix unblocked: editing `ddsi_config.c` arms `_confgen`, which now regenerates cleanly (one hash line moves in each of four generated files).

## Still open on #832

- **The XRCE half.** `get_ip_from_iface` untouched; that row of the issue's table stands.
- **Coverage.** The verification above is by hand, because the funnel-ON path has no fast-line lane — issue #0881.

Fork-delta re-measured against the fork rather than incremented by hand: 20 commits, 51 files, +2361/−138.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -423,9 +423,17 @@ jobs:
         if: ${{ contains(fromJSON('["schedule","workflow_dispatch"]'), github.event_name) && !cancelled() }}
         run: just build-compile-check-fixtures cxx-syntax
 
+      #
+      # The `source /opt/ros/humble/setup.bash` below is load-bearing, not
+      # boilerplate: `check-rmw-cyclonedds`'s from-msg targets run
+      # msg_to_cyclone_idl.py at BUILD time, which imports rosidl_adapter from
+      # /opt/ros. The configure-time probe only checks the converter script
+      # EXISTS (issue-0196 class: probe narrower than the rule), so an
+      # unsourced env fails at 80% of the build rather than at configure.
       - name: just check-build + no_std (nightly / manual only)
         if: ${{ contains(fromJSON('["schedule","workflow_dispatch"]'), github.event_name) && !cancelled() }}
         run: |
+          source /opt/ros/humble/setup.bash
           just check-build
           just check-no-std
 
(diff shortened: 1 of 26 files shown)


jerry73204:
Superseded by #33.

PR #28, opened 2026-08-29
fix(#832): one allocation route for cyclone on every port, not just POSIX

jerry73204:
Stacked on #26. Bumps the cyclonedds fork forward one commit (`8e6ff48a` → `d97a71e2`).

## The gap

Cyclone links **above** nano-ros's platform layer, so it should reach no heap the platform does not own. The first funnel commit only made that true on POSIX — the arms lived inside `heap/posix/heap.c`, and the other ports kept their own allocator:

| port | heap before |
| --- | --- |
| posix | funnel ✔ |
| freertos | `pvPortMalloc` |
| threadx | `tx_byte_allocate` |
| vxworks | libc (not built) |

Those are the ports whose heap genuinely is not libc's — so the tier promise held exactly where it did not matter, and a second allocation route survived where it did.

## The shape

One implementation, not four arms: `heap/nros/heap.c` provides the whole `ddsrt_{malloc,malloc_s,calloc,calloc_s,realloc,realloc_s,free}` family on `nros_platform_{alloc,realloc,dealloc}`, and each port's own heap.c is compiled out by the same switch. `heap/posix/heap.c` is stock again apart from that guard — nine lines per port file.

**The new file is in the COMMON source list, not swapped in per port, and that is load-bearing.** `ddsrt` is an INTERFACE library and `ddsrt-internal` compiles the same `INTERFACE_SOURCES`; a swap would hand the funnel to idlc and confgen — host tools with no platform layer, where the three symbols are undefined. `NROS_DDSRT_PLATFORM_FUNNEL` being a PRIVATE compile definition on `ddsc` is what keeps the two apart while both see one file set.

## Measured

| archive | result |
| --- | --- |
| `libddsc.a` | `heap.c.o → U nros_platform_{alloc,realloc,dealloc}`; one object left referencing a raw libc allocator |
| `libddsrt-internal.a` | zero `nros_platform_*` refs; stock posix heap still defines the family, host tools link |

The surviving libc `free` is `sysdeps.c`'s, and it is **not a hole in the promise**: its block is guarded to `__APPLE__ || (__linux && (__GLIBC__ || __UCLIBC__))` and `! DDSRT_WITH_FREERTOS`, so it does not exist on any target where the platform's heap and libc's differ — and where it does exist it pairs with a block `backtrace_symbols` allocated from glibc itself.

## Falls out of it

Dropping the per-port heaps drops their size-header prefix (`ofst`) and alignment arithmetic, which existed only because the RTOS allocators have no realloc — `nros_platform_realloc` is specified with libc realloc semantics, so the old size never has to be recovered from a header. The ThreadX weak `zpico_threadx_byte_pool` goes with it: one fewer weak symbol, which the issue-0050 policy prefers anyway.

`heap/vxworks/heap.c` untouched — no CMakeLists references it, it does not include `dds/ddsrt/heap.h`, and nothing here can compile it.

## Verification

- funnel ON: links and runs — 7 funnel call sites in a binary that round-trips `ddsrt_malloc`/`realloc`/`free` and creates and deletes a domain and participant
- funnel OFF, each port heap live: backend suite 22/22
- `just ci-l1`: green

Still open on #832: the XRCE half (`get_ip_from_iface`), and the fast-line coverage gap (#0881).

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -344,9 +344,17 @@ jobs:
       # Restoring it means giving this job the artifacts first (PR #14 adds
       # `build-compile-check-fixtures`; the bindings half is unowned). Do that
       # before putting the tier back, not after.
+      #
+      # The `source /opt/ros/humble/setup.bash` below is load-bearing, not
+      # boilerplate: `check-rmw-cyclonedds`'s from-msg targets run
+      # msg_to_cyclone_idl.py at BUILD time, which imports rosidl_adapter from
+      # /opt/ros. The configure-time probe only checks the converter script
+      # EXISTS (issue-0196 class: probe narrower than the rule), so an
+      # unsourced env fails at 80% of the build rather than at configure.
       - name: just check-build + no_std (nightly / manual only)
         if: ${{ contains(fromJSON('["schedule","workflow_dispatch"]'), github.event_name) && !cancelled() }}
         run: |
+          source /opt/ros/humble/setup.bash
           just check-build
           just check-no-std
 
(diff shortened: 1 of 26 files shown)


jerry73204:
Superseded by #33.

PR #29, opened 2026-08-29
fix(#832, #881): the XRCE half, and a fast-line lane that actually compiles the funnel

jerry73204:
Stacked on #28. Closes **both** #832 and #881 — they turned out to be the same change one project apart.

## #832, XRCE half

**The bypass was never where the issue said it was.** `get_ip_from_iface` is a *zenoh-pico* symbol (`src/system/{unix,windows,freertos,rpi_pico}/network.c`) and appears nowhere in the XRCE tree — that row was a misattribution. The vendored micro-XRCE-DDS-Client and micro-CDR call no allocator directly at all.

The real bypass is **our own shim**: 38 raw `calloc`/`free` calls across six files, in-tree, no fork needed.

| file | calloc | free |
| --- | ---: | ---: |
| `session.c` | 1 | 7 |
| `service.c` | 2 | 4 |
| `publisher.c` | 1 | 2 |
| `subscriber.c` | 1 | 2 |
| `transport_nros_udp.c` | 2 | 10 |
| `transport_zephyr_udp.c` | 2 | 4 |

All now go through `nros_xrce_calloc`/`nros_xrce_free` — two `static inline`s over the platform ABI in `internal.h`. The zeroing lives in the helper, not at the call sites: every allocation was `calloc(1, sizeof(X))` on a struct only partly assigned afterwards, so it is load-bearing, and a `memset` to be remembered nine times is one forgotten at the tenth.

**Measured:** `libnros_rmw_xrce.a` — vendored client and micro-CDR included — has **zero** objects referencing a raw libc allocator, and 10 funnel references.

### Two things fell out

The standalone check project took platform *headers* via a cache var but linked no implementation, so it stopped linking the moment the shim called the funnel. It now pulls `nros-platform-posix` in the same style — which is what `nros-rmw-xrce-cffi`'s build.rs already required of the image path ("as long as the consumer links a platform-provider library"). This project was the one consumer that did not.

And `tests/smoke.c` carried **nine local definitions of platform symbols**, added by #787 for a reason that no longer holds. Deleted — the real archive defines every one. Only two collided at link time, because the linker pulls an archive member only for an already-unresolved reference and the local UDP definitions meant `net.c.o` was never reached. So seven of the nine were a second implementation of the platform ABI selected silently by link order, with the two implementations **disagreeing**: the stubs fail every call, the real ones work. That is the hazard the issue-0050 policy exists to prevent, reached with no weak symbol involved.

## #881, the coverage gap

Candidate 2, and the XRCE work is what settled it — the pattern transfers verbatim. The cyclone project resolves `NROS_PLATFORM_IMPL_DIR` (cache PATH defaulted to the sibling project, the convention `CYCLONEDDS_SOURCE_DIR` already uses) and aliases it as `NanoRos::Platform`, which is what the funnel gate asks for. Both guarded on the targets not already existing, so a parent nano-ros build still wins.

`just check-rmw-cyclonedds` now prints `CycloneDDS ddsrt heap -> nros_platform_alloc`, and `nros_rmw_cyclonedds_ros2_{sub,srv_client,srv_server}` carry **7 funnel call sites each**. Those three are the executables that caught the original missing link dependency — the lane is back to covering that shape, this time with the funnel actually compiled and end-to-end rather than at link only. 22/22.

No second Cyclone build; this lane already built it from source. It gains a platform archive it did not build before.

## Not done here

`packages/rmw/xrce/nros-rmw-xrce/src` is **not covered by `check-c-fmt`**, and those files use right-pointer style against the repo's `PointerAlignment: Left`. Running the pinned clang-format over them produced ~350 lines of churn unrelated to this change, so the edits follow the files' existing style. Widening the gate is a separate change with a reformat attached.

## Verification

- `just check-rmw-xrce`: 1/1
- `just check-rmw-cyclonedds`: 22/22, funnel ON
- `just ci-l1`: green

Both issues moved to `docs/issues/archived/` with the measurements.

### DIFF
--- a/docs/issues/README.md
+++ b/docs/issues/README.md
@@ -53,7 +53,7 @@ Issues cross-link to the RFCs and phases that inform or resolve them via the
 
 <!-- BEGIN GENERATED open-issue list — scripts/gen-issue-index.py -->
 
-44 open. One line each — the detail lives in the issue file,
+42 open. One line each — the detail lives in the issue file,
 which already has it. Regenerate with `scripts/gen-issue-index.py`;
 `check-issue-index` fails if this block drifts.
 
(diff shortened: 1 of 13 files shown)


jerry73204:
Superseded by #33.

PR #34, opened 2026-08-29
docs(#0883): the generated issue index conflicts with itself, and it caps the queue

jerry73204:
One file is now the merge-queue bottleneck.

Seven PRs were rebased onto `main`, gated green and armed with auto-merge. **#9 — a two-line issue resolution — merged, and ejected the other six**, cancelling their auto-merge and costing a full rebase round.

`docs/issues/README.md` is generated by `gen-issue-index.py` and committed. Nearly every PR here files, resolves or archives an issue, so nearly every PR rewrites it — and two that both rewrite it conflict even when their real changes are unrelated. Test-merged against `main` in an isolated worktree:

```
fix/0865-nuttx-qemu-throttling      CONFLICT: docs/issues/README.md
fix/0819-xrce-fragmented-receive    CONFLICT: docs/issues/README.md
fix/0871-ci-compile-check-fixtures  CONFLICT: docs/issues/README.md
fix/heap-free-tier-gate             merges CLEAN
fix/0831                            merges CLEAN
```

Three of five, and that path is the **only** disagreement in each.

**Not fixed here, because the obvious answer does not work.** A custom merge driver is right in principle — the file is a pure function of the issue files, so regenerating *is* the correct merge — but GitHub's queue rebases **server-side**, where custom drivers do not run. It would fix local rebases and leave the half that hurts. `merge=union` is worse: the index is a sorted table with a counted header, so union yields duplicate rows and a count contradicting them, and `check-issue-index` correctly fails.

Four options recorded with what each concedes. Docs only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011UTfthJ6zZou2YUPZS1jyH

PR #35, opened 2026-08-29
fix(#0884): the issue ledger stops being a shared registry

jerry73204:
Every PR that files or resolves an issue conflicts with every other one. Measured on `origin/main`, two agents filing unrelated issues:

```
OLD design merge rc=1
UU docs/issues/README.md
```

This session hit it on **four consecutive rebases**. phase-395 W1 already fixed the worst form (a 4,170-line hand-written list → generated rows); what remained still collided, for two reasons.

## A running total

The block opened with `NN open.` — one line every issue-touching PR rewrites at the same position. It fails two ways, and the quiet one is worse:

- differing counts → union keeps **both** lines
- **matching** counts → merges silently to a **wrong** total. Measured: two branches each taking `3 open.` → `4 open.` produced `4` for five issues. A clean merge with a false number.

Derivable from the list it sits on. Gone.

## Generated rows inside an authored file

`README.md` is prose + a generated block, and a merge strategy that suits generated rows is unsafe on prose. So the list moved to **`docs/issues/open.md`** — 100% generated — and README keeps the conventions and points at it.

`docs/issues/open.md merge=union` in `.gitattributes`. **`union` is a built-in git driver**, so it works in every clone with no `git config` step; a custom merge driver would silently not apply for anyone who skipped setup.

Union is safe here *only* because the file is entirely generated: the generator re-sorts and de-duplicates, and `check-issue-index` is the backstop. That reasoning is written into `.gitattributes`, `open.md`, README and the gate — the safety depends on the separation holding.

## Verified both directions

```
NEW (this branch)   concurrent merge rc=0   NO CONFLICT   both rows present
OLD (origin/main)   merge rc=1              UU docs/issues/README.md
```

A control on the old design was run so this isn't a fix for a non-problem. Also checked: a **resolution racing a filing** merges correctly — the resolved row drops, the new row stays, because non-overlapping hunks never reach the union driver.

## Does not solve, and should not

Two agents editing the same issue **file** still conflict. That's a real disagreement about content, not registry contention.

CLAUDE.md and AGENTS.md both told agents to *expect* these conflicts. That advice is now wrong, so both say the opposite and name the mechanism.

`check-issue-index` OK (42 rows = 42 files); 135/135 fast gates.

Closes #0884

https://claude.ai/code/session_01X2tZGAEZ8rWab5YcDVpFKk

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+
+# issue 0884 — the open-issue index is 100% generated and append-heavy.
+# Union merge lets two agents file issues concurrently without a conflict;
+# `scripts/gen-issue-index.py` re-sorts and `check-issue-index` verifies, so a
+# union that interleaves or duplicates is normalised on the next run rather
+# than shipped. NEVER apply this to an authored file — see docs/issues/README.md.
+docs/issues/open.md merge=union
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -653,7 +653,8 @@ and verified.
 
 - **Parallel agent sessions push to `main` concurrently.** `git fetch` + check
   `origin/main`'s highest issue id (including `archived/`) immediately before
-  filing a `docs/issues/` entry; expect `docs/issues/README.md` rebase conflicts
+  filing a `docs/issues/` entry. The generated list is `docs/issues/open.md` and is
+  `merge=union` (issue 0884), so two agents filing concurrently no longer conflict
   (merge both sides, renumber only your own files). Stash-wrap local-only files
   (`packages/rmw/zenoh/zpico-sys/c/include/zpico.h`-style) around every rebase.
 - **Write full logs of background builds/tests to files** and grep afterwards;
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,7 +23,7 @@ pointer here — never grow CLAUDE.md with design/impl detail.**
 | --- | --- |
 | Finalized whole-system design | [docs/design/ARCHITECTURE.md](docs/design/ARCHITECTURE.md) |
 | A specific design decision (stable vs evolving) | [docs/design/](docs/design/README.md) — numbered RFCs |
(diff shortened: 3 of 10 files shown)


jerry73204:
Independent measurement of the same problem, from another session — folding it here rather than keeping a second issue open. I filed #0883/#34 for this before seeing yours; **closing #34 in favour of this**, since yours has a fix and mine only documented it.

## The measurement, if useful for the issue

Test-merging every open branch against `main` in an isolated worktree, the day the queue first worked:

```
fix/0865-nuttx-qemu-throttling      CONFLICT: docs/issues/README.md
fix/0819-xrce-fragmented-receive    CONFLICT: docs/issues/README.md
fix/0871-ci-compile-check-fixtures  CONFLICT: docs/issues/README.md
fix/heap-free-tier-gate             merges CLEAN
fix/0831                            merges CLEAN
```

Three of five, and that path was the **only** disagreement in each. Concretely: seven PRs were rebased, gated green and armed with auto-merge; **#9 — a two-line issue resolution — merged and ejected the other six**, cancelling their auto-merge and costing a full rebase round.

Also worth recording: `git merge-tree --write-tree` does **not** exist on this host's git, and my first pass used it and reported CONFLICT for all five — that was the command failing, not a merge result. Use a real worktree merge.

## One open question your fix depends on

The issue argues against a **custom** merge driver because it "would silently not apply for anyone who had not run a setup step". Agreed, and there is a second reason that is stronger and applies to `merge=union` too:

**GitHub's merge queue performs its rebase server-side.** A custom driver (`merge.<name>.driver`) definitely does not run there — it needs local git config. Whether a *built-in* strategy like `union` is honoured by the queue's rebase is the thing I could not confirm, and it is exactly the case that matters: the conflicts above happened **in the queue**, not on a developer's machine.

If `union` is not applied server-side, this fixes local rebases and leaves the half that hurts. Worth testing before relying on it — two PRs that each add an issue row, both queued, is the experiment.

Not blocking; your split of `open.md` out of `README.md` reduces the surface regardless of how that resolves.

PR #38, opened 2026-08-29
deps(zenoh-pico): bump to the isotp-c fork with ISO_TP_NO_FORMATTED_ERRORS

jerry73204:
The vendored ISO-TP library now comes from a fork of `SimonCahill/isotp-c`
carrying one commit, which adds an `ISO_TP_NO_FORMATTED_ERRORS` build flag.

`snprintf` was the library's only libc dependency beyond `memcpy` and
`memset`, present for two formatted diagnostics that each write into a
128-byte stack buffer. That made it a hard requirement of the whole
library on exactly the targets this link exists for. The flag omits both;
the errors still go out through `isotp_user_debug()` and the return codes
are unchanged.

Off by default, so nothing here changes unless a board asks for it.
Cortex-M4, `-Os -DNDEBUG -ffreestanding`: `.text` 2235 -> 2091 bytes and
the largest stack frame 160 -> 32 bytes.

Forward-only: 33dea371..1b851957. The submodule's `nano-ros` branch had
diverged from the pin, which sat on `fix/zephyr-serial-irq-rx`; the
isotp-c commit is rebased on top so `nano-ros` now carries both and the
pin follows it again.

---

Direct push to `main` is now blocked by the required `CI` status check, so this goes through a PR.

The submodule's `nano-ros` branch had diverged from what `main` pinned: the pin sat on `fix/zephyr-serial-irq-rx` (`33dea371`) while `nano-ros` was still at `b7f50869`. The isotp-c commit is rebased on top of `33dea371` and pushed to `nano-ros`, so that branch now carries both and the pin follows it again. Forward-only, as required.

Verified: zenoh-pico builds with `Z_FEATURE_LINK_ISOTP=1 -DZ_FEATURE_LINK_ISOTP_VENDORED=1` on the rebased submodule.


jerry73204:
Superseded by #50. This pinned `1b851957`, which the ASCII pass over the prose has since rewritten to `dac320e3`; #50 carries the same content plus that cleanup, and also brings the phase-doc update and the arbitration harness.

PR #39, opened 2026-08-29
refactor(phase-399): 203 gates left the root justfile, and `just` now answers one question

jerry73204:
`just --list` was 215 lines opening with a wall of `check-*`. Measured before designing:

```
root justfile   332 recipes, 7,066 lines
  check-*       203 recipes, ~3,041 lines with rationale (43%)
  ungrouped     211 of 332
```

The 21 platform modules were never the problem. **203 implementation gates shared a namespace with the eight verbs a person types.** Nobody memorises `check-cmake-find-program-shadowed` — it runs as a unit of `check-fast` and is discovered by *failing*, which prints its own name.

## W1 — `import`, not `mod`

The whole design decision. `mod` gives `just check fast` and forces a rename across five workflows, `run-gates-parallel.sh`'s awk, CLAUDE.md, AGENTS.md, every issue quoting a command. `import` merges into the same namespace — **no call site anywhere changes.**

```
justfile     7,066 -> 4,141 lines
check.just           2,942 lines, 200 recipes
recipe set   185 before / 185 after, byte-identical
```

`check-fast-serial` deliberately stays at root: `run-gates-parallel.sh` parses its dependency list from the root file, so moving it would have *silently emptied* the parallel runner.

## W2 — partly

`default` prints 28 lines grouped by when you need it. **Not done, and the doc says so:** 211 recipes are still ungrouped and `just --list` still describes each gate with the tail of its rationale paragraph. ~200 further edits; putting them in this commit would have made neither reviewable.

## W3

AGENTS.md said `ci-l1` is "the SAME tier the merge group runs" — false since phase-396 W1. CI runs `check-fast` + `test-unit`; `ci-l1` also runs `check-build` + `check-api-parity`. The local tier is a deliberate **superset**.

## A latent bug the move surfaced

`check-just-recipe-refs` followed `mod` but not `import`, so imported recipes read as **undefined**. Blind to `sdk-env.just` since that import landed; moving 200 gates made it report all 200 missing at once. Visible recipes 133 -> 336. Five self-tests pin it.

Plus the `check-lane-contracts` event-attribution fix reported earlier.

`just check-fast` green, 135 gates, before and after.

https://claude.ai/code/session_011UTfthJ6zZou2YUPZS1jyH


jerry73204:
Superseded by #43 — same commits, from a branch named honestly.

This branch was `fix/0884-event-attribution`. I never reserved issue **0884**; I reserved 0883 and phase 399, and another session legitimately holds `refs/issue-ids/0884` for an unrelated issue. The commits and PR body reference only phase-399, so nothing false was claimed in the content — but the branch name implied a claim, which is the exact collision `just issue-new` exists to prevent.

Re-pushed as `refactor/phase-399-justfile-surface`.

PR #43, opened 2026-08-29
refactor(phase-399): 203 gates left the root justfile, and `just` now answers one question

jerry73204:
Replaces #39, which was opened from a branch named `fix/0884-event-attribution` — an issue ID I never reserved and which another session legitimately holds. Same commits, honest branch name.

`just --list` was 215 lines opening with a wall of `check-*`. Measured before designing:

```
root justfile   332 recipes, 7,066 lines
  check-*       203 recipes, ~3,041 lines with rationale (43%)
  ungrouped     211 of 332
```

The 21 platform modules were never the problem. **203 implementation gates shared a namespace with the eight verbs a person types.** Nobody memorises `check-cmake-find-program-shadowed` — it runs as a unit of `check-fast` and is discovered by *failing*, which prints its own name.

## W1 — `import`, not `mod`

`mod` gives `just check fast` and forces a rename across five workflows, `run-gates-parallel.sh`'s awk, both agent docs, and every issue quoting a command. `import` merges into the same namespace — **no call site anywhere changes.**

```
justfile     7,066 -> 4,141 lines
check.just           2,942 lines, 200 recipes
recipe set   185 before / 185 after, byte-identical
```

`check-fast-serial` stays at root: `run-gates-parallel.sh` parses its dependency list from there, so moving it would have **silently emptied** the parallel runner.

## W2 — partly, and the doc says so

`default` prints 28 lines grouped by when you need them. **Still open:** 211 ungrouped recipes, and `just --list` describing each gate with the tail of its rationale paragraph. ~200 further edits; bundling them would have made neither change reviewable.

## W3 + docs

AGENTS.md said `ci-l1` is "the SAME tier the merge group runs" — false since phase-396 W1. CI runs `check-fast` + `test-unit`; `ci-l1` also runs `check-build` + `check-api-parity`. CLAUDE.md gains that plus four pitfalls: Kconfig last-wins (0876), the column-0 recipe/YAML trap, the committed-generated-file queue stall (0883/0884), and verdict-vs-no-verdict.

## Two latent defects this surfaced

- **`check-just-recipe-refs` followed `mod` but not `import`**, so imported recipes read as *undefined*. Blind to `sdk-env.just` since that import landed. Visible recipes 133 -> 336; five self-tests pin it.
- **`check-ros-env-spelling` allowlists by PATH**, and moving `check-dep-chain` dropped it out of an exemption it still earns. Entry added naming the move.

Plus the `check-lane-contracts` event-attribution fix (it read the `check` job's guard as `pull_request` only; it runs on every event).

`just check-fast` green, 135 gates.

https://claude.ai/code/session_011UTfthJ6zZou2YUPZS1jyH


jerry73204:
Rebased onto `7ee0b5e65`. The conflict was two regions where main added recipes and this branch had deleted the whole area, so taking either side would have been wrong. Resolved by comparing **every recipe body** between main's `justfile` and this branch's `justfile + just/check.just`: **345 vs 345, nothing missing, nothing extra, two body differences — both this PR's own intended changes** (`default` = the new help text, `export` = the `import` line).

That comparison caught things taking a side would have buried:

- **Four recipes this refactor had silently dropped** — `changelog`, `changelog-add`, `changelog-release`, `dev-tools`. They existed at the branch point and were in no split file; `dev-tools` is documented in CLAUDE.md. Restored.
- **Three gates main added while this was open** — `check-infra-queryable-counts`, `check-no-tracked-workspace-roots`, `check-orphan-generated-stamp`. `justfile:515` already listed one as a lane **dependency** while the recipe was undefined, so these were load-bearing.
- **Three main edits to moved recipes**: the `graph query entry points are DECLARED` probes in `check-c`/`check-cpp` (phase-381 W4), and `check-issue-index` gaining `gen-issue-index.py --self-test`.
- `prune-artifacts` default dir `mixed` -> `features`.

Two further commits, both fallout this PR owns because it created the second location:

`a676e12db` — `check-gate-selftests` reported **117 problems**, every gate script at once, each "backs no `check-*` recipe any more". Same latent hole `check-just-recipe-refs` already carried: `import` merges into the root namespace, so a tool reading only `justfile` sees none of the 200 that moved. Swept the other 13 scripts that read `justfile` (`grep -rln '"justfile"\|/justfile' scripts/ .githooks/`); seven already cover `just/`, one — `check-skippable-tests-tolerant` — was blind in principle and is widened. It loses no coverage today, but a gate whose reach is narrower than its rule still passes, which is issue 0196.

`3c4a8179d` — the `grep -q` ratchet is per FILE, so `check-cpp`'s move read as growth. Confirmed a relocation before touching the baseline: total **91 -> 91, delta 0**, `justfile: 3 -> 2` offsetting `just/check.just: 0 -> 1`.

`just check-fast`: **138 gates green** (135 before — the three restored gates now run). `just ci-l1`: green, 1020 passed, 1 skipped for an unmet host capability.

PR #61, opened 2026-08-29
fix(zephyr): forward NROS_ZEPHYR_HEAP_SIZE to cargo

### LINKED ISSUE #41: rlsf funnel: CONFIG_HEAP_MEM_POOL_SIZE no longer governs allocation, and starvation is silent
## Summary

`main` (1654842f1) does not boot to the application on `fvp_baser_aemv8r`. The image comes up, gets DHCP, enumerates the interface, and then stops — never reaching SNTP or the node's own startup.

The previous pin this consumer used, `589a9d0cf` ("fix(phase-385): raise the AN536 heap"), boots the identical application fine. So this is a regression somewhere in the ~359 commits between them.

## Symptom

```
*** Booting Zephyr OS build v3.7.0 ***
<inf> net_config: Initializing network
<inf> net_config: Running dhcpv4 client...
Secondary CPU core 1 (MPID:0x0) is up
Secondary CPU core 2 (MPID:0x1) is up
Secondary CPU core 3 (MPID:0x2) is up
<inf> nros: nros: Waiting for DHCP to get IP address...
<inf> net_dhcpv4: Received: 172.20.51.1
<inf> net_config: IPv4 address: 172.20.51.1
<inf> nros: nros: Network interfaces found: 1
                                                   <- stops here
```

Guest time at the last line is 10.159 s; the model was then stopped by the host-side timeout. On `589a9d0cf` the same image continues past this point to SNTP and then to application startup.

## What was ruled out

I first hit this while pinning a consumer to a feature branch, so the obvious suspicion was my own change. It is not:

- `main` **plus** the feature commit — hangs at that line
- `main` **without** the feature commit (plain `1654842f1`) — hangs at that line, identically
- `589a9d0cf` plus the same feature commit — boots normally, runs a full autonomous mission

So the feature branch is not implicated, and the regression reproduces on unmodified `main`.

## Environment

- Board: `fvp_baser_aemv8r` / `fvp_aemv8r_aarch64/smp`, Arm `FVP_BaseR_AEMv8R` 11.31.28
- Zephyr v3.7.0 LTS
- RMW: CycloneDDS. Features: `rmw-cffi,alloc,platform-zephyr,ros-humble,param-services`
- C++ API lane (`CONFIG_NROS_CPP_API`), boot tier on `main`
- Consumer: autoware-safety-island, `.github/scripts/run-zephyr-fvp-ci.sh` phase 1

## What I have NOT done

I have not bisected the 359 commits, so I cannot name the culprit and am deliberately not guessing at one. Happy to bisect if that would help — it is roughly nine builds — but I wanted the report filed first in case the symptom is immediately recognisable to someone who knows what landed in that range.

The last line printed comes from nano-ros's own startup path, so whatever blocks is after interface enumeration and before the SNTP step.

## Impact

Blocks moving a consumer's pin onto `main`. That consumer is currently pinned to a side branch (previous working base plus its feature commit) as a stopgap.

jerry73204:
Fixes #41.

## The problem

`NROS_ZEPHYR_HEAP_SIZE` is documented in `nros-platform`'s `zephyr_heap` and **unreachable from a Zephyr build**. The Zephyr integration forwards an explicit set of knobs into the cargo environment via `_nros_resolve_knob`, and this one was never registered — so it never appeared in the generated cargo command. Exporting it had no effect whatsoever; the arena stayed at its 64 KiB default.

This matters because `60b4e0c1e` moved `z_malloc` **and** `__rust_alloc` off Zephyr's kernel heap onto that arena. `CONFIG_HEAP_MEM_POOL_SIZE` therefore no longer governs application allocation, and a consumer needing more than 64 KiB had no working way to ask for it.

Starvation presents as a **silent hang** — no fault, no log, no error code. The image stops between `Network interfaces found: 1` and node startup and stays there. It took a 9-step bisect to attribute, and then several more wrong turns to work out why setting the documented knob changed nothing.

## The fix

Register the knob the same way every other one is registered, and add the matching Kconfig symbol so it is settable from a `.conf` fragment like its neighbours, with the environment still winning.

The help text says outright that `CONFIG_HEAP_MEM_POOL_SIZE` is no longer the knob, and that outgrowing the arena hangs rather than errors — because the next person to hit this will read Kconfig, not the commit log.

## Verification

Measured, not inferred — I had already talked myself into three wrong explanations, so this is checked at the two points where it can be observed directly.

**In the generated build rule** (`build.ninja`):

```
before:  (absent)
after:   NROS_ZEPHYR_HEAP_SIZE=4194304
```

**In the linked ELF** (`nm -S | grep zephyr_heap4HEAP`):

```
before:  0000000000010b38   (68,408 bytes — the 64 KiB default)
after:   0000000000400b38   (4,197,176 bytes)
```

**At runtime**, a consumer image on `fvp_baser_aemv8r` that previously hung now boots, and its full FVP CI passes all six phases — controller smoke, unit tests, DDS loopback, CAN loopback, TAP networking, scheduling statistics.

## Note on scope

This is deliberately separate from #40 (callback tracing). The two are unrelated concerns and the consumer happens to need both; they should be reviewable independently.

## Still open from #41, not addressed here

Arena starvation remains silent. This change makes the arena *settable*, so a consumer can now fix it once they know — but the first encounter is still a hang with nothing in the log naming the heap. A message at allocation failure would have turned that bisect into a one-line read, and is worth doing separately.

### DIFF
--- a/packages/boards/nros-board-mps2-an385-freertos/Cargo.lock
+++ b/packages/boards/nros-board-mps2-an385-freertos/Cargo.lock
@@ -565,6 +565,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-mps2-an385/Cargo.lock
+++ b/packages/boards/nros-board-mps2-an385/Cargo.lock
@@ -621,6 +621,7 @@ dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
  "nros-platform-mps2-an385",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-mps3-an536-freertos/Cargo.lock
+++ b/packages/boards/nros-board-mps3-an536-freertos/Cargo.lock
@@ -510,6 +510,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-nuttx-qemu/Cargo.lock
+++ b/packages/boards/nros-board-nuttx-qemu/Cargo.lock
@@ -497,6 +497,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-s32z270-freertos/Cargo.lock
+++ b/packages/boards/nros-board-s32z270-freertos/Cargo.lock
@@ -510,6 +510,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-threadx-linux/Cargo.lock
+++ b/packages/boards/nros-board-threadx-linux/Cargo.lock
@@ -507,6 +507,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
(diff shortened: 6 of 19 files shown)


jerry73204:
Closing: fully subsumed by #75, which merged the whole stack. `git cherry -v origin/main origin/$(git rev-parse --abbrev-ref HEAD)` marks every commit on this branch `-` — already upstream by patch-id — so it would merge as a no-op.

The content is on main under the queue's rebased shas: `6d95c1a8b` (forward the knob), `d25923826` (resolve it in a build script), `c13896f0b` (the 14 leaf locks).

PR #71, opened 2026-08-30
fix(zephyr): heap starvation says so, and an impossible size fails the build

jerry73204:
Addresses the remaining half of #41. **Stacked on #61** — the build-time check reads the resolved knob that PR registers, so review that one first.

## The problem

Arena exhaustion is completely silent. `nros_platform_alloc` returns NULL, the caller stalls, and the image stops mid-boot: no fault, no log, no error code. The only evidence is *where* it stopped.

That is how a 64 KiB default cost a nine-step bisect to attribute, and then several more wrong turns to explain.

## What it actually was

Instrumenting the allocator gave the answer in one line:

```
nros: HEAP EXHAUSTED: request 477104 bytes, arena 66048 bytes
```

A **single allocation 7.2× the size of the whole arena**. Not gradual exhaustion, not a marginal shortfall — one request that could never have succeeded, fully determined at build time.

It is the executor arena. Confirmed causally rather than by correlation — two builds differing only in `NROS_EXECUTOR_ARENA_SIZE`:

```
executor arena 458752  →  failing request 477104   (Δ 18352)
executor arena 100000  →  failing request 118352   (Δ 18352)
```

Same constant overhead both times. So the executor arena is one heap allocation out of the platform arena, and the two sizes are coupled by an invariant nothing checked:

```
NROS_EXECUTOR_ARENA_SIZE + overhead  <=  NROS_ZEPHYR_HEAP_SIZE
```

## The fix

**Build time** — the important half. Both values are known at configure time, so an impossible combination is now a build error naming both knobs and the required size, rather than an image that builds, links, and hangs:

```
nros: the executor arena cannot fit in the platform heap.
  NROS_EXECUTOR_ARENA_SIZE = 458752
  NROS_ZEPHYR_HEAP_SIZE    = 65536
  needed (arena + ~24576 overhead) = 483328
```

**Runtime** — for exhaustion that is not statically decidable. `printk` rather than `LOG_*`, because the logging subsystem may itself allocate and this is the path that just failed to allocate; emitted after the spinlock is released, so nothing reentrant runs under the funnel's lock.

## On the overhead constant

It is **empirical, not derived**, and the comment in the source says so. The two-point measurement above pins it at 18352; the check rounds to 24 KiB and treats it as a floor rather than a formula, which is why it adds margin instead of comparing bare sizes. If someone knows the real composition, tightening it is welcome — but a slightly conservative check that fires is worth more than an exact one that does not exist.

## Verification

- Fires on 458752-against-65536, with the message above.
- Passes at the sizes a working consumer image actually uses.
- The runtime message is the output quoted at the top — produced by a real image, not constructed by hand.

### DIFF
--- a/packages/boards/nros-board-mps2-an385-freertos/Cargo.lock
+++ b/packages/boards/nros-board-mps2-an385-freertos/Cargo.lock
@@ -565,6 +565,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-mps2-an385/Cargo.lock
+++ b/packages/boards/nros-board-mps2-an385/Cargo.lock
@@ -621,6 +621,7 @@ dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
  "nros-platform-mps2-an385",
+ "nros-zephyr-build",
 ]
 
 [[package]]
--- a/packages/boards/nros-board-mps3-an536-freertos/Cargo.lock
+++ b/packages/boards/nros-board-mps3-an536-freertos/Cargo.lock
@@ -510,6 +510,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-zephyr-build",
 ]
 
 [[package]]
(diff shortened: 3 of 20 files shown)


jerry73204:
Closing: fully subsumed by #75, which merged the whole stack. `git cherry -v origin/main origin/$(git rev-parse --abbrev-ref HEAD)` marks every commit on this branch `-` — already upstream by patch-id — so it would merge as a no-op.

The content is on main under the queue's rebased shas: `6d95c1a8b` (forward the knob), `d25923826` (resolve it in a build script), `c13896f0b` (the 14 leaf locks), `ee0d1933a` (heap starvation says so).

PR #76, opened 2026-08-30
docs(#0917): the emulated LAN9118 RX FIFO cannot hold an 8-fragment RTPS burst

jerry73204:
Files what is left on the an536 lane after #0836.

**The finding.** A subscription whose samples span several RTPS fragments delivers nothing at all in most runs, while a single-datagram sample on the same participant, same link, same run delivers every time. Six trials per size, one host participant, no Autoware:

| payload | fragments | reader got it |
| --- | --- | --- |
| 908 B | 1 | **6 of 6** |
| 10588 B | 8 | **2 of 6** |

The shape matters more than the ratio: it is **binary**. Where the 8-fragment stream came up it stayed fresh for the whole run; where it did not, nothing arrived in 25 s of a 10 Hz publisher — ~250 samples, all sent. No middle state, so it is not a throughput ceiling. Something has to succeed once, early, and a run that misses it never recovers.

**Ruled out by measurement**, each against *this* failure rather than by analogy with 0836:

- lwIP — every pool `err=0`, peaks 16 of 64/128, `udp.recv` climbing, ICMP answering
- the socket handoff — recvmboxes empty, `rcvevent=0`, so Cyclone's `recv` thread keeps up
- the fragments being sent — 1416 `DATA_FRAG` + 1239 `HEARTBEAT_FRAG` on the wire in a failing run
- Cyclone receive buffers — 64 KiB/16 KiB to 1 MiB/128 KiB, no effect (the earlier test of that knob was against 0836's discovery failure, not this one)
- the NACKFRAG path itself — it does fire at a low publish rate, 17 in a traced run

**What the trace shows.** 33 `defrag_rsample_drop` on one defrag admin, while the wire carries zero NACKFRAG from it: partial samples accumulate to `DefragReliableMaxSamples` and are evicted before anything repairs them, each new sample evicting an older one. Self-sustaining, which is the mechanism for "never recovers". Why the reader stops asking for fragments it is missing, when the same reader asks at a lower rate, is the open question.

Filed rather than fixed — the rig lives in the ASI consumer and the failure needs repeated runs to read, so this records the reproducer and the eliminations while they are fresh.

`just check-fast` — 138/138 gates OK.

### DIFF
--- a/docs/issues/0917-an536-fragmented-sample-never-syncs.md
+++ b/docs/issues/0917-an536-fragmented-sample-never-syncs.md
@@ -134,6 +134,48 @@ lan9118 driver never calls the lwIP link-stat macros, so loss at exactly this
 layer is invisible to `lwip_stats`. Every pool below it reads clean while the
 NIC is dropping the tail of every burst.
 
+## Measured with the link paced to 100 Mbps
+
+The lane can now be honest about the wire: `scripts/setup-tap.sh --rate 100mbit`
+in the ASI consumer puts `netem rate` on the tap so frames are serialised at
+the speed the emulated PHY advertises, instead of arriving as fast as the host
+can write them. Everything below is measured with that on.
+
+**Pacing alone changes the headline failure.** The stock image — unchanged, 5 ms
+polled drain — goes from delivering the 8-fragment sample in 0-1 runs of 6 to
+delivering it in **6 of 6**. So "the fragmented sample never arrives" was in
+large part the unpaced emulator, not the product. That is a lane-fidelity
+finding and it belongs at the top of this issue rather than buried in it.
+
+**Packet loss below IP** (wire datagrams counted with `dumpcap -p`, against the
+island's cumulative `ip.recv`; the layer `LINK_STATS` cannot see because the
+lan9118 driver never calls those macros):
+
+| build | link | loss below IP |
+| --- | --- | --- |
+| stock, 5 ms poll | unpaced | 20.8 % |
+| stock, 5 ms poll | paced 100 Mbps | 11.8 % |
+| interrupt-driven RX | paced 100 Mbps | **-0.3 %, 13.3 %, 17.3 %, -0.3 %, 14.1 %** |
+
+The last row is the important one and it is **bimodal**, not an average: two
+runs of five lose essentially nothing, three lose 13-17 %, same binary. So
+interrupt-driven RX can reach zero loss on this lane and does not do so
+reliably, and reporting a mean here would invent a number that no run produced.
+
+**Freshness is NOT explained by loss.** The controller calls a trajectory stale
+after 0.5 s, and it is stale in every arm — stock or interrupt-driven, paced or
+not. Including a run that lost **0 %**. A sample that arrives with no packets
+dropped and is still not delivered at 2 Hz is not a network-stack problem, so
+the residual is above lwIP: Cyclone's receive/defrag path and the executor's
+take rate on an emulated Cortex-R52 carrying ~200 packets/s. That is issue 0889
+territory, not the driver's.
+
+**So the arithmetic below was half right.** The budget correctly says a 5 ms
+polled drain cannot service this FIFO at line rate, and interrupt-driven RX
+does remove the loss when it works. It does not follow — and the measurements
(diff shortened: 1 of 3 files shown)


jerry73204:
Root cause found — pushed as a third commit, and the issue title now names it.

**The burst does not fit in the NIC.** QEMU's LAN9118 model sizes its RX FIFO in *words*: `rx_fifo_size = 2640` (10 560 bytes), and `lan9118_can_receive` refuses a frame unless 384 words (1536 B) are free. A 10 588-byte sample goes out as 8 RTPS fragments, ~1420 B each on the wire — ~11 360 bytes back-to-back. Six fit; the seventh is marginal against the headroom rule; the eighth never has room.

The island's defrag trace matches exactly. Contiguous reassembly reaches:

```
159 x [0..8064)   6 fragments   <- every sample gets here
  3 x [0..9408)   7 fragments   <- three ever
  0 x [0..10588)  8 fragments   <- none, ever
```

A hard cutoff at six frames is capacity, not loss.

The guest half is `poll_task_entry`: `vTaskDelay(poll_interval_ms)` then drain <=16 packets, with `poll_interval_ms = 5` against a 1000 Hz tick. Nothing drains between the frames of one sample.

That accounts for the whole shape reported above — 1 fragment never touches the limit, 2-4 fit, 8 lose their tail every time so no sample completes, the defrag admin fills with partials and evicts them, and the reader never recovers.

**Measured mitigation:** polling every tick instead of every 5 ms, same binary otherwise — **2 of 6 to 9 of 11**. Better, not a fix: the guest is not scheduled between frames of one sample at all, so a burst outruns any fixed cadence. Closing it wants interrupt-driven RX (the model does raise RX interrupts) or a drain that continues while packets are pending — the poll task sleeps *first* and drains second, so the interval is a floor on how long a full FIFO waits.

Diagnostic note: `LINK_STATS` reads all zero because the lan9118 driver never calls the lwIP link-stat macros, so loss at this exact layer is invisible to `lwip_stats` — every pool below read clean while the NIC dropped the tail of every burst.

jerry73204:
Correction, and the issue now reads the other way round.

**The lane keeps the stock FIFO.** The widened-FIFO run (6/6 delivered and fresh, vs 0–1/6 stock) stays in the issue as *diagnosis* — it isolates the binding resource, which nothing guest-side could have shown — but shipping it would remove the very constraint the product has to satisfy. `2640` words is what the real LAN9118 has.

**The timing budget is what decides the fix.** The part is 10/100 and the driver advertises 100BASE-TX, so at line rate:

```
frame on wire        1440 B (1420 + preamble/IFG)  ->  115 us
RX FIFO              10560 B, usable 9024 after the
                     1536 B can_receive headroom    ->  holds 6 frames
time to fill         6 x 115 us                     ->  691 us
an 8-fragment burst  8 x 115 us                     ->  922 us
```

| drain latency | verdict |
| --- | --- |
| `poll_interval_ms = 5` (today) | 5000 us — **7x too slow** |
| polling every 1 ms | 1000 us — still too slow |
| RX interrupt (~50 us) | fits, with an order of magnitude to spare |

That is a defect on real silicon: a 5 ms polled drain cannot service a 10.5 KB FIFO on a 100 Mbps link, and RTPS fragmentation of an ordinary ROS 2 topic sends far more than six back-to-back frames.

**So my earlier verdict in this PR was wrong.** I measured interrupt-driven RX as no better than the 5 ms poll and concluded it was not the fix. The reason it measured that way is that QEMU's tap delivers frames as fast as the host writes them, with no pacing to the modelled PHY — a burst that occupies 922 us of wire arrives effectively instantaneously, the guest is never scheduled inside it, and no drain latency can help. The lane is *harsher* than the hardware, not more faithful. Corrected in place rather than left as a contradiction two sections apart.

**What to do**, now in the issue:

1. Land interrupt-driven RX — the prototype is complete and boots; it needs review, not discovery.
2. Keep drain-first/wait-second with `poll_interval_ms` as a ceiling regardless.
3. Do **not** widen the emulated FIFO.
4. Pace the tap to the modelled link speed in the lane, so the fix can be demonstrated rather than argued from arithmetic. Needs root, so operator/CI:

```
sudo tc qdisc add dev tap1 root tbf rate 100mbit burst 32kbit latency 5ms
```

PR #80, opened 2026-08-30
fix(#0915, #0921): the cdylib takes the stable ABI; build scripts follow the submodule symref

jerry73204:
Three issues, one branch — they were found in sequence, each while verifying
the one before.

## #0915 — abi3 was designed in and never enabled

0897's shape specifies the cdylib as `extension-module + abi3-py310`. Only the
first shipped, and they do different jobs: `extension-module` stops pyo3
emitting the link directive so the artifact names no interpreter — **loadable**
against one chosen at runtime. `abi3` restricts pyo3 to the stable ABI so the
symbols it leaves undefined are ones every CPython >= 3.10 exports unchanged —
**correct** against a version other than the one it compiled against. Without
the second, "no `DT_NEEDED`" buys loadability and nothing else.

Before: 97 undefined `Py*` symbols including six CPython internals
(`_Py_Dealloc`, `_PyObject_MakeTpCall`, `_Py_CheckFunctionResult`, …) plus
`Py_CompileStringExFlags` and four `PyEval_*` outside the limited API.

**0897 acceptance criterion 1 is now met, for the first time** — one build,
two CPython minor versions, same machine:

```
libpython in driver:  0
private syms:         _Py_DecRef _Py_FalseStruct _Py_IncRef
                      _Py_NoneStruct _Py_TrueStruct     (all stable-ABI)
CPython 3.10.12:      SystemModel (1 nodes)
CPython 3.12.14:      SystemModel (1 nodes)
```

on the `$(eval "robot1" in ("robot1","all"))` file that had `host-tests` red.
The second interpreter came from `uv python install 3.12`, no root — worth
knowing, since "only one Python here" is why this criterion went untested since
the design was written. The port cost no code: pyo3 gates non-limited symbols
at compile time, so it compiled, and `Py_CompileStringExFlags` simply became
`Py_CompileString`.

Upstream also fixes the diagnostic reported from the field — "this build has no
Python backend" on a machine with working Python. It now names the cause, lists
the paths searched, and says the driver and the half are two artifacts.

## #0921 — build scripts watched a file a commit does not move

Both `nros-cli-core/build.rs` and `nros-launch-resolve/build.rs` bake the
play_launch commit and keep it fresh by watching `<gitdir>/HEAD`. True for a
DETACHED submodule. On a BRANCH, `HEAD` holds `ref: refs/heads/<branch>` and
never changes — the commit moves the ref file. So the script did not re-run,
the binary kept the previous sha, and the 0409 guard compared a lie: the exact
failure 0419's comment says it prevents. **The blind case is the one that
matters** — a submodule is on a branch precisely while someone is developing a
change to it.

One shared helper, both sites together (fixing one alone leaves the guard
comparing a *fresh* pin against a stale one). Proven end-to-end:

```
git commit --allow-empty      # moves the ref and NOTHING else
just setup-launch-resolve
nros-launch-resolve --version -> play_launch 030f34a2…
```

`tests/submodule_watch.rs` covers branch, detached, uninitialised and plain-repo
shapes. The helper is `include!`d only by build scripts, which nothing else
compiles — so it had no test at all, which is how this survived two prior edits
to the same two files.

Also: `setup-launch-resolve`'s skip path was a bare `exit 0`, so "I rebuilt it"
and "I decided not to" were indistinguishable. Both paths now report and both
name the pin.

## #0919 — a diagnostic, not a fix

`host-tests` fails on 16 undefined zenoh-pico platform symbols in
`bridge-cyclonedds`'s `native_entry`. CI resolves a `zpico-sys` unit no local
build produces — different hash, and by the error one built without the alias
TU. **Eight local routes link**, including the exact `nros build` command with
CI's profile and RUSTFLAGS, after a fresh `nros sync`. So the difference is
inside CI's resolution and no local run can supply it.

This prints it, on the failure path only: `cargo tree -e features -i zpico-sys`
plus both generated manifests, for the failing workspace *and* `rust` — the
useful reading is the difference, since `rust`'s entry links in the same job.

Issue 0919 records what is established, the eight routes, and two things worth
fixing regardless: `platform-posix` reaches the graph through exactly one path
(a generated file, unchecked), and `any_explicit` gates configuration and the
alias TU differently, so a build relying on `auto_posix` gets zenoh-pico
configured for POSIX with its platform symbols undefined.

## Verification

`just ci-l1` green (1020 passed, 1 skipped for an unmet host capability);
`just check-fast` 138/138; parser workspace suite green upstream.

https://claude.ai/code/session_01A2zFRQS5rDsPBNWH85JjJB


jerry73204:
Rebased onto current main and **#0919 is now fixed here**, not merely diagnosed.

I stopped waiting for the CI diagnostic and went looking for the graph shape instead. All eight earlier routes failed to reproduce because every one had `platform-posix` enabled; reproducing needs the opposite — zpico present **without** it. That is exactly the bridge entry's shape: it depends on `nros-rmw-zenoh` DIRECTLY (so `platform-aliases` is on by default and zpico compiles) while reaching it through a board that selects only another RMW.

Forced by making the generated selection crate name one RMW:

```diff
-nros-board-linux = { …, features = ["rmw-cyclonedds", "rmw-zenoh"] }
+nros-board-linux = { …, features = ["rmw-cyclonedds"] }
```

and it fails with the reported 16 symbols, deterministically, locally, in about a minute.

**The defect is four lines:**

```rust
let any_explicit = use_posix || use_zephyr || … ;
if !any_explicit && auto_posix {
    use_posix = true;          // a platform IS now resolved
}
```

`any_explicit` is computed BEFORE the inference and never updated, so on a hosted target that resolved POSIX by inference it stays **false** while `use_posix` is **true**. The alias TU was gated on it, so zenoh-pico got configured FOR POSIX and then the platform forwarders that configuration requires were never emitted — half-configured, detectable only at link time in a consumer.

The gate now asks `platform_resolved`, computed AFTER the inference. Verified on the same configuration: the graph that failed now links, and the normal two-RMW graph still links.

The CI diagnostic stays. It would have shown `platform-posix` absent in one line, and the next failure of this family should not need eight routes.

`just ci-l1` green (1027 tests, 1 host-capability skip).

PR #97, opened 2026-08-30
fix(#0899, #0906, #0920, #0924): the FreeRTOS talker crash, and the three defects under it

jerry73204:
Four issues, one thread pulled end to end: a FreeRTOS talker that died mid-run,
what was actually killing it, and what the fix exposed underneath.

## What is fixed

**#0899 — a use-after-free, two victims.** `_zp_unicast_failed` runs on the
LEASE task and frees `_wbuf` and `_mutex_tx` while an application task can be
inside the same transport publishing. That accounts for two assertions that
looked unrelated: `_z_wbuf_clear` empties the slice list so the publisher's next
`__unsafe_z_prepare_wbuf` trips `iobuf.c`'s assert, and `vSemaphoreDelete` frees
the queue so the next lock reaches `xQueueTakeMutexRecursive` on freed memory.
The crashing handle was byte-identical to the `_mutex_tx` the lease task had
just deleted.

`_z_session_t` gains a recursive `_mutex_transport` — the lock cannot live in
the transport, because the thing being freed cannot guard its own lifetime — and
the teardown side is a **handshake, not a critical section**: holding the lock
across `_z_link_free` deadlocks in lwIP, so the lease task publishes
`_tp._type = _Z_TRANSPORT_NONE` under the lock and lets go.

    before:  19 / 19 / 19 publishes, FreeRTOS queue.c assert every run
    after:   76 / 77 / 77 / 77 / 76, five runs, no assert and no hang

**#0906 — the client lease was shorter than the router's keep-alive cadence.**
The ROS router announces `lease: 60000` with `keep_alive: 2`, so an idle
`rmw_zenohd` speaks every 30 s. `Z_TRANSPORT_LEASE` was 10 s and zenoh-pico
takes `min(peer_lease, Z_TRANSPORT_LEASE)`, so the session expired while the
router was still 20 s from its next keep-alive — deterministically, every
session, on every platform we ship. The capture that opened the issue was not a
router going silent; it was a router being early, measured against a deadline no
peer had agreed to.

    before:  77 published, 19 heard, 58 publish errors, reconnect every ~20 s
    after:   330 s run — 327 published, 325 heard, 0 errors, exactly 2 TCP
             handshakes (one per node)

Native Linux, same router: 105 publishes, **one** handshake. It had been
reconnecting at ~20 s too, silently.

**#0924 — a graceful close waiting on a peer that will never answer.**
`_z_close_tcp` did `shutdown(fd, SHUT_RDWR)` then `close(fd)`. Every caller
reaches it because the transport is already being torn down, and the commonest
reason is that the peer stopped answering — so `shutdown()` sends a FIN and
waits for an ACK that never comes. `SO_LINGER` with a zero timeout aborts
instead: RST, and `close()` returns immediately.

    before:  247 publishes, 227 failures, delivery frozen from the first outage
    after:   3 alternating freeze/kill cycles, 150 s outages: recovers every time

**#0920 — codegen wrote generated files non-atomically.** `fs::write` truncates
then writes; an interrupted run left a zero-byte source file that red-lined
`check-examples` with a relative path and no leaf name. Now a sibling temp plus
`rename`.

**Plus** the three lock lines `c13896f0b` missed, which were red on `main`
before this branch and are what the pre-push hook stopped me on.

## Corrections in the history

Two wrong root causes are retracted **in the issues**, with the evidence that
killed them: an lwIP concurrency theory read off a stale comment in a driver
header (the board registers `tcpip_input`, not `ethernet_input`), and a
`_z_wbuf_reset` capacity ratchet that is a real upstream defect but unreachable
on the buffer in question. A `0xdeadbeef` I reported as poison was the
debugger's, not the image's.

#0906 also did not fix #0924, though it looked like it: at a 60 s lease the
outages under test merely ended shortly after expiry, so the peer thawed and
ACKed the pending FIN. The lease decides how *often* the wedge is reached, never
whether it is possible.

## Measurement traps, all recorded in the issues

Each of these produced a confident wrong verdict first:

- A leftover **router** keeps serving the port, so the "outage" never happens; a
  leftover **talker** keeps feeding the listener, so a dead run still scores
  messages. Both faked a PASS.
- `pgrep -f` for cleanup matches the shell running the script and kills its own
  caller. Use `pgrep -x`.
- **SIGKILL cannot reproduce #0924** — it resets the socket, so `close()` returns
  immediately. Only SIGSTOP exercises the FIN handshake.
- **Diagnostic `printf`s hide these bugs.** Probe-laden builds run clean where
  the same code stalls. Use `volatile` stage counters read from gdb after the
  hang.
- A peer must be attached, and **both halves rebuilt**, or a stale peer gives
  numbers that look like a fix.

## Submodule

`zpico-sys/zenoh-pico` moves to `b3bce713` on the fork's `nano-ros` patch line,
pushed as a fast-forward so that line keeps linear history. My tip is a strict
descendant of the pin #70 landed, so the bump is forward-only;
`check::submodule-commits-reachable` confirms a fresh clone can fetch it.

## Verification

141/141 fast gates, `ci-l1` green. Tier 2 (`ci-matrix`) is what this diff earns —
it touches codegen, the zenoh build config, a board's lwIP options and the
submodule pin — and is running; I will post the result here before this is ready
to merge.

**Known gap:** every runtime measurement above is mps2-an385 + zenoh. The lease
constant is baked into every zenoh image on every platform, and Zephyr, NuttX,
ThreadX and ESP32 are unexercised by these numbers. That is precisely what the
matrix lane is for.

Also filed, not fixed: **#0930** (the built QEMU can be older than its pin, and
nothing says so — it is, right now, by the commit that changes LAN9118 RX).


jerry73204:
## Checkpoint — what is and is not verified

Pushed as `d2f816cf3` (18 commits). Taking this out of draft so it can reach the
queue; the verification status is below rather than implied, because this branch
grew well past its original scope.

**Verified locally**
- 147/147 `just check fast`, and `just ci-l1` green.
- `#0899` / `#0924`: 76-77 publishes per run with a peer attached, no assert;
  three alternating freeze/kill outage cycles at 150 s each, delivery recovering
  every time. Before: 19 publishes then a FreeRTOS assert, 3/3.
- `#0906`: 330 s run — 327 published, 325 heard, 0 errors, exactly 2 TCP
  handshakes. Native Linux: 105 publishes, one handshake.
- `#0959`: `zenoh_integration` 121 s -> 4.16 s, stable across runs.
- `#0942`: the threadx-riscv64 compile error is fixed and its gate is proven
  against the pre-fix file.
- `#0943` / `#0944`: verified by REMOVING the rust target and reading both
  surfaces report it with a working remedy.

**NOT verified, and this is the gap that matters**
- Every runtime measurement above is `mps2-an385` + zenoh, or native. The
  `#0906` lease constant is baked into every zenoh image on **every** platform —
  Zephyr, NuttX, ThreadX and ESP32 are unexercised by these numbers.
- Tier 2 (`just ci-matrix`) has not completed on this branch. Two attempts were
  blocked by defects this branch then fixed (`#0942`, `#0943`); the third needs
  a `lane=all` fixture build that has not been re-run since.

If the matrix lane is owned elsewhere now, that is the thing to point at it.

**Submodule** — `zpico-sys/zenoh-pico` moves to `fa7ad0f5` on the fork's
`nano-ros` patch line, pushed as a fast-forward. My tip is a strict descendant of
the pin currently on main, so the bump is forward-only, and
`check::submodule-commits-reachable` confirms a fresh clone can fetch it.

PR #99, opened 2026-08-30
fix(#0935): a `.launch.py` resolves again — context crosses the dlopen boundary

jerry73204:
Every Python launch file **aborted** through the shipped resolver, including one that imports nothing:

```
No LaunchContext set - Python execution must be called through LaunchTraverser
panic in a function that cannot unwind / thread caused non-unwinding panic
```

### Cause

`exec_file` communicated through a process-wide thread-local that stopped being process-wide. Since issue 0897 split the Python half into a `dlopen`ed object, **both** sides statically link `play_launch_parser`, so each has its own `CURRENT_LAUNCH_CONTEXT`. The binary set one; Python ran inside the object and read the other, found `None`, and took the process down.

`$(eval …)` never noticed because its request carries the expression and its response the result. `exec_file` now has that shape too — configurations in the request, captures in the response (NEWSLabNTU/play_launch@d16b354, ABI 1 → 2).

### The test that was missing

`launch_py_resolves_as_shipped` invokes the **installed** binary by path against a `.launch.py` and asserts the Python-declared node reaches the model. Every existing test linked the Python half in-process, where one copy of the parser makes a thread-local look shared — which is exactly why this passed everywhere while being broken as shipped. Confirmed by deleting `libplay_launch_parser_pyexec.so`: it goes red.

play_launch also gains two ABI-level tests asserting on the **response**; both fail when the drain is removed.

### Issue 0914 residue, closed with it

`host_python_available()` becomes a shared helper and `multihost_partition_bake` uses it. Without a probe those tests *failed* on a Python-less host where they should skip — "no Python here" and "the pair is broken" produce the same parse error.

### Also: three locks main left behind

`Cargo.lock` and the two `nros-nuttx-*-ffi` leaf locks are one line each, adding no package entries — the refresh the gate permits, not a dependency change. They are red on main itself, so any branch cut from it fails `check fast` before running its own code. Not mine to have caused, cheaper to fix here.

`just ci l1` green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1927,6 +1927,7 @@ dependencies = [
  "nros-platform-esp32-qemu",
  "nros-platform-mps2-an385",
  "nros-platform-stm32f4",
+ "nros-zephyr-build",
  "zpico-alloc",
 ]
 
--- a/docs/issues/archived/0935-launch-py-aborts-through-the-shipped-pyexec.md
+++ b/docs/issues/archived/0935-launch-py-aborts-through-the-shipped-pyexec.md
@@ -2,7 +2,7 @@
 id: 935
 title: "Every `.launch.py` ABORTS through the shipped resolver — `exec_file`'s
   inputs and outputs travel by a thread-local the dlopen split duplicated"
-status: open
+status: resolved
 type: bug
 area: tooling, cli
 related: [0914, 0897, 0915, phase-332]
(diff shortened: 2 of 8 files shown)


jerry73204:
Superseded by #103, which contains this work.

When #103 was rebased onto main, these two commits came along with it:

- the `.launch.py` fix (`exec_file` carries context both ways across the dlopen boundary) — `git cherry` confirms it is already there;
- the 0935 resolution — verified by substance rather than commit hash: the issue is archived as `resolved` in #103 and gone from `open.md`, and the play_launch submodule is pinned at `d16b354`.

The three lockfile refreshes this PR also carried are no longer needed — the rebase reported them as already applied, main having landed the same fix independently. #103 passes `check fast`, which is the gate that would catch any remaining drift.

Nothing is lost by closing this.

jerry73204:
Closing: rebasing onto `main` leaves this branch **byte-identical to main** — both commits are already upstream as `313a27609` and `4acc524d6`, so `git rebase` drops them and the resulting tree hash equals `origin/main`'s. There is nothing left to merge.

Issue 0935 is `status: resolved` on main.

PR #124, opened 2026-08-31
feat(#0951): the per-target readers resolve `[image.*]` first

jerry73204:
Stacked on #123 (its two commits are the first two here and will disappear once #123 merges).

`[image.<id>]` has been the buildable unit since RFC-0065, but the readers answering per-target questions still consulted `[deploy.<t>]` alone. A workspace that finished migrating would not fail — **it would get DEFAULTS**, which is the failure mode this table exists to prevent.

## Two foundations

- **`SystemToml::image_for`** — `[image.<id>]` folded over `[image_defaults]`, in one place. `image_rmw_for` was the only image reader on this type and it hardcoded one field; this is that function with the field removed. Forgetting the base is a silent wrong answer: the block parses, the field reads `None`, and the caller falls through to a default the author already overrode workspace-wide.
- **`resolve_target` gains an image rung**, above deploy. Without it, deleting the last deploy block does not fail: the ladder returns `None`, every per-target lookup misses, and the plan quietly reverts to `x86_64` / `native` / `debug`.

## The readers

| reader | note |
| --- | --- |
| `derive_target_rtos` | decides which `[tiers.<n>.<rtos>]` sub-table is read, so a wrong answer silently applies another RTOS's scheduling parameters. **Had no direct test for two phases**; has six now. `kind` is the last rung because it cannot separate `nuttx-qemu-arm` from `nuttx-qemu-riscv`. |
| `schema_build_json` | merged per **field**, not per block — mid-migration a workspace legitimately has the board on the image and the rustc triple only on the deploy. `target`/`optimize` stay deploy-only: the first is derived from the board descriptor, the second is authored **nowhere** in this tree. |
| `codegen_system` launch fallback | now scoped to the selected target rather than "the first deploy block with a `launch`, in BTreeMap order" — arbitrary whenever several were declared. Image `launch` is relative to `launch/`; a deploy's is not. |
| `doctor` | images get their own section, delegating to `validate_image_launch` rather than re-deriving the join — a second implementation is how doctor comes to disagree with the builder about whether a workspace is healthy. |

## The quiet one

`synthesise_self_bringup` wrote `image: Default::default()` while filling `deploy` from `[package.metadata.nros.deploy.*]`. Every reader above would have read an empty map there and **lost its values for self-bringup packages**. Nothing in the tree shows that gap, because a self-bringup is synthesised rather than authored. It now projects both.

## Verification

`rm -rf <ws>/build/nros/models && nros sync` across all 31 workspaces, before and after — a plain sync serves a content-addressed cache and reads as "no change". **Model content is byte-identical.** The only diff is `system.toml` provenance hashes, which moved because the issue-number comment fix edited those files after the previous capture.

`just ci-l1` green; `just check fast` 146/146; 800 lib tests.

### DIFF
--- a/docs/issues/0951-site-config-keys-on-deploy-not-board.md
+++ b/docs/issues/0951-site-config-keys-on-deploy-not-board.md
@@ -94,6 +94,38 @@ Still open — the build half:
   destination is not uniform — the first version sent `domain_id` and `locator`
   to `[image.*]`, a table with no such keys.
 
+Found by an audit of every remaining `[deploy.*]` reader, and NOT yet fixed:
+
+* **`resolved_domain_id` / `resolved_locator` never read `[host.*]`.** The
+  deprecation lint tells users to move `domain_id` and `locator` there, and
+  `SystemToml::host` then has ZERO production readers in this repo — only the
+  upstream rlm placement resolver consumes it. A user who follows the lint's own
+  advice gets the `[system]` default silently baked into firmware
+  (`NROS_SYSTEM_DOMAIN_ID`). No in-tree fixture misbehaves today, because
+  nothing authors either key outside `[system]`; it is a hole the lint opened by
+  pointing at a table nobody reads. The wrinkle to solve first: `resolve_target`
+  returns an IMAGE id while `[host.*]` is keyed by MACHINE name, and the link
+  between them is the image's `args.host` binding — so a plain
+  `self.host.get(t)` would be wrong too.
+* **`nros new --deploy` still scaffolds into the retiring table.**
+  `scaffold_deploy.rs` writes `board` and `target` into `[deploy.<name>]` —
+  exactly the two fields the lint measured as actually firing — so the
+  scaffolder emits a workspace that immediately trips its own deprecation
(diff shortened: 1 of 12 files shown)


jerry73204:
Updated: the board-facts fix moved here from #123 (which was already in the merge queue and could not be updated).

An audit of every remaining `[deploy.*]` reader found that #123 left the gate and the resolver disagreeing — `check-site-config` learned to read `[image.*]`, but `board_facts::pick_deploys` still enumerated `[deploy.*]` alone. `examples/workspaces/realtime-rust` (seven images, zero deploy blocks) could not resolve at all, and `cmake/NanoRosBoardFacts.cmake` swallows that error, so the SDK roots would simply vanish from the cargo environment.

Testing that fix surfaced two more, both now covered:
- `--board` matched textually, while the gate canonicalises through the board catalog;
- comparing descriptors by `source` compares the descriptor FILE, and `nros-board-nuttx-qemu/nros-board.toml` declares TWO boards — so `--board nuttx-qemu-riscv` resolved the **arm** board's SDK roots. Identity is `names`, which is per-entry.

`just ci-l1` green; `just check fast` 146/146; 804 lib tests.

PR #128, opened 2026-08-31
docs(phase-400 W2): the orchestration half measured — 3.96 CPU-s, declined

jerry73204:
W2's own stated next step was *"`cargo build --timings` on this leaf attributing those 9, not a refactor."* Done. The answer is **3.96 CPU-s of 118.3 — 3.3 %** — and it does not buy the refactor. Docs-only; no code change.

## What leaves — leaf-verified, 15 crates + the 2 gated crates themselves

| | | | |
| --- | --- | --- | --- |
| 0.64s | `ros-launch-manifest-model` | 0.09s | `hashlink` |
| 0.52s | `thiserror-impl` | 0.09s | `nros-rmw` |
| 0.51s | `ros-launch-manifest-sched` | 0.08s | `arraydeque` |
| 0.47s | `encoding_rs` | 0.07s | `nros-core` |
| 0.29s | `thiserror` | 0.06s | `byteorder` |
| 0.29s | `ros-launch-manifest-types` | 0.06s | `nros-serdes` |
| 0.28s | `nros-orchestration-ir` | 0.05s | `hash32` |
| 0.26s | `heapless` | 0.01s | `stable_deref_trait` |
| 0.19s | `yaml-rust2` | **3.96s** | **TOTAL** |

## What does not leave — the whole finding

```
1.70s  zerocopy        <- "the largest single unit freed", per the original W2 write-up
0.45s  hashbrown
0.41s  serde_yaml_ng
0.32s  unsafe-libyaml
0.30s  ahash
-----
3.18s  contested — held by other requirers in the leaf
```

The standalone `nros-macros` graph credits all 3.18 s to orchestration. The leaf does not. Stopping at the workspace graph would have reported **7.1 s / 6 %** — nearly double, and wrong in the same optimistic direction as the three earlier estimates this phase already records.

## Method — two halves, because neither answers alone

| | source | figure |
| --- | --- | --- |
| which crates leave | `just leaf-graph <leaf>/rust/target --side host --exclusive-to nros-orchestration-ir --exclusive-to ros-launch-manifest-model`, on a **freshly built** leaf (the W4 stale-record rule) | 15 of 112 host crates |
| what they cost | cold `cargo build -p nros-macros --release --timings`, scratch target dir | 80 units / 53 pkgs / 18.3 CPU-s / 8.95 s wall |

**3.96 s is a ceiling, not a point estimate.** The durations come from a standalone `--release` build; the leaf's host units compile under the build-override profile and run ~4x faster — the doc's own cold leaf report has `nros-orchestration-ir` at **0.07 s** where this run has **0.28 s**. The real leaf figure is plausibly 1–2 s.

## Why DECLINED and not deferred

Against 3.96 s (ceiling): split `nros-orchestration-ir` (the cut separates **items**, not files — `lib.rs` and `executor_sizing.rs` are both on the every-arm path *and* touch the model types), a default-off `model` feature, a `launch` feature threaded through ~22 leaf manifests plus the codegen emitter, a required-feature `compile_error!` path, a `TierBake` boundary type, 13 `unused_mut` suppressions, and another round of leaf-lockfile churn under the `just lock-update`-only rule.

Recorded as declined rather than left open on purpose: *"not yet implemented"* is an invitation to recompute the same estimate a fourth time, and the estimate is not the problem — three were optimistic and this one is finally attributed.

## One correction to the existing record

The revert's table says `board_path_for`, `FRAMEWORKS`, `executor_sizing::*_SLOTS` are "needed on every arm", implying the `nros-rmw` tail is unavoidable. It is not. `qos_override::{lower_all, is_qos_override}` — the sole reason `nros-rmw`, and with it `nros-core`/`nros-serdes`/`heapless`, is in the host graph at all — sits inside the `if let Some(model_lit)` branch at `main_macro.rs:597`. That tail **is** launch-only, which is why the leaf set is 15 crates and not the 9 previously recorded. It still costs 3.96 s.

## Not done, stated so it is not assumed

The leaf's own cold `--timings` (a 1.2 GB delete plus full rebuild) was not run — the doc's stated acceptance. The ~4x ceiling argument stands in for it. And the 118.3 s baseline predates W2a, so the current leaf is nearer 110 s, moving 3.3 % to 3.6 %.

### DIFF
--- a/docs/roadmap/phase-400-rust-dependency-weight-audit.md
+++ b/docs/roadmap/phase-400-rust-dependency-weight-audit.md
@@ -1,14 +1,33 @@
 # Phase 400 — Rust dependency weight audit
 
-**Status (2026-08-30). IN PROGRESS — W1 and W2a landed. W2's orchestration half
-was ATTEMPTED AND REVERTED: its 43-crate estimate measured as 6, because
-`nros-orchestration-ir` is needed by every arm of the macro and is itself what
-pulls the heavy tail. The lever moves to splitting that crate; see W2. **W3 is
-also RETRACTED**: its 20.6 s of bindgen belongs to upstream `zephyr-sys`, and the
-four `*-sys` crates the plan named are workspace-EXCLUDED with zero consumers, so
-they contribute nothing to any measured build. W4, the leaf-graph attribution, is
-the only item left with a defensible premise — and should have been first. Waves
-are ordered by measured value, not by discovery order.**
+**Status (2026-08-31). CLOSED — the single-leaf campaign is at its floor.**
+
+LANDED: W1 (an unused dep), W2a (cbindgen off by default, 8.1 s), W4 (`just
+leaf-graph` — ask the build, not the workspace), W5 (shared cargo dirs for
+Zephyr; a subsequent image costs about half).
+
+DECLINED ON MEASUREMENT, not deferred: W2's orchestration half (3.9 CPU-s,
+3.3 %) and W3's bindgen-to-committed-output (0 s from our crates).
+
+WHAT CLOSING IT MEANS. The largest single item left on one leaf is 2.18 s of
+bindgen inside upstream `zephyr-sys` — 25 % of that leaf's cargo wall, on its
+critical path, and owned by a west module we do not control. Everything cheaper
+has landed or been priced and refused. Reopening should need a NEW measurement,
+not a new estimate.
+
+THE ONE THING NEVER MEASURED, recorded so it is not mistaken for a dead end:
+every number in this phase is ONE leaf on an idle 32-core box. Multi-leaf
+concurrency — N leaves each taking `nproc` — is what `jobserver-pool.sh` exists
+for and what nothing here has touched. That is where a win could still be, and
+it is a measurement rather than a build.
+
+HOW TO READ THE NUMBERS BELOW. Three estimates in this document were computed
+from the WORKSPACE graph and then assumed to describe a leaf's build (31.9 %,
+43 crates, 20.6 s); all three were wrong and all three were optimistic. A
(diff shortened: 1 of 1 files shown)


jerry73204:
**Correction pushed** — a claim in the description above is refuted by measurement.

The description said 3.96 s was *"a ceiling, not a point estimate"*, worth about 4x, because a leaf compiles host units under cargo's build-override defaults (`opt-level = 0`, `codegen-units = 256` — the workspace declares no `build-override`) while my measurement was a standalone `--release` build. Plausible, and wrong.

Re-run cold with `CARGO_PROFILE_RELEASE_OPT_LEVEL=0 CARGO_PROFILE_RELEASE_LTO=off CARGO_PROFILE_RELEASE_CODEGEN_UNITS=256`:

```
                        O3+LTO     O0        graph total  18.3 s -> 14.1 s
  removable set (17)     3.96 s   3.89 s     <- the lever, unchanged
  nros-macros            3.62 s   0.48 s     <- the crate that STAYS
```

Every crate in the removable set is within **0.02 s** of itself across the two profiles — small crates whose cost is frontend, not codegen, so optimisation level buys nothing. The 1.3x the graph total moves is almost entirely `nros-macros` (3.62 → 0.48 s), and `nros-macros` is the crate the gate *keeps*.

**So the lever is ~3.9 s either way.** The verdict is unchanged; it now rests on a measurement instead of an assumption. Read "3.9 s" wherever the description says "3.96 s (ceiling)" and ignore the "plausibly 1–2 s" sentence.

One number is left unreconciled rather than explained away: the cold leaf report puts `nros-orchestration-ir` at 0.07 s against ~0.27 s in both runs here. `nros-macros` agrees across the two measurements (0.48 s vs 0.59 s), so it is not a profile gap. Even at the leaf's figure the set is ~3.7 s.

PR #131, opened 2026-08-31
docs(#0958): the Cyclone RMW discards the subscription options

jerry73204:
subscription_create takes the options struct and names none of it — the parameter is spelled /*options*/ — and grep rx_buffer_hint over the cyclonedds backend returns nothing.

The field is declared in the ABI, read by rust_adapter, produced by the Rust executor since phase-392 W3a and by the C++ register variants since phase-402, then dropped here.

The consequence is worse than an unimplemented feature: every consumer-visible mechanism for sizing a receive buffer from the message type terminates in the zenoh shim, so a Cyclone consumer can do everything the campaign asks and observe nothing, with no sign that the backend is where it stopped. The autoware-safety-island an536 lane is that consumer.

Records what is NOT affected, because the two will otherwise be conflated in a measurement: the executor arena is sized by nano-ros, not the backend, so phase-408 W3 shrinks a Cyclone image regardless. Measure the arena, not the backend.

Closed by EITHER wiring the hint, or declaring it unsupported in the backend docs so consumers stop looking. Not acceptable is the present state, where it is discarded at a comment.

phase-392 W3f now points here. Docs only; just ci-fast passed.

### DIFF
--- a/docs/issues/0958-cyclonedds-ignores-rx-buffer-hint.md
+++ b/docs/issues/0958-cyclonedds-ignores-rx-buffer-hint.md
@@ -5,7 +5,7 @@ title: "The Cyclone RMW discards `rmw_subscription_options_t` entirely, so no
 status: open
 type: bug
 area: rmw
-related: [issue-0896, issue-0917, phase-392, phase-408]
+related: [issue-0896, issue-0917, issue-0969, issue-0970, phase-392, phase-408]
 ---
 
 ## What was measured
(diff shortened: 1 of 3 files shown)


jerry73204:
Amended: option 1 is now closed off, not left open.

Reading the take path establishes what the hint could have meant here — the only site is `dds_ostream_init(&os, 0, 1)`, an output stream that starts empty and grows by `realloc`. That ostream exists only because the backend deserializes wire CDR into a typed sample and re-serializes it. #0969 (PR #154) deletes the round trip, and the ostream goes with it.

After #0969 the backend owns no receive buffer to size at all: the serdata is Cyclone's, sized by the arriving sample, and the destination is the caller's arena slot, already derived from the type by phase-403 W3/W5. So the hint is **inapplicable** here rather than unimplemented — a property of the architecture, not of anyone's backlog.

Close via option 2, naming #0969 as the reason there is nothing left to size. The original complaint is unchanged: a backend that ignores a shared-ABI field must say so where it ignores it.

PR #138, opened 2026-08-31
phase-405 + phase-407: one configuration surface, and the scope you name is the specification

jerry73204:
Two campaigns, one branch: **phase-405** (one configuration surface) finished, and **phase-407** (the scope you name is the specification) landed W1–W4.

## phase-405 — one configuration surface

W1–W6 complete. The recurring find was that the doc's own claims were wrong in both directions, and measuring beat assuming:

- **W3** — "`ros_edition` has THREE defaulting sites, all humble" was six in cmake plus five in Rust, and one says `jazzy`. The gate meant to prevent this grepped `ros-humble` while every site writes bare `humble`: **it matched zero of six and printed OK for its whole life.** The six were not equivalent — `nros_find_interfaces` discarded a workspace's declared edition, measured as `"ros_edition": "humble"` for `std_msgs` under `-DNANO_ROS_ROS_EDITION=jazzy`. That is RFC-0056's wire mismatch: builds clean, fails on the wire.
- **W6** — one of the two "gated duplications" was already an end state; the other had **two checkers** on the same pair with contradictory prose. Issue 0940 turned out to be 12 unreachable site-config blocks across 7 bringups, not 1.
- **W4** — `nano_ros_entry` from eleven keywords to six. `MODEL` had zero callers.

## phase-407 — the scope you name is the specification

The rule: **named → must work; unnamed → may skip, and is always reported.** Naming *is* the specification, so there is no second declaration to drift.

- **W1** — the fixture stamp recorded the lane's *nominal* coordinates, so a module that skipped a missing prerequisite still read as covered and the run skipped its tests to a green sweep. The laundering was one line: `if [ "$have" = "all" ]; then return 0`. The build already knew, via a joblog it has written since issue 0599.
- **W2** — 21 skip sites; the decision moved into `lane-skip.sh` so 14 needed no edit. `NROS_LANE_INCLUDED` unset means *named*, so a missed site fails loudly.
- **W3** — `just <verb> [scope…]`. `just test zephyr`, not `just zephyr test`. Default scope is **probed, never stored**.
- **W4** — a CI job is `just setup <scope>` + `just test <scope>`. `just setup` could not take a preset while its own menu advertised presets.

## CI reporting defects fixed along the way

- `check-submodule-pins` **passed every PR while comparing nothing** — shallow checkout means no `origin/main`, and it `exit 0`'d. A real pin rewind reached a pushed branch and only a local run caught it.
- `post-submit` reported **success with its only expensive job skipped**, so the tier-2 matrix — every fixture build and E2E cell in that lane — had never run.
- `check-feature-set-ssot` matched a spelling no site used.

Each is the same shape: reporting success for work not done. Three new gates enforce the class.

## Issues resolved

0940, 0941, 0946, 0947, 0949, 0950, 0955. Filed and left open with evidence: 0957.

## Verification — read this part

`just check fast`: **155 gates OK** (was 143). Gate selftests **71/184** (was 57). `just ci l1`: green.

**Tier 2 was run on a frozen tree with freshly rebuilt fixtures** (8/8 modules OK). Result:

```
1795 run · 1595 passed (2 flaky) · 6 skipped
Real failures: 17          (nextest's raw "200 failed" counts skip! panics)
```

Of the 17:

- **5 are proven pre-existing** — the lane/justfile meta-tests fail *identically* on a fresh `origin/main` worktree (`CiTier::Tier1 declares 'just ci' but the justfile has no such recipe`).
- **12 are undetermined**: esp32_emulator ×4, native_api ×3, zephyr ×3, emulator, logging_smoke. They fail solo, so not load flakes. Proving pre-existing needs fully provisioned fixtures on a main worktree; that determination is running separately and this PR should not merge until it reports.

Context for why a backlog is plausible: `post-submit`'s tier-2 job has **never** run, and `host-tests` has been red for **20 consecutive runs**, dying before its tests execute. Nobody has run tier 2 in a long time — which is the thesis of the CI fixes above, now pointing at this branch.


jerry73204:
Superseded by a stack of four focused PRs — #148, #149, #150, #151.

A 36-commit branch could not converge: main took **37 commits in the last hour**, so every rebase was invalidated before the push completed, and each one cost a full conflict pass through the same files main's #0951 had already redesigned.

Split at natural seams, smallest first:

- **#148** — 0949, one `set()`, independent
- **#149** — three CI lanes that reported success for work they never did
- **#150** — phase-407 + phase-410 (stacked on #149)
- **#151** — phase-405 (stacked on #150)

Three fixes this branch carried were dropped rather than rebased: main's #0951 made `[deploy.*]` site config unrepresentable, superseding 0940 and 0950 with a better fix, and 0941 is re-opened because its resolution described code no longer in the tree.

PR #142, opened 2026-08-31
feat(phase-400 W6): the executor tenant, and a census that makes its gate real

jerry73204:
Stacked on #140.

## The gate could not be evaluated

Every wave in phase-400 is gated on "the ladder knob count rises … **measured the same way as the table at the top**", and that table records no method. Two people counting by hand get two numbers, so no wave could show it moved anything.

`just check config-knob-census` is that method — buildless, on the fast line, with a ratchet:

```
knobs in the RFC-0049 ladder       13     (8 executor · 3 zenoh.tx · 2 transport)
Kconfig CONFIG_NROS_* declarations 85
NROS_* env read by a build.rs      31
ZPICO_* env read by a build.rs      5
```

**Do not read `3 → 13` as this phase's delta** — the doc says so too. The `3` was counted by an unknown method, so the rows are not comparable term-for-term. What is now true is that the number is reproducible.

Only the **ladder** count is gated, and only downward. An env name legitimately survives migration — it stays the front-end — so "the env count falls to zero" was never the goal and gating it would punish the design. A knob *leaving* the ladder loses its platform and board rung and its `nros config explain` row, which is worth refusing.

## The executor tenant

Less was missing than the doc implied, and something else was. `ExecutorKnobs`, `resolve_executor` and both production callers already existed — I duplicated the resolver before the compiler caught it, and my first reading ("declared but no reader") was wrong: `config.rs` and `model_ingest.rs` call it, which a grep for `ExecutorKnobs` misses.

What was actually missing: **two of the eight knobs were invisible to `nros config explain`** — in a report whose entire job is to say where a value came from. `action_clients` and `arena_size` have *derived* defaults, so they did not fit the `&[(&str, usize)]` table the other six were listed in, and were left out.

Neither derivation is duplicated to fix it: `action_clients` defaults to the **resolved** `max_cbs` (build.rs then clamps to it, so the default *is* the clamp), and `arena_size` defaults to `0` — the documented Kconfig sentinel for "derive it" — printed as `derived`. `nros-node/build.rs` stays the one place that knows the arena formula.

Verified against the running tool, both halves of the wave's own gate:

```
executor.arena_size                derived    builtin  [NROS_EXECUTOR_ARENA_SIZE]
$ NROS_EXECUTOR_MAX_CBS=17 nros config explain --platform posix
executor.max_cbs                   17         env      [NROS_EXECUTOR_MAX_CBS]
executor.action_clients            17         builtin  [NROS_EXECUTOR_ACTION_CLIENTS]
```

## Recorded, not fixed

The platform/board rungs **do not reach a bare `cargo build`**. `nros-node/build.rs` resolves env → Kconfig → default and never opens the platform or board TOML, so a build outside cmake still compiles at crate defaults — the failure `ExecutorKnobs`'s own doc comment describes. Closing it is a decision about core's dependency graph (`nros-node`'s build-deps are deliberately "dependency-free"; the ladder lives in `nros-board-common`, which `nros-zpico-build` *does* depend on). The phase doc now states both options rather than picking one silently.

The census carries its own negative controls, which `check-gate-selftests` requires — and they earned their keep immediately: writing them surfaced that the matcher missed `env::var_os` and `&"NAME"`, so the first number I recorded (27) was low. It is 31.

`just ci-l1` green; `just check fast` 147/147.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1857,6 +1857,8 @@ dependencies = [
  "heapless 0.8.0",
  "libc",
  "log",
+ "nros-board-common",
+ "nros-build-paths",
  "nros-core",
  "nros-ghost-types",
  "nros-lifecycle-msgs",
--- a/docs/roadmap/phase-400-unified-configuration.md
+++ b/docs/roadmap/phase-400-unified-configuration.md
@@ -1,6 +1,7 @@
 # Phase 400 — the unified config system: build it, migrate onto it, retire the rest
 
-**Status (2026-08-30): W2-W5 and W7 done; W1, W6, W8 outstanding.** Design is
+**Status (2026-08-31): W2-W5 and W7 done; W6's first and largest tenant
+(`NROS_EXECUTOR_*`) done, its remaining tenants and W1 / W8 outstanding.** Design is
 [RFC-0086](../design/0086-unified-configuration-transport-tenant-and-coupling.md),
 which amends RFC-0049 (Stable) and adopts RFC-0071 D8. Nothing here proposes a
 new mechanism — the ladder and `nros config explain` already exist and work.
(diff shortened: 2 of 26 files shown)


jerry73204:
Updated: the deferred decision is now made and implemented.

**Option 1** — `build.rs` reads the descriptor the lane already points it at. Two obvious routes were closed first, neither for a small reason:

- **`build.rs` cannot know its platform.** `nros-node` has no `platform-*` cargo feature by design — phase-248 C2 deleted that axis so core stays platform-agnostic via the vtable. Reversing it to migrate a sizing knob would be the tail wagging the dog.
- **The lane cannot export resolved values.** That is issue 0460 exactly: cmake's `set(ENV{...})` reaches the C lane and not the cargo one.

What works is the idiom already used everywhere here — the lane exports a value and a **pointer**, the build script reads the file. Two facts made it small: the board descriptor already knows its platform, and `corrosion_set_env_vars` already attaches those vars to the target's own build command. So `board-facts` emits `NROS_PLATFORM_NAME` (not `NROS_PLATFORM` — cmake's `-DNROS_PLATFORM=cffi` names the platform *layer*), and `build.rs` resolves env → Kconfig → board → platform → builtin.

Every failure there is **fatal**, not a fall-through. My first version used `.ok()`, and a platform file with one rejected key compiled at crate defaults while reporting success — caught only because my own test fixture had that key.

Verified end to end: platform rung `max_cbs = 11`, board rung `max_sc = 23`, env still wins at `99`, malformed file panics naming the key.

Cost measured, not estimated: 18 leaf locks moved and **16 gained only the two path-dep lines** (board crates already depending on `nros-board-common`). Only `nros-verification` pulls serde/toml/cc fresh, in a host build graph.

`just ci-l1` green; `just check fast` 149/149.

PR #150, opened 2026-08-31
phase-407 + phase-410: the scope you name is the specification, and CI is breadth × depth

jerry73204:
Split out of #138. **Stacked on #149** (the CI reporting fixes) — review that one first.

Two phases that turned out to be one change: what a run is asked to cover, and what CI does with the answer.

## phase-407 — the scope you name IS the specification

> **Named → must work. Unnamed → may skip, and is always reported.**

A test that cannot run skips, and a skip is indistinguishable from a pass. That is *correct* locally — a developer on Zephyr has not provisioned FreeRTOS. The caveat is the whole issue: when someone **expects** a platform and the prerequisite is missing, the same mechanism reports success.

- **W1** — the fixture stamp recorded the lane's *nominal* coordinates, so a module that skipped a missing prerequisite still read as covered. The laundering was one line: `if [ "$have" = "all" ]; then return 0`. The build already knew, via a joblog it has written since issue 0599 — the fix is one `awk`, no new bookkeeping.
- **W2** — 21 skip sites; the decision moved into `lane-skip.sh` so 14 needed no edit. `NROS_LANE_INCLUDED` unset means *named*, so a missed site fails loudly.
- **W3** — `just <verb> [scope…]`: `just test zephyr`, not `just zephyr test`. Default scope is **probed, never stored** (a stored set drifts).
- **W4** — a CI job is `just setup <scope>` + one tier command. `just setup` could not take a preset while its own menu advertised presets.

## phase-410 — CI is BREADTH × DEPTH, and only depth is expensive

```
just ci l3 / ci matrix build   cross build + link, no boot      27–46 s
tier 2 run + warm fixtures     9.5 min + 11 min                 20+ min
```

**~25×.** So build depth gates every merge and run depth cannot.

**Run depth on `push(main)` STARVES with ten agents.** `post-submit` sets `cancel-in-progress: true` — right for "is main good now", fatal for a 20-minute lane: merges arrive faster than it completes, so every run is cancelled by the next and tier 2 **completes never**. Moved to a clock.

**Two of four declared tiers were run by nothing at all**, and `check-tier-has-ci-owner` now enforces that a tier nobody invokes is a failure, not a silent gap. It caught a coverage regression I had introduced an hour earlier (`just test tier2` runs only ¼ of what `just ci matrix` does).

**W4 — "lane" was overloaded three ways**, meaning two. `l1` was never a rung — it visits no coordinates — so it is now `just ci gate`. Depth is positional because **`just` does not parse `name=value` for a module recipe** (measured: yields the literal string).

## Two finds underneath

`check-lane-contracts` **could not see private recipes at all** — its pattern required `^[a-z]`, so every `_`-prefixed recipe was invisible, including `_lane-gate` which `ci matrix` calls. Its closure stopped silently at each private boundary. Widened; still clean, now across 3 tiers instead of 2.

Six lane guards **could never fire** (issue 0955): `[ -z "${NUTTX_DIR:-}" ]` where `sdk-env.just` exports all 23 such variables with a default. They break the *unusual* way — an intended skip became a cmake failure.

## Not verified

The interlocked jobs cannot run until a self-hosted runner is registered. `just check fast`: 155 gates OK; `ci l1`/`ci gate`, `ci matrix build` and the scope routes exercised locally.

### DIFF
--- a/.github/workflows/build-wide.yml
+++ b/.github/workflows/build-wide.yml
@@ -1,0 +1,81 @@
+# BUILD + LINK, wide — the mandatory promise (phase-410 W2).
+#
+# CI is BREADTH x DEPTH, and only DEPTH is expensive:
+#
+#   breadth  which coordinates a run visits   tier1 tier2 nightly full   low cost
+#   depth    what it does with each           build+link | build+run     HIGH cost
+#
+# MEASURED 2026-08-31: `just ci l3` — cross build + link plus ELF symbol
+# interrogation, no QEMU and no tests — is 46 s. The tier-2 RUN over the same
+# tree is 9.5 min on top of an 11 min warm fixture rebuild. Depth is ~25x
+# breadth, so build depth can afford to gate every merge and run depth cannot.
+#
+# "It compiles and links for every target" is the regression that hurts most and
+# costs least to catch. That is this lane.
+#
+# WHY push(main) AND NOT the merge queue
+#
+# The queue's budget is the required `CI` context, deliberately cheap so ten
+# agents can land (phase-395). This runs AFTER a merge, where its job is to
+# bound how far a link regression travels — one commit, not one nightly.
+#
+# WHY cancel-in-progress IS correct here and is NOT correct for run-depth
+#
+# A newer commit's answer subsumes an older one's, and at 46 s this run
+# FINISHES before the next merge arrives. A 20-minute run-depth lane under the
+# same setting starves: with ten agents landing through a queue that batches
+# four, each merge cancels the last and the tier completes never — a lane that
+# always cancels looks busy while reporting nothing. Run depth is on a clock,
+# in the `run-*` workflows.
+name: build-wide
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: build-wide-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  l3:
+    name: build + link (cross ELFs, no boot)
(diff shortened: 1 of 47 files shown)


jerry73204:
**The 12 undetermined tier-2 failures are settled: they are not from this PR.**

Proved by diff rather than argued. Across every tree that feeds a built image — `packages/`, `examples/`, `zephyr/`, `integrations/`, `cmake/` — this branch changes exactly **two files**, both in `packages/testing/nros-tests/src/` (test harness, compiled into no target image):

```
packages/testing/nros-tests/src/buckets.rs | 58 +++++++++-------
packages/testing/nros-tests/src/ci_lane.rs | 22 ++++-----
```

So the esp32, zephyr, threadx and QEMU images built here are byte-identical to main's, and a runtime difference cannot originate in this branch.

Filed as **0968** with the list, the proof, and **no root cause claimed** — one sample was examined (threadx cyclone service: client shows `Locator:` empty, server produced no output at all, i.e. never started), but whether that is one cause or several is unknown. Issues 0859–0862 were all filed from a sweep in this repo and all four retracted, two carrying confident wrong causes; that issue deliberately stops at the list.

Context for why it went unnoticed: `post-submit`'s tier-2 job has **never run** (interlocked on an unset variable, and a skipped job does not colour its run), and `host-tests` has been red for 20 straight runs on 0967 — fixed in #149. Between them, no fixture-backed test has run in CI for a long time.

Also worth stating: 5 of the original 17 real failures were lane/justfile meta-tests, and **this PR fixes them** — the tier ladder had drifted from the justfile and its own gate could not see module recipes at all.

PR #162, opened 2026-08-31
fix(codegen): the C++ header states the DERIVED bound, or no size at all

jerry73204:
Fixes #964.

## The problem

The C++ pack computed `SERIALIZED_SIZE_MAX` with an *estimator*: a flat 512 per
nested message, a flat default capacity per string, and it ALWAYS returned a
value, so it could never say "unbounded".

Measured over 126 types in 12 stock ROS Humble packages:

* 86 have no derived bound, and the header stated a size for every one of them.
* Across the 40 that ARE bounded, the estimate matched the derived bound
  **zero times**.

So one type carried two numbers. phase-403 W6 exports the derived bound on three
transports while the header went on advertising the estimate, and the estimate is
the one a user reads, because it is what the constant is named.

It also breaks phase-380's rule head-on: a number nobody chose, substituted where
the honest answer is "no bound exists". That rule is why an unbounded type is a
build error at all, and this path opted out of it silently.

The cost was not hypothetical. The island's receive buffer, arena and payload
classes were budgeted for `Control` at 2052 bytes when it serializes to 114.

## What changed

`SERIALIZED_SIZE_MAX` is now `nros_serdes::size::max_serialized_size` -- for
messages, service request/response and action goal/result/feedback alike -- and a
type with no bound states no size.

The C pack already did this. The two packs now derive it through ONE funnel
(`generator::common::derive_header_bound` -> `bounds::header_bound`), which also
owns the budget check and the poison identifiers, so the two headers and the
exported inventory cannot drift into disagreeing about whether a bound exists,
what it is, or what the hole is called when there is none.

The poison has a different SHAPE in C++, because C++ cannot express the C one. A
`#define X <undeclared identifier>` costs nothing to include and errors when
NAMED; a `static constexpr size_t X = TOKEN;` errors at the point of DEFINITION,
which would break every consumer of the header rather than only the ones asking
for a size. The C++ form is a static data member of an INCOMPLETE type
([class.static.data] permits a non-defining static-member declaration to have
incomplete type), which keeps both properties, and the diagnostic prints the
identifier, so it names the type and the member that costs it the bound.

Both halves are compiler-checked, not asserted in prose:
`tests/cpp_bound_compile_check.rs` compiles the generated header under
`-Wall -Wextra -Werror`, then compiles a use of the constant and greps the error
for the token. g++ and clang++ both name it.

The estimator survives for exactly one job and is renamed to say so.
`types::storage_serialized_size` sizes the FFI publish glue's stack buffer for a
type that has NO bound -- a question about the generated struct's fixed inline
storage, not about the ROS type -- and it is emitted into no header, exported on
no transport, and reachable by no user. A BOUNDED type's publish buffer is now
the derived number, so the flat-512 under-count is gone wherever a bound exists.

The C++ drivers also gained the cross-package nested lookup the C drivers got in
W6, so a C++ type that nests across packages reports a real bound instead of
`Unresolved`.

## Second commit: the two consumed types

The first commit makes an example that names `SERIALIZED_SIZE_MAX` for an
unbounded type stop compiling. Two types are affected, across 21 call sites in
15 files:

| type | reached via |
| --- | --- |
| `std_msgs/String` | `Subscription<M>::try_recv` |
| `example_interfaces/Fibonacci` Result + Feedback | `get_result`, `Stream::try_next`, `publish_feedback`, `complete_goal` |

Both are CONSUMED, not defined here, so there is no `.msg` to bound and a cap in
`nros-codegen.toml` is the only remedy. That remedy only became reachable when
`nros_find_interfaces` and `_nros_generate_declared_interfaces` learned to
forward `CODEGEN_CONFIG` (#152, merged) -- before that, a consuming package could
not bound a stock type at all.

## Measured, before -> after

C++ header:

| type | before | after |
| --- | ---: | ---: |
| `autoware_control_msgs/Control` | 2052 | 114 |
| `autoware_vehicle_msgs/SteeringReport` | 527 | 24 |
| `autoware_adapi_v1_msgs/OperationModeState` | 572 | 27 |
| `nav_msgs/Odometry` | 1804 | none (unbounded) |
| `nav_msgs/Odometry`, both `frame_id` fields capped at 64 | 1612 | 880 |

12-package stock corpus: 126 types stating a size -> 40, and all 40 now equal the
derived bound (0 of 40 did before). The other 86 state none.

Verified end to end on `examples/native/cpp/listener`: it builds, and the
generated `std_msgs/msg/String` header states

    static constexpr size_t SERIALIZED_SIZE_MAX = 269;

which is the 256-byte cap plus CDR framing. The cap set the BOUND, not just the
storage container. I checked the value rather than the exit code, because a cap
keyed on a name that does not exist does nothing and says nothing.

## Goldens

The eight fingerprint-corpus `.hpp` files moved, deliberately. Nothing else did:
the C `.h` goldens are byte-identical, which is the point, and the C++ ones now
agree with them type for type. Regenerate with
`NROS_UPDATE_GOLDEN=1 cargo test -p rosidl-codegen --test codegen_golden`.

## Gate status

Rebased onto `main` at 97dc38946, clean. Three gates are red on this branch and
all three are red on untouched `main`, verified by running them in a worktree at
that same commit:

* `fixtures-manifest` and `kconfig-overridden-values` -- both `nros: command not
  found`. Green once the CLI is on PATH.
* `cli-tests` / `a_missing_framework_names_how_to_supply_it` -- fails identically
  on `main`.

Note that `cli-tests` is inside `check::build`, and a red step withdraws every
step after it, so a bare `just ci l1` on this branch never reaches the C++
compile. Run `just build-examples` with `NANO_ROS_ROOT` pinned to the checkout to
get that coverage.

### DIFF
--- a/cmake/NanoRosMessageBounds.cmake
+++ b/cmake/NanoRosMessageBounds.cmake
@@ -6,8 +6,9 @@
 # them, so every size downstream was still a number a human typed -- and on the
 # one bring-up that measured it, every one of those numbers was wrong at least
 # once. `NROS_MAX_LARGE_SUBSCRIBERS` and `NROS_SUBSCRIBER_LARGE_SIZE` were read
-# off generated C++ headers BY EYE, and the headers they were read off state an
-# ESTIMATE rather than a bound (W6's own finding).
+# off generated C++ headers BY EYE, and the headers they were read off stated an
+# ESTIMATE rather than a bound (W6's own finding; fixed by issue 0964, so a C++
+# header now states the same derived number this module composes).
 #
 # This module composes the fragments and derives the FOUR knobs a bound
 # inventory can actually answer.
@@ -508,8 +509,9 @@ function(_nros_message_bounds_write_output _path _status _reason _ceiling)
     string(APPEND _c "# resolved SystemModel carries today (phase-403 W4).\n")
     string(APPEND _c "#\n")
     string(APPEND _c "# Derivation: nros_serdes::size::max_serialized_size, the same rule the\n")
-    string(APPEND _c "# runtime's M::MAX_SERIALIZED_SIZE_XCDR* uses. NOT the C++ pack's\n")
-    string(APPEND _c "# SERIALIZED_SIZE_MAX, which is an estimate (phase-403 W6).\n")
+    string(APPEND _c "# runtime's M::MAX_SERIALIZED_SIZE_XCDR* uses -- and, since issue 0964,\n")
+    string(APPEND _c "# the same rule the C AND C++ headers state, so nothing here can\n")
+    string(APPEND _c "# disagree with a generated header about the same type.\n")
     string(APPEND _c "#\n")
     string(APPEND _c "# small/large class split: ${_ceiling} B (policy -- ZPICO_SUBSCRIBER_SIZE_THRESHOLD)\n")
     string(APPEND _c "\n")
--- a/docs/issues/0964-two-different-sizes-for-the-same-type.md
+++ b/docs/issues/0964-two-different-sizes-for-the-same-type.md
@@ -63,3 +63,100 @@ estimate, so the receive buffer, arena and payload classes were all budgeted for
 (1) matches what the C pack does today and what phase-380 requires. It is a
 behaviour change with a blast radius across every C++ consumer, which is why it
 is filed rather than done.
+
+## LANDED -- option 1, in codegen
+
+`SERIALIZED_SIZE_MAX` is now `nros_serdes::size::max_serialized_size`, for
+messages, service request/response and action goal/result/feedback alike. The C
+pack and the C++ pack derive it through ONE funnel
+(`generator::common::derive_header_bound` -> `bounds::header_bound`), so the two
+headers and the exported inventory cannot disagree about whether a bound exists,
+what it is, or what the hole is called when there is none.
+
+Re-measured over the same 12 stock packages (126 types by this crate's parser,
+not the 120 the filing quotes; the phase-403 W7 table already says 126):
+
+| | before | after |
+| --- | ---: | ---: |
+| types stating a size | 126 | 40 |
+| types with NO derived bound | 86 | 86, and each now states NO size |
+| bounded types where the stated number equals the derived bound | 0 of 40 | **40 of 40** |
+
+Reference types, C++ header before -> after:
+
+| type | before | after |
+| --- | ---: | ---: |
+| `autoware_control_msgs/Control` | 2052 | 114 |
+| `autoware_vehicle_msgs/SteeringReport` | 527 | 24 |
+| `autoware_adapi_v1_msgs/OperationModeState` | 572 | 27 |
+| `nav_msgs/Odometry` | 1804 | none -- unbounded |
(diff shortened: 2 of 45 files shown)


jerry73204:
Gate note, after chasing the example matrix to the end.

`just build-examples` does NOT pass on this branch. It also does not pass on
`main`, and the failure is identical. Measured in the SAME provisioned checkout,
with a freshly built (non-stale) CLI and a clean `bootstrap.sh doctor` on both
sides:

| run | failing example targets | crates.io `nros` resolution errors |
| --- | ---: | ---: |
| `fix/cpp-derived-bound` | 15 | 15 |
| `main` @ 97dc38946 | 15 | 14 |

The failing target sets are the same 15. Set-difference branch-minus-main is
EMPTY, so this branch introduces no example failure. Every one of them is a Rust
example failing to resolve `nros` from the crates.io index, which is the dropped
`[patch.crates-io]` entry described in issues 0363 / 0197 and tracked by #0966 /
PR #153. Zero of the failures are C++.

Stated plainly so nobody reads more into the gates than is there: because the
matrix dies in the Rust examples, **the C++ example compile never runs, on either
side**. This PR's C++ codegen change therefore has no example-matrix coverage. It
is covered by `tests/cpp_bound_compile_check.rs` (which compiles a generated
header under `-Wall -Wextra -Werror` and greps the poison diagnostic under both
g++ and clang++) and by the end-to-end check on `examples/native/cpp/listener`
quoted in the description, and not by the matrix.

Three earlier attempts at this comparison were invalid and are worth recording so
the next person does not repeat them:

1. Branch-built CLI against a `main` worktree -- the stale-CLI guard correctly
   fired, because the CLI was fresh for the branch and stale for `main`.
2. A separate `git worktree` for `main` -- it has its OWN `target/`, so it lacked
   `nros-launch-resolve` and died before reaching the example step.
3. Anything before `bootstrap.sh doctor` was clean -- ninja, make and
   `play_launch_parser` were all missing, and each one surfaced as a different
   error at a different stage, which reads as a cascade of real bugs rather than
   one unprovisioned environment.

Only a detached `main` in the same fully provisioned directory is a valid
control.

jerry73204:
Closing as **superseded by `5f3c08545`** (phase-408, #0896), which landed the same change on main: the C++ pack emits the REAL derived bound, absent for an unbounded type, and spends it on the receive buffer. The estimator this PR set out to remove is gone from the user-facing constant.

Confirmed by reading main rather than inferring: `publish_buffer_size()` in `generator/cpp.rs` now takes `derived: Option<usize>` and falls back to `storage_serialized_size()` only for the unbounded case, and its doc comment cites the same defect this PR's description does — one type carrying two numbers.

A rebase is not the right move here. All 15 files conflict, including the `fingerprint-corpus` expected-output fixtures, because both sides rewrote the same generated values; resolving those means deciding which codegen output is correct, which is an authoring decision rather than a merge one.

If anything in this PR is *not* covered by `5f3c08545` — the `cargo-nano-ros` change, or the issue-0964 doc edits — it is worth reopening as a narrow follow-up against current main instead.

Issue #964 should be checked for whether it is now resolved by that commit.

PR #168, opened 2026-08-31
docs(phase-400 W6, phase-349 W2): the zenoh tenant is another campaign's

jerry73204:
Stacked on #165.

I set out to do W6's zenoh tenant and phase-349 W2 together. **Neither turned out to be the work I expected**, and both findings are worth more than the code would have been.

## The zenoh tenant is not W6's to migrate

Second family in a row where a *derivation* campaign owns the number.

- `ZPICO_MAX_QUERYABLES` is **already** a checked override over a derived default (phase-392 W5.f: "stops being an independent opinion")
- `ZPICO_MAX_SESSIONS` is posed by that phase explicitly — *"either it joins the model or it stays a knob and this phase says so"*
- `SERVICE_BUFFERS` is a product of the two

Giving them a platform/board rung now would answer phase-392's question the wrong way: a rung gives a global a per-platform default, and that phase is deciding whether these stop being globals at all. Exactly the buffer lesson (phase-403), one family over.

The census now records the **owning campaign** in the reason column. W6's own backlog is **23, not 31**, and its clearest remaining tenant is the platform heap/stack trio — genuine platform facts no campaign is deriving.

Stated once in the doc, so it isn't rediscovered a third time: *"migrate the long tail into the ladder" was right for the executor tenant and is the wrong default for the rest. Check for an owning campaign before migrating a family.*

One mislabel corrected: `NROS_MAX_{ARRAY_LEN,BYTE_ARRAY_LEN,STRING_VALUE_LEN}` are **parameter** value bounds (`nros-params/build.rs`), not message bounds.

## phase-349 W2 is unblocked

Its note's central claim — "nothing anywhere sets `NROS_BOARD_TOML`" — has been false since phase-351. `board_facts.rs` sets it; `corrosion_set_env_vars` delivers it to the target's own build command. #142 then proved the channel end to end in `nros-node/build.rs`, using the very `watch_path` hazard that note names.

`NROS_NETSTACK` is emitted too and **nothing reads it** — the same declared-but-unread shape, now with a writer and no consumer instead of the reverse.

**What I did not do, and why.** W2's remaining work is not the channel but the *selection site*. `LinkPolicy::freertos_lwip()` hardcodes FreeRTOS ⇒ lwIP and the upstream port row exists (`zenoh-pico/src/system/freertos/freertos_plus_tcp/`) — but a port's C sources are not chosen from a Rust source list (`add_common_c_sources` adds only `system/common` and has no caller for the platform dir), so the selection runs through zenoh-pico's own CMake. I looked in the wrong place first and the note now says where not to, rather than my half-implementing a port change across build paths I hadn't located.

## Also: a red on main that `check fast` cannot see

`check-action-client-arena-budget.py` walks build directories for a compiled binary beside the knob it was compiled against — untracked by construction, which is the case `no-tracked-file-find`'s own text carves out. It needed the `# walk-ok:` marker. That gate is in the compile tier, so **`ci-l1` was red on main while the fast line was green**.

`just ci-l1` green; `just check fast` 158/158.

### DIFF
--- a/docs/roadmap/phase-349-rtos-integration-shells.md
+++ b/docs/roadmap/phase-349-rtos-integration-shells.md
@@ -1,7 +1,9 @@
 # Phase 349 — RTOS integration: make FreeRTOS an imported library like the rest
 
-**Status (2026-08-13). W1 LANDED — with it, [phase-348](archived/phase-348-source-time-provider-discovery.md)
-is complete. W2–W6 open.**
+**Status (2026-09-01). W1 LANDED — with it,
+[phase-348](archived/phase-348-source-time-provider-discovery.md) is complete.
+W2 is UNBLOCKED (its prerequisite exists now; see its note) but not started.
+W3–W6 open.**
 
 **Implements:** [RFC-0072](../design/0072-rtos-integration-nano-ros-is-a-guest.md).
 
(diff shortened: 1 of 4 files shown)


jerry73204:
Superseded by #174, which contains these commits rebased onto main. Consolidated because the stack had drifted: #169 merged underneath, and keeping three branches meant three rebases of the same content.

PR #170, opened 2026-08-31
feat(phase-400 W6): `[knobs.memory]` — the platform heap and stack

jerry73204:
Stacked on #168 (which is stacked on #165). Needs **#169** for a green `ci-l1` — that PR fixes a red already on main.

The tenant the re-scope in #168 identified as the clearest one left: numbers that genuinely vary by platform and board and that **no derivation campaign owns**, unlike the zenoh caps (phase-392) or the rx/tx buffers (phase-403).

`NROS_FREERTOS_HEAP_KB` and `NROS_FREERTOS_APP_STACK_KB` now resolve over the full ladder — builtin < platform < board < env — and print in `nros config explain`.

## Stored in bytes, always

The front-ends disagree about units: the FreeRTOS pair is KiB, `NROS_ZEPHYR_HEAP_SIZE` is bytes. A table where "heap" means one thing on one platform and another elsewhere is a unit bug waiting to be written, so the ladder has one unit and each front-end converts at its rung, beside the table recording which spelling it has.

Verified through the running tool:

```
memory.heap_bytes       2097152   builtin   [NROS_FREERTOS_HEAP_KB]
memory.heap_bytes       1234567   platform  [NROS_FREERTOS_HEAP_KB]   # platform file
NROS_FREERTOS_HEAP_KB=4096 -> 4194304   env  # KiB front-end, bytes rung
```

## The env-pointer dance is now written once

`BuildRungs::from_build_env()` reads `NROS_PLATFORM_NAME` + `NROS_BOARD_TOML`, watches the file's **content** (issue 0491, not the path spelling), and fails **fatally** rather than falling through to defaults.

`nros-node/build.rs` grew its own copy when the executor tenant landed — two build scripts resolving the same rungs differently is exactly the drift `check-knob-single-reader` exists to catch one level up. The shared version is the one to use, and the doc says that copy should adopt it.

Both readers keep their behaviour when no lane names a platform: the FreeRTOS heap still leaves FreeRTOSConfig.h's 3 MiB alone unless something below the ladder has an opinion, because "no default" is not a number a rung can invent.

## 2 of the trio, not 3

`NROS_ZEPHYR_HEAP_SIZE` is read by `option_env!` in `nros-platform`'s source with its own build-script resolver, so it needs that crate to take the `nros-board-common` build-dep — the same decision this wave already made for `nros-node`, and worth making deliberately rather than in passing.

Ladder 13 → 15; W6's backlog 23 → 21.

`just check fast` 158/158. `ci-l1` fails only on main's `no-tracked-file-find` red; with #169's one-line fix applied locally it is green.

### DIFF
--- a/docs/roadmap/phase-349-rtos-integration-shells.md
+++ b/docs/roadmap/phase-349-rtos-integration-shells.md
@@ -1,7 +1,9 @@
 # Phase 349 — RTOS integration: make FreeRTOS an imported library like the rest
 
-**Status (2026-08-13). W1 LANDED — with it, [phase-348](archived/phase-348-source-time-provider-discovery.md)
-is complete. W2–W6 open.**
+**Status (2026-09-01). W1 LANDED — with it,
+[phase-348](archived/phase-348-source-time-provider-discovery.md) is complete.
+W2 is UNBLOCKED (its prerequisite exists now; see its note) but not started.
+W3–W6 open.**
 
 **Implements:** [RFC-0072](../design/0072-rtos-integration-nano-ros-is-a-guest.md).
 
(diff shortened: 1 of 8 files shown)


jerry73204:
Added a commit restoring **W7, W8 and "What this phase does NOT do"** to the phase doc.

I deleted them by accident in #165: the census commit replaced the remaining-tenants section with `s[s.index('### Remaining tenants'):]`, which runs to **end of file**. That landed on main.

Nothing caught it — no gate reads a phase doc's wave list, and the wave *table* at the top still names W8, so the file contradicted itself in a way only someone looking for W8 would notice. I was that reader, one wave later, asking what was next.

Recovered verbatim from `caa634aff~1` and diffed byte-for-byte. The narrow lesson: an anchored replacement needs *both* ends — `s[a:]` is a deletion wearing the costume of an edit.

jerry73204:
Superseded by #174, which contains these commits rebased onto main. Consolidated because the stack had drifted: #169 merged underneath, and keeping three branches meant three rebases of the same content.

PR #177, opened 2026-08-31
feat(phase-400 W8): the ladder's membership decides what the one-reader gate checks

jerry73204:
Stacked on #174.

W8's gate is *"for every migrated knob, exactly one reader"*. It existed and passed — over a **hand-kept list of eight, while the ladder held fifteen**. Its own docstring named the hole:

> a knob that is in the ladder but not here is simply unchecked, which is why W6 and W8 move together

…and then relied on whoever migrated a knob remembering to add a row. Three waves didn't.

So the list is no longer kept there. It comes from the census's `ladder` class; this file supplies only the **owner**, and a knob reaching the ladder without one now fails. Two hand-kept lists of "what's in the ladder" is the same duplicate-fact shape the ladder exists to remove.

## It found real work immediately

Seven unowned knobs, and in two of them a **second reader**:

- **`nros-zpico-build`'s no-platform fallback re-read the tx trio.** Three migrated knobs with two readers that could resolve different values — precisely issues 0135 and 0316. It also reported `KnobSource::Env` for values that came from **builtins**, so provenance was wrong in the one place provenance *is* the product. Replaced by `platform_config::tx_env_only`, the same role `executor_env_only` already had.
- **`shim_config_from_env` read `ZPICO_TX_BATCH` and `_FLUSH_MS`** — and had them overwritten eleven lines later by the resolved values. Dead work that looked authoritative: harmless while the overwrite stays put, a silent disagreement the moment it moves.

A `derived` knob may still keep an owner — `NROS_SUBSCRIPTION_BUFFER_SIZE` is on the ladder as the fallback for a type with no declared bound while phase-403 takes over its value. What the gate refuses is an owner for a knob the census doesn't know at all.

**14 migrated knobs, one reader each.**

## Correcting myself

I said this gate "would have caught the duplicate I created". **It would not** — that duplicate was the env-*pointer* dance (`NROS_PLATFORM_NAME`, `NROS_BOARD_TOML`), which are infra, not migrated knobs. This gate catches duplicate readers of ladder *values*, a narrower thing. The argument for doing W8 first was weaker than I made it; the wave earned its place on what it actually found.

`just ci-l1` green; `just check fast` 158/158. Pool inventory regenerated (line drift).

### DIFF
--- a/book/src/reference/static-pool-inventory.md
+++ b/book/src/reference/static-pool-inventory.md
@@ -62,9 +62,9 @@ its byte cost yet. Absence of a figure is not a claim that it is free.
 | `NROS_SMOLTCP_SOCKET_TIMEOUT_MS` | computed — see `packages/drivers/net/nros-smoltcp/build.rs:41` | `packages/drivers/net/nros-smoltcp` |
 | `NROS_SUBSCRIPTION_BUFFER_SIZE` | 1024 | `packages/core/nros-node` |
 | `NROS_XRCE_STREAM_HISTORY` | computed — see `packages/rmw/xrce/nros-rmw-xrce-cffi/build.rs:259` | `packages/rmw/xrce/nros-rmw-xrce-cffi` |
-| `ZPICO_BATCH_MULTICAST_SIZE` | computed — see `packages/rmw/zenoh/nros-zpico-build/src/runner.rs:482` | `packages/rmw/zenoh/nros-zpico-build` |
-| `ZPICO_BATCH_UNICAST_SIZE` | computed — see `packages/rmw/zenoh/nros-zpico-build/src/runner.rs:481` | `packages/rmw/zenoh/nros-zpico-build` |
-| `ZPICO_FRAG_MAX_SIZE` | computed — see `packages/rmw/zenoh/nros-zpico-build/src/runner.rs:480` | `packages/rmw/zenoh/nros-zpico-build` |
+| `ZPICO_BATCH_MULTICAST_SIZE` | computed — see `packages/rmw/zenoh/nros-zpico-build/src/runner.rs:488` | `packages/rmw/zenoh/nros-zpico-build` |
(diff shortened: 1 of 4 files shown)


jerry73204:
## The merge queue is frozen repo-wide — do NOT re-queue

This is not a property of this PR. `just queue-triage` reports `queue` failing in the merge group for **8 different pull requests**, and #177 was added and ejected three times (19:22, 20:01, 20:21 UTC) without a code failure. Its `gate` run passes every time.

**Root cause is the self-hosted interlock, not any change.** `queue.yml`'s `L3 (cross build + link)` job fails in `runner-doctor`:

```
[MISSING] Zephyr SDK 0.16.8 is not at .../scripts/zephyr/sdk/zephyr-sdk-0.16.8
[MISSING] no arm-zephyr-eabi-gcc under .../zephyr-sdk-0.16.8
          the SDK directory exists but its toolchains are not unpacked —
          an interrupted install leaves exactly this state.
runner-doctor: FAIL — at least one label claims something this host does not have.
```

`vars.NROS_SELF_HOSTED_READY` is currently `true`. `queue.yml`'s own header says that variable may only be turned on **after** `just runner-doctor <labels>` passes on the machine — it is on while the runner cannot back its `nros-sdk-zephyr` label, so L3 runs, fails, and takes every batch with it. The last successful merge was 19:53 UTC; everything queued after it has been ejected.

**Two remedies, maintainer's call:**

1. Provision the runner (preferred — keeps the coverage):
   `scripts/ci/runner-provision.sh nros-sdk-zephyr` on the runner host, then `just runner-doctor nros-sdk-zephyr,nros-big`.
2. Drop the interlock until then: `gh variable set NROS_SELF_HOSTED_READY --repo NEWSLabNTU/nano-ros --body false`. The queue is then carried by the hosted job alone — a real gate, just narrower, which is the degraded mode the workflow was written to have.

Not flipping it from here: it changes the merge gate for every contributor, and option 1 restores full coverage rather than reducing it.

An issue file would be the right home for this, but it cannot land while the queue is frozen — hence this comment.

PR #182, opened 2026-08-31
feat(phase-400 W6): the params tenant, and a census that had been under-counting by 29

jerry73204:
Stacked on #177 — review that first; this PR's own diff is the single commit `1ba14271b`.

## The `params` tenant

The five `NROS_MAX_*` **parameter-storage** bounds (not message bounds — that mislabel is corrected in the census) move onto the RFC-0049 ladder: `ParamKnobs`, `[knobs.params]` in a platform or board toml, `BuildRungs::param_rungs()`, and one reader in `nros-params/build.rs` composing env → Kconfig → rung → builtin.

```
$ nros config explain --platform zephyr
params.max_parameters              32         builtin  [NROS_MAX_PARAMETERS]
...
$ NROS_MAX_PARAMETERS=77 nros config explain --platform zephyr
params.max_parameters              77         env      [NROS_MAX_PARAMETERS]
```

Checked for an owning campaign first, per the lesson that has bitten this wave twice. The only doc hit is a usage *record* — the ASI consumer's `NROS_MAX_PARAMETERS=256` living in its own `build.sh` — which is the argument **for** a rung, not against one.

## The census had been under-counting by 29

Wiring the gate exposed a defect in the measuring instrument, and it is the more important half of this PR.

The census found readers by matching a fixed list of helper **names** (`env_usize`, `env::var`, …). Wrapping these five reads in a local `knob()` took them out of the count **while `--check` stayed green** — the gate only fails on a name it *sees* and cannot classify.

Matching *any* call is the opposite error: `.define("ZPICO_X", ..)` emits a C macro and `.with_env(..)` sets a **child's** environment; counting those added 67 phantom names. So the matcher now carries `READ_CALLEES` and `NON_READ_CALLEES` and **fails on a callee in neither** — the same "a new knob forces a decision" rule this census already applied to knobs, applied to the idioms that read them.

That surfaced **29 names** read through `req` / `list` / `env_get` / `flag`:

| | count | |
| --- | --- | --- |
| infra | 18 | board descriptor facts, include and source dirs |
| **sizing** | **11** | RMW static pools, smoltcp driver pools + timeouts, XRCE stream depth, zpico ring depth |

**W6's backlog is 28, not 17.** It went up because the instrument stopped lying. Two candidate tenants fall out of it — an `rmw` one and a `net` one — neither owned by another campaign.

## `gen-pool-inventory` had the same blind spot

Same cause, one tool over: it rendered the five defaults as `computed — see build.rs:25`, a churning line number in place of the figure the table exists to publish. Its own comments already record this regression twice (`knob`, then `env_usize_min`). The ladder shape puts the builtin **third**, so it needs a third pattern; both `knob` spellings now have selftest probes.

## Verification

- `just check fast` — 158/158 OK
- `just ci l1` — passed (compile + unit)
- `just check pool-inventory` / `config-knob-census` / `knob-single-reader` / `gate-selftests` — OK
- 18 leaf locks regenerated via `just lock-update` for the new build-dep edge; diff reviewed — `+ "nros-board-common",` and nothing else, no registry packages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1895,6 +1895,7 @@ name = "nros-params"
 version = "0.5.0"
 dependencies = [
  "heapless 0.8.0",
+ "nros-board-common",
  "nros-ghost-types",
  "nros-zephyr-build",
 ]
--- a/docs/roadmap/phase-400-unified-configuration.md
+++ b/docs/roadmap/phase-400-unified-configuration.md
@@ -1,7 +1,8 @@
 # Phase 400 — the unified config system: build it, migrate onto it, retire the rest
 
-**Status (2026-08-31): W2-W5 and W7 done; W6's first and largest tenant
-(`NROS_EXECUTOR_*`) done, its remaining tenants and W1 / W8 outstanding.** Design is
+**Status (2026-09-01): W2-W5, W7 and W8 done; W6's `executor`, `transport`,
+`zenoh.tx`, `memory` and `params` tenants done, its remaining tenants and W1
+outstanding.** Design is
 [RFC-0086](../design/0086-unified-configuration-transport-tenant-and-coupling.md),
 which amends RFC-0049 (Stable) and adopts RFC-0071 D8. Nothing here proposes a
 new mechanism — the ladder and `nros config explain` already exist and work.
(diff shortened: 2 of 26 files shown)


jerry73204:
Stacked on #177, which is blocked by a repo-wide merge-queue freeze — see https://github.com/NEWSLabNTU/nano-ros/pull/177#issuecomment-5484160860. The unprovisioned `nros-sdk-zephyr` runner fails `runner-doctor` in every merge group. Not a defect in either PR; do not re-queue until the runner is provisioned or `NROS_SELF_HOSTED_READY` is turned off.

PR #151, opened 2026-08-31
phase-405: one configuration surface — the gates that matched nothing

jerry73204:
Split out of #138. **Stacked on #150** — review #149 → #150 → this.

phase-405 W1–W6, plus issues 0946, 0947, 0957. The recurring find: the doc's own claims about how many times a fact was spelled were wrong **in both directions**, and measuring beat assuming every time.

## W3 — the ROS edition had six defaulting sites and a gate that matched none

The brief said "THREE sites, all `humble`". Measured: **six in cmake, five in Rust**, and one says `jazzy` (the docker peer axis — correctly, it names the peer not the image).

`check-feature-set-ssot.sh` has always declared *"the edition is chosen in exactly one place"* — while grepping for `ros-humble`, the **cargo-feature** spelling. Every site writes bare `humble`. **It matched zero of six and printed OK for its whole life.**

The six were not equivalent: two consulted `NANO_ROS_ROS_EDITION`, four went straight to the literal. Measured on `cpp-port-minimal-publisher` with `-DNANO_ROS_ROS_EDITION=jazzy`, pinning `NANO_ROS_ROOT` at the pre-fix commit:

```
pre-fix   std_msgs, builtin_interfaces  →  "ros_edition": "humble"
post-fix  both                          →  "jazzy"
```

`nros_find_interfaces` was discarding a workspace's declared edition — RFC-0056's wire mismatch, which builds clean and fails on the wire.

*(An earlier run of that comparison appeared to show no defect. It was wrong — the worktree resolved `NANO_ROS_ROOT` back to the main checkout, so both sides ran the fixed cmake.)*

## W6 — one "gated duplication" was already finished, the other had two checkers

"SDK roots exist four times" — no. The **value** lives once; the rest is a generator, its output, and readers. Closed as already-correct, zero files changed.

The zenoh tx trio had its direction declared since phase-290 *and* **two checkers** — an 80-line Rust test and a 140-line python gate, same three pairs, same two files, contradictory prose. Collapsed to one, with `--write` so the mirror is generated rather than asserted.

## 0947 — NuttX respects the ROS edition

Two vocabularies that could not express each other: the CMake lane read a string symbol NuttX's Kconfig never declared (so it was permanently `humble`), the Make lane read a bool choice. `jazzy` existed on neither. One Kconfig choice + a derived string; measured by *evaluating* the Make fragment, since `patsubst` would otherwise silently yield `ros-"jazzy"`.

## 0946 — the locator had two producers

[...]

PR #184, opened 2026-08-31
fix(queue): the two files every PR appends to stop being conflict sites

jerry73204:
Measured which paths actually collide instead of guessing: test-merged **and** test-rebased every dirty branch against `main` and read the conflicted paths.

**`docs/issues/open.md` was not among them** — 0884's `merge=union` is working, under rebase as well as merge. Two other things are.

## 1. `just/check.just`'s gate registries

`fast-serial` and `build` name every gate as a dependency, packed up to fifteen names per line, with new gates appended at the tail. **22 of the last 31 commits touching this file edited such a line.** PR #151 conflicted on this file and nothing else; four open PRs touch it.

Union merge isn't available here — `.gitattributes` says so outright ("NEVER apply this to an authored file"), and a union of two recipe bodies is both of them.

So: **one name per line, sorted.** Two gates with different names then land at different, usually non-adjacent lines and git merges them unaided.

This costs nothing, because the order was never load-bearing: `fast` runs these concurrently, and `run-gates-parallel.sh` `sort -u`s the list it derives from this very block. *A list whose consumer sorts it cannot have been ordered on purpose.*

Proven with a controlled experiment rather than asserted — two branches each adding one differently-named gate, same scenario on both formats:

```
OLD (packed tail line)     -> CONFLICT: just/check.just
NEW (sorted one-per-line)  -> CLEAN
```

The derived gate set is unchanged; the only delta is the new gate itself, which landed at its alphabetical position mid-file rather than on the tail — the property working as intended. New gate `check-gate-lists` keeps it that way, with a selftest in both directions.

## 2. `docs/issues/README.md`'s stale second copy of the index

0884 moved the generated open-issue list to `open.md`, and `check-issue-index` reads `open.md` — but the `BEGIN GENERATED` block was **left behind in README.md**, regenerated by nothing and verified by nothing. It had drifted to **46 rows against `open.md`'s 59** while two open PRs were editing it.

A stale index is worse than no index: it reads as current. Deleted, replaced by a pointer; `check-issue-index` now refuses its return, since a generated block inside an authored file cannot take the union merge that makes `open.md` safe. Guard verified to fire and to clear.

## 3. A doc-ref red that was blocking L1 for everyone

phase-410 W1 renamed `pr-checks.yml` → `gate.yml`; this page still cited the old name, so `check-ci-doc-workflow-refs` was red on `main`. The paragraph argued *for* keeping the old name, so it needed more than a substitution — it now records the rename and why it was still right.

## Not fixed here

- PR #162's 15 conflicts are genuine parallel work on the same codegen surface; its `expected/*.hpp` are golden files whose resolution is "regenerate".
- **`main` is red in `cli-tests`** (`rosidl-codegen` golden: `RX_MAX_SERIALIZED_SIZE` 133 vs 136), reproduced on a pristine `origin/main` checkout. Last regenerated by `5f3c08545` (phase-408 #0896), with nothing landing after it in `rosidl-*` — so it merged with goldens not matching its own generator, invisible to the merge gate because `check-build` isn't on the required set. Left alone deliberately: it's the active area of PR #162 and regenerating goldens under another agent is not mine to do.

## Verification

- `just check fast` — **160/160 OK**, nothing skipped
- `just ci l1` — fails **only** on the pre-existing `cli-tests` golden red above, which reproduces on pristine `main`
- Sweep for the same shape tree-wide: only these two blocks have it.


jerry73204:
**Sequencing note for whoever is rebasing the dirty PRs.**

This reformats the `fast-serial` / `build` dependency blocks wholesale, so it will conflict **once** with any in-flight branch that adds a gate. That is the cost of removing the hotspot, and it is paid a single time.

Cheapest order: land this first, then rebase the others — each one's conflict resolution is then "put my gate name on its own line at its sorted position", which is mechanical and won't recur. Resolving in the other order means re-resolving the same block twice.

If a rebase hits the block, `just check gate-lists` will tell you whether the result is well-formed before you push.

PR #191, opened 2026-09-01
fix(#0982): two codegen tests were red on main, behind the fast lane

jerry73204:
`cargo test -p rosidl-codegen` fails on **clean `origin/main`**, in two places,
from one cause.

`BoundState::classify` gained a transport-framing allowance on the RX bound —
`transport_framed(n) = n.next_multiple_of(4)`, so a 133-byte type states 136. TX
stays exact: we write what we serialise, and the framing underneath is the
transport's business. The C emitter took the rule. Two consumers did not move
with it:

| consumer | had | correct |
| --- | ---: | ---: |
| C golden `.h` | 136 | 136 ✓ |
| **C++ golden `.hpp`** | **133** | 136 |
| **`message_size_bound_parity`** | **`max(x1,x2)`** | `transport_framed(max(x1,x2))` |

That is issue 0896's own defect — one rule with a second copy — twice more: along
the **language** axis (C right, C++ golden stale) and along the **test** axis (an
expectation recomputing what an emitter computes).

## Which side was wrong is measured, not assumed

The C golden already said 136 for the same types. Regenerating the C++ golden
makes both packs agree, which is 0896 layer 2's stated invariant — so the
generator was right and the golden stale, not the reverse. Verified per type:

```
configured/Bounded   C=136  C++=136  AGREE
configured/Capped    C=160  C++=160  AGREE
inline/Bounded       C=136  C++=136  AGREE
```

Diff is RX only: three files, two lines each (`RX_MAX_SERIALIZED_SIZE` and the
`rx_size_bound` template that mirrors it).

The parity test now calls `bounds::transport_framed` rather than a local
`next_multiple_of(4)`. A third copy of the rule, living in a test, is the same
defect wearing test clothes.

## Why nobody saw it

`cli-tests` **does** run these and **is** registered — in the `build` list.
`just check` runs `check::fast` first and stops at the first failure, so one red
fast gate silently withdraws every lane behind it.

Issue 0952 recorded that mechanism for backend **builds**. This is the first
instance of it hiding a test **lane** — and during an ABI break, a red fast gate
is the normal state.

## Deliberately not fixed here

The ordering. A withdrawn lane should report **SKIPPED** rather than nothing —
the same "no verdict is not a pass" principle as #0975 — but that is a CI-shape
decision, not a codegen one. Recorded in the issue instead of bundled in.

## Verification

- `cargo test -p rosidl-codegen`: green
- whole `cli-tests` lane (`--workspace`): green
- `just check fast`: 164 gates OK

## How it was found

Reading issue 0896 to see what remained of phase-392's W3, then running its tests
instead of trusting its text — which had already been stale twice in the *other*
direction on that same issue (the arena hint and the `_with_info` gap were both
described as open and both closed).


jerry73204:
Superseded by #0981 (`7e4a7a731`), which landed the same two fixes — the regenerated C++ golden and `transport_framed` in the parity test — while this was in flight.

Closing rather than rebasing, because 0981's diagnosis is better and mine is partly wrong. I recorded the cause as `just` stopping at the first failure, so a red fast gate withdraws `check::build` behind it. That is true LOCALLY. The operative reason these stayed red is the one 0981 found: `check-build` is `schedule`/`workflow_dispatch` only, so no PR and no merge group ran them at all. 0981 also measured the cost (8.1 s cold, 0.1 s warm) and moved both tests onto the fast line, which is the fix.

Landing this would have added a duplicate issue with the weaker explanation. Issue id 0982 released.

PR #193, opened 2026-09-01
docs(#0983): RETRACTED — a 21-minute `check fast` was another session's load, not a slow gate

jerry73204:
`just check fast` took **21 minutes**, ~95 % of it one gate:

```
check-fast (parallel): 161 gate(s) ran at -P32, 3 SKIPPED;
  slowest action-client-arena-budget 1257311ms
```

For scale, the slowest gate in two other fast-lane runs on this host: `rmw-ret-sign` **11.7 s**, `leaf-lockfiles` **17.3 s**.

## The gate is not slow — that is the finding

Run alone on the same tree at the same moment:

```
SOLO 7.23 s      rc=0      (192 advisory images)
```

**7.23 s solo against 1257 s in the lane — a factor of ~174.** The recipe's own documented expectation (*"~18 s on a fully built 100 GB dev tree"*) is therefore **correct**; it was measured for a gate run by itself.

The mechanism is contention, not the algorithm. `find_image_roots()` walks the repo for cargo profile dirs — I/O-bound, stats every directory it considers — while 31 other gates hammer the same disk.

**This does not ask for the walk to be removed.** It is legitimate and correctly marked `# walk-ok:`: the artifacts are untracked by construction, so `git ls-files` cannot see them, which is the carve-out `check-no-tracked-file-find` explicitly makes. The defect is **placement plus concurrency**.

Same "solo vs in-sweep" shape CLAUDE.md records for QEMU lanes ("six nuttx lanes failed 3/3 in-sweep, passed solo") — appearing for the first time in a *gate* rather than a test.

## Why it matters

`check fast` is what `pre-push` runs and what CLAUDE.md tells every agent to run before pushing. At 21 minutes that instruction stops being affordable — and an instruction nobody can afford gets followed selectively, which CLAUDE.md names as worse than a smaller one followed honestly.

It has already cost something concrete: a fix for #0975 was rebased, verified locally, and pushed; the 21-minute wait meant another session fixed and merged the same PR in the meantime, and the push re-created a branch that had been deleted on merge (cleaned up, no harm). The wasted work is minor — the lost race is the point.

## Recommendation

Move it to `check build`, where the other artifact-inspecting gates already live (`no-tracked-file-find`). One line, and arguably where it belongs: an image-inspecting gate can only say anything on a tree with built images, so it is not a source-level check. Two alternatives (a dedicated concurrency slot; narrowing the roots) are recorded in the issue.

## What was NOT established

The **174× rests on one observation** and should not be acted on as a ratio until someone re-times the lane. The 7.23 s solo figure is solid; the in-lane number is a single sample. Also unknown: whether other walking gates degrade the same way under `-P32` — if they do, this is a lane-runner problem rather than one gate's.

### DIFF
--- a/docs/issues/0983-slow-arena-budget-gate-dominates-fast-lane.md
+++ b/docs/issues/0983-slow-arena-budget-gate-dominates-fast-lane.md
@@ -1,0 +1,97 @@
+---
+id: 983
+title: "`check-action-client-arena-budget` costs 7 s alone and 21 MINUTES inside
+  `check fast` — a filesystem walk against 160 concurrent gates, and it makes
+  the pre-push lane unaffordable"
+status: open
+type: bug
+area: build
+related: [issue-0900, issue-0196]
+---
+
+## Symptom
+
+`just check fast` took **21 minutes**, essentially all of it one gate:
+
+```
+check-fast (parallel): 161 gate(s) ran at -P32, 3 SKIPPED;
+  slowest action-client-arena-budget 1257311ms
+```
+
+For scale, the slowest gate in two other fast-lane runs on this same host, days
+apart: `rmw-ret-sign` **11.7 s** and `leaf-lockfiles` **17.3 s**. So this one
+gate is ~70x the previous worst and ~95 % of the lane's wall time.
+
+## The gate is NOT slow. Measured.
+
+Run alone on the same tree, same moment:
+
+```
+$ /usr/bin/time python3 scripts/check-action-client-arena-budget.py
+check-action-client-arena-budget: 192 image(s) carry arena they cannot use (ADVISORY)
+SOLO 7.23 s      rc=0
+```
+
+**7.23 s solo against 1257 s in the lane — a factor of ~174.** Nothing about the
+gate's own work changed; only what else was running. That also matches the
+recipe's own documented expectation ("~18 s on a fully built 100 GB dev tree"),
+so the author measured it correctly and the number is right — *for a gate run
+by itself*.
+
+This is the "solo vs in-sweep" shape CLAUDE.md already records for QEMU lanes
+("six nuttx lanes failed 3/3 in-sweep, passed solo"), appearing for the first
+time in a GATE rather than a test.
+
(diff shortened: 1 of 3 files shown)


jerry73204:
**Retracted, and the PR is rewritten to say so.** The headline claim is refuted by re-measuring.

I filed this on ONE observation. Two clean re-runs of the same lane on the same tree:

```
run 1   21 min     slowest action-client-arena-budget 1257311ms
run 2   13 s       (my instrumentation broke sccache — see below)
run 3   12.88 s    161 gate(s) at -P32, slowest rmw-ret-sign 11601ms
```

**The lane is ~13 seconds, and the arena gate is not even its slowest member.** The 174× is not reproducible, so the recommendation this PR carried — move the gate off the fast line — is **withdrawn**. Nothing should be changed on the strength of it.

**What actually happened:** several agent sessions build in this repo concurrently. During run 1 another session was compiling and a fixture build had just written a ~100 GB tree, so the page cache was cold. The lane was timing someone else's load, not the gate.

That is the same error this repo already records as issues 0859–0862 — four ghost issues filed from artifacts built against a different tree, each with a confident wrong root cause. The issue's own "what was NOT established" section said the ratio rested on a single sample and should not be acted on until re-timed; that caveat is why this is a correction rather than a discovery.

**Measurement note kept in the doc:** the run-2 instrumentation set `TMPDIR` to a long scratchpad path, exceeding `SUN_LEN` for sccache's unix socket (`path must be shorter than SUN_LEN`) and failing `leaf-lockfiles`. Changing the environment to observe it changed the result.

**The one actionable residue, kept:** an audit of the fast lane found **30 of its 167 gates** use a filesystem-walking primitive — 19 `find(1)`, 5 `grep -r`, 3 `os.walk`, 4 `rglob`/`glob('**')` — all concurrent at `-P32`. Not a defect at 13 s, but a standing sensitivity to I/O pressure. The useful instruction for the next person: **measure the machine before blaming a gate.**

`status: open → wontfix`, moved to `archived/`, original report kept struck because the reasoning is the part worth reading.

PR #198, opened 2026-09-02
ci: run `check cli-tests` on both merge-gating events

jerry73204:
`cli-tests` lived only in `check-build`, which is `schedule` / `workflow_dispatch` only — so **no event on the path to `main` ran the CLI sub-workspace's own tests**, and a red there could land and stay.

It did. Issue 0896's C++ RX bound shipped with a golden *and* a parity expectation that both contradicted the emitter, and `main` carried those two reds until someone ran `just ci l1` locally. Same shape as issue 0319 (a backend suite with no lane, red on main for two days), one workspace over.

## Not by adding a job name to the required set

CLAUDE.md forbids that, for a measured reason: one aggregator `CI` is required, and a check that produces no verdict blocks **forever** — which deadlocked two PRs on 2026-08-28.

This adds a **step to the `check` job**, whose result the `CI` aggregator already consumes (`needs: [changes, check, scaffold-journey, colcon-parity, sdk-index]`). The required-set configuration is untouched, so there is no new context that can hang.

## Affordable — the only argument this lane accepts

- **34 s warm**, and the job already builds the nros CLI two steps up, so the compile is incremental.
- **Resolves no fixture**: no SDK, no cross toolchain, no QEMU. That's the `check-lane-contracts` rule for a merge-gating lane; the gate now sees **4** such invocations instead of 3 and passes.
- **Needs no ROS** — a property of the suite, not an assumption I made. `cargo-nano-ros`'s integration tests read `AMENT_PREFIX_PATH` and print `[NO-ROS] … skip` when it's unset, because this sub-workspace is documented to run under a plain `cargo test`. Verified by running the whole workspace with `AMENT_PREFIX_PATH`, `ROS_DISTRO`, `CMAKE_PREFIX_PATH` and `LD_LIBRARY_PATH` unset and a minimal `PATH`: **zero failures**.

## On `merge_group` as well as `pull_request`

Unlike `compile-smoke`, which is PR-only because `check-build` would do strictly more in the queue. But `check-build` is **not** on the merge group (phase-396 W1), so for this suite the queue would otherwise run nothing. "Cheap and deterministic" is exactly what that lane is documented to accept.

**Not** added to `check fast`: that runs on every push via the `pre-push` hook, and 34 s there is a cost paid on every push to buy a guarantee only the merge path needs.

## Verification

- `just check lane-contracts` — OK, and it now covers this step
- `just check fast` — 164/164
- This PR's own `check` job is the real test: it runs the new step on `pull_request`, and again on `merge_group` when it lands.

### DIFF
--- a/.github/workflows/gate.yml
+++ b/.github/workflows/gate.yml
@@ -530,12 +530,16 @@ jobs:
       # fixture: no SDK, no cross toolchain, no QEMU — the
       # `check-lane-contracts` rule for a merge-gating lane.
       #
-      # WHAT IT COSTS, measured on this runner rather than extrapolated from a
-      # warm local tree: 6 m 15 s for the tests plus 2 m 07 s to provision
-      # `nros-launch-resolve`. The local figure was 34 s, and quoting it here
-      # would have been wrong by an order of magnitude — a warm incremental
-      # `cargo test` is not what a runner does. The CLI build two steps up does
-      # NOT amortise this: that builds `--bin nros`, while this compiles the
+      # WHAT IT COSTS, measured on this runner in both cache states, because a
+      # single number is what made the first two estimates wrong:
+      #
+      #   cold (empty sccache):  6 m 15 s tests + 2 m 07 s provisioning
+      #   warm:                  1 m 32 s tests +    17 s provisioning
+      #
+      # The local warm-tree figure was 34 s; quoting THAT here would have been
+      # an order of magnitude out, and quoting only the cold one would have
+      # over-stated the steady state by four. The CLI build two steps up does
+      # NOT amortise either: it builds `--bin nros`, while this compiles the
       # sub-workspace's test targets.
       #
       # That is real latency on the path to landing, and it buys the guarantee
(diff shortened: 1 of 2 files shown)


jerry73204:
**Measured on the runner — correcting the cost claim in the description.**

The `34 s` above was a local warm-tree figure. On CI:

| step | time |
| --- | --- |
| Build nros-launch-resolve | 2 m 07 s |
| `just check cli-tests` | 6 m 15 s |

~8 min of runner time on the path to landing, an order of magnitude off my estimate. The CLI build two steps up does **not** amortise it: that builds `--bin nros`, while this compiles the sub-workspace's test targets. Both the workflow comment and CLAUDE.md now carry the measured figure instead.

**CI also caught a precondition I missed.** The first run failed with `nros-launch-resolve not built at … — run just setup-launch-resolve`. My local no-ROS run proved the suite is independent of the *environment* and said nothing about its *artifact* preconditions — my tree already had that helper built. Fixed by having the lane build what it resolves, which is `check-lane-contracts`' own rule, and reproduced in both directions locally before pushing.

If ~8 min on every PR *and* every merge group is more than you want to pay, the cheap variant is dropping `merge_group` from the `if:` and keeping `pull_request` — say the word and I'll trim it. Keeping both is what makes the guarantee hold for the merged state.

PR #205, opened 2026-09-02
fix(phase-392 W5.g): the entity facts were read before anything produced them

jerry73204:
W5.g found no `"queryable table sized from the declaration"` line in any of three
workspaces, and could not explain why `mixed` was silent when its umbrella was
demonstrably created.

One ordering inversion explains all three rows: **the consumer ran before every
producer, on every configure.**

## Cause

`nros_synth_runtime_umbrella` called `nros_entity_facts_env` inline, under a
comment asserting the accumulation *"is complete only after the SUBDIRS loop
above has processed them all"*. True of the **node** packages. False of the
**entry**, which `nano_ros_add_entry` declares last by design —
`NanoRosEntry.cmake` says so in its own words, *"the first point in a configure
that is guaranteed to be AFTER every `nano_ros_node_register()`"* — and that is
where `nros_record_entity_facts` runs.

Traced on a `mixed` reconfigure rather than reasoned about:

```
line 17:     facts_env(nros_ws_runtime-static) seen=<empty>   <- consumer
lines 19-85: model OK: system_model.yaml, ... (7 models)      <- producers
```

## Fix

Defer to the **top-level** scope — the same idiom and reasoning as
`_nano_ros_support_schedule_flush` one module over, including its warning that
deferring to the *current* directory fires at the end of whichever scope called
first, and so is the bug rather than a smaller version of it.

The line now prints on `mixed`, and one `zpico-sys` unit compiles
`ZPICO_MAX_QUERYABLES = 8` where it was 32.

## Delivery is PARTIAL — no saving claimed

After a full rebuild:

| cargo root | value |
| --- | ---: |
| `nros_ws_runtime_*` (got the env) | **8** |
| `nros_ws_runtime_*` (second unit) | 32 |
| `nano-ros_*` (repo-root, 4 units) | 32 |

One of six units took it. The others are separate cargo units under a different
workspace root — issue 0616's shape, where `-C metadata` keys a unit by the path
it was reached by. **Sizing one of six is not a number to quote**, so this
reports slots and measures no bytes.

The application half also remains absent — *"application count undeclared (no
model here describes wiring)"* is issue **0973**: only the infra figure is ever
delivered while no model describes wiring.

## Verification

`just check fast`: 167 gates OK. The status line was confirmed present after the
fix and absent before, on a real reconfigure of `examples/workspaces/mixed`.

### DIFF
--- a/cmake/NanoRosRuntimeCrate.cmake
+++ b/cmake/NanoRosRuntimeCrate.cmake
@@ -306,11 +306,29 @@ nros_armv8r_cflags_env(nros_ws_runtime-static)
     # cargo from the workspace root).
     nros_board_facts_env(nros_ws_runtime-static)
     # phase-392 W5.c — and the entity figures the RMW sizes its queryable table
-    # from. Applied HERE, once, because this staticlib is shared by every entry
-    # in the configure and the accumulation is complete only after the SUBDIRS
-    # loop above has processed them all.
+    # from.
+    #
+    # DEFERRED to the end of the TOP-LEVEL scope (phase-392 W5.g), because the
+    # comment this replaced was wrong about the ordering and the mechanism has
+    # therefore never run anywhere. It said the accumulation "is complete only
+    # after the SUBDIRS loop above has processed them all" — true of the NODE
+    # packages, false of the ENTRY, which `nano_ros_add_entry` declares LAST by
+    # design ("the first point in a configure that is guaranteed to be AFTER
+    # every nano_ros_node_register()"). `nros_record_entity_facts` runs there.
+    #
+    # MEASURED on `examples/workspaces/mixed`: the consumer logged
+    # `seen=<empty>` at configure line 17 and the seven models were recorded at
+    # lines 19-85. It read the accumulator before anything filled it, every
+    # time, in every workspace — which is why W5.g found no status line in any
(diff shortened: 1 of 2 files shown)


jerry73204:
**Correction pushed (`f28692c11`): the `32 -> 8` claim is withdrawn.**

This PR said one `zpico-sys` unit "compiles `ZPICO_MAX_QUERYABLES = 8` where it was 32". Not established. Two reasons:

- **8 is also the non-hosted default.** `queryable_default_from` returns `UNDECLARED_HEADROOM` when nothing is delivered and the target is not hosted, so observing 8 does not by itself show the env arrived.
- **The before/after built the same thing twice.** The revert silently did not apply — `git stash push` created no entry and the `||` fallback therefore never ran — so both halves built WITH the fix. Both reported `SERVICE_BUFFERS` = 36,032, and identical numbers looked like a clean control when they were two runs of one configuration.

**What stands unchanged** is the ordering defect and its fix, which is what this PR is for: the consumer logged `seen=<empty>` ahead of all seven producers, the status line was absent on every configure before and present after. That was traced directly, not inferred.

**What is withdrawn** is any byte or slot saving. The phase doc now says the delta is unmeasured instead of implying one.

Caught by building both sides — which is also why the mistake is a correction here rather than a number someone quotes later.

PR #214, opened 2026-09-02
docs(phase-412): the campaign to stop hand-picking counts and sizes

jerry73204:
phase-403 proved an image can derive its own buffer sizes; phase-409 did the same for the `Executor` value. This is the campaign that finishes the job: no hand-picked count or size in a board `.conf` that the image already knows.

Opened from an audit of the mr-canhubk344 island. Four knobs now derive. The rest split cleanly, and the split is the point of the phase.

## Six are already computed, and nothing reads them

The entity inventory publishes eight per-kind counts every configure. Exactly ONE becomes a knob (`NROS_DERIVED_EXECUTOR_MAX_CBS`). No new arithmetic is needed for the rest -- a consumer and a precedence entry each:

| knob | island hand-set | published input | derives to |
| --- | ---: | --- | ---: |
| `NROS_MAX_SUBSCRIBERS` | 12 | `COUNT_SUBSCRIPTION` | 10 |
| `NROS_RMW_SUBSCRIBER_SLOTS` | 12 | `COUNT_SUBSCRIPTION` | 10 |
| `NROS_MAX_PUBLISHERS` | 16 | `COUNT_PUBLISHER` | 14 |
| `NROS_MAX_QUERYABLES` | 4 | `COUNT_SERVICE_SERVER` | 0 |
| `NROS_EXECUTOR_MAX_NODES` | 6 | `INVENTORY_COMPONENT_COUNT` | 4 |
| `NROS_EXECUTOR_ACTION_CLIENTS` | 0 | `COUNT_ACTION_CLIENT` | 0 |

This is the defect the repo keeps producing: correct, tested, unreachable from a real build. `rx_buffer_hint` sized nothing (0896). The bound inventory had no reader (0963). A cap could not reach codegen (#152). Now eight counts with one consumer. All four were found by an audit rather than a gate.

## Two questions are blocking, not advisory

`NROS_MAX_SUBSCRIBERS` and `NROS_MAX_QUERYABLES` are zenoh SESSION limits, not executor ones -- the shim may declare entities the image never registers as callbacks. And a component is not always a node, so `COMPONENT_COUNT` may not equal `MAX_NODES`.

Both must be answered by reading the code before wiring, because derived values carry no headroom by design (phase-403's rule, which makes the running image a checker of its own declaration). That property is only worth having if the count is right; getting it wrong under-counts, and under-counting halts the board.

## Six have no derivation, each named with its blocker

The arena (40960 B, the largest remaining number) is blocked on **phase-403 step 2**: QoS depth MULTIPLIES the bound -- 86108 B at depth 10 against 24516 at depth 1 for the same ten subscriptions -- and the entity record carries no depth. Defaulting is the worst option, since ROS's default of 10 inflates tenfold and 1 under-sizes unsafely.

`CONFIG_MAIN_STACK_SIZE` needs frame analysis rather than an inventory; **phase-409** established the method and never turned it into a knob. The heap has no static model. The graph cache and liveliness are sized by the PEER graph and are probably never derivable from this image.

Work owned by other phases is NAMED here, not duplicated -- phase-403 owns steps 2 and 3 and this phase consumes them.

## W4: the gate that should have caught this

Assert the inverse of the current knob gates: not "is the derived value right" (they check that) but "is every published input READ by something". A new count then arrives with either a consumer or a deliberate exemption naming why.

## Known

Everything here inherits issue #0991's first-configure lag, so every measurement must come from a build that configured at least twice. Worth fixing before W1 multiplies it by six.

Gates: `just check fast` 168/168, no skips.

### DIFF
--- a/docs/roadmap/phase-412-derived-counts-and-sizes.md
+++ b/docs/roadmap/phase-412-derived-counts-and-sizes.md
@@ -1,0 +1,152 @@
+# Phase 412 — every count and size the image can state about itself
+
+**Status (2026-09-02). Opened from the mr-canhubk344 island audit.** phase-403
+proved an image can derive its own buffer sizes; phase-409 proved the same for
+the `Executor` value. This phase is the campaign that finishes the job: no
+hand-picked count or size in a board `.conf` that the image already knows.
+
+## Why a campaign rather than another knob
+
+The island's board `.conf` was, until this week, twenty-odd numbers a person
+chose. Every one of them is a claim about the image, and every one can be wrong
+in two directions. Too small halts the board; too large costs RAM on a part
+that has 320 KiB. Neither failure names the knob that caused it, and none of the
+numbers carries the reasoning that produced it.
+
+Four of them now derive (phase-403 W8/W9). The audit that followed found the
+rest split cleanly into two groups, and the split is the point of this phase:
+
+* **The number is already computed and published, and nothing consumes it.**
+  Six knobs. No new arithmetic — a consumer and a precedence entry each.
+* **No derivation exists yet.** Six more, each blocked on something specific
+  and nameable.
+
+The first group is the recurring defect this repo keeps producing: a mechanism
+that is correct, tested, and unreachable from a real build. `rx_buffer_hint`
+sized nothing (0896). The bound inventory had no reader (0963). A cap could not
+reach codegen (#152). The entity inventory publishes eight per-kind counts and
+exactly one of them becomes a knob (this phase). Each was found by an audit, not
+by a gate, which is why W4 below exists.
+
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #217, which carries this exact commit as its first of two and then implements W1 on top.

Closing rather than merging so the phase doc and its first wave land together: #217 amends this doc with things W1 only learned by doing -- both blocking questions answered from the code, the measurement (17170 bytes, RAM headroom 0.87% -> 4.76%), the rule that the `-1` DERIVE sentinel is safe only where the consumer supplies its own default, and three delivery failures that argue for a stronger W4 than this version specifies.

No content is lost. Nothing here needs re-filing.

PR #217, opened 2026-09-02
phase-412 W1: the session pools derive from the entity inventory

jerry73204:
Supersedes #214 (the phase doc), which is folded in here as the first commit.

The entity inventory published eight per-kind counts every configure and exactly ONE became a knob. This wires four more.

| knob | island hand-set | derived |
| --- | ---: | ---: |
| `NROS_MAX_SUBSCRIBERS` | 12 | 10 |
| `NROS_RMW_SUBSCRIBER_SLOTS` | 12 | 10 |
| `NROS_MAX_PUBLISHERS` | 16 | 14 |
| `NROS_MAX_QUERYABLES` | 4 | 0 |

    RAM   324834 (99.13%) -> 312088 (95.24%)   -12746 B
    DTCM   93744 (71.52%) ->  89320 (68.15%)    -4424 B

17170 bytes, and RAM headroom goes from 0.87% to 4.76%. Every derived value matches an independent enumeration of the four components' sources.

## Both blocking questions were answered by reading the code first

**No shim addend.** `ZenohSubscriber::new` has exactly one caller (`create_subscription`); the graph cache lives in its own `graph_cache.sub` field and liveliness tokens in `liveliness[ZPICO_MAX_LIVELINESS]`, so neither shares the subscriber pool.

**But the raw per-kind count was still the wrong input**, which is why the question was worth blocking on. A declared action is ONE entity that costs several session slots -- a server 3 queryables and 2 publishers, a client 1 subscription. `ACTION_SERVER_PUBLISHERS` and `ACTION_CLIENT_SUBSCRIPTIONS` join `ACTION_SERVER_QUERYABLES` beside the calls that decide them, and `check-infra-queryable-counts` now holds all three. I verified the new checks can FAIL rather than pass silently: setting `ACTION_SERVER_PUBLISHERS` to 3 reports the drift, and restoring it goes green.

**`NROS_EXECUTOR_MAX_NODES` is deliberately NOT wired.** A component is one node, but the bridge creates TWO nodes per bridge outside the component model, so `COMPONENT_COUNT` is a lower bound. Under-counting halts the board; it moves to W2 with that blocker named.

## The rule this cost

**The `-1` DERIVE sentinel is safe only where the consumer supplies its own default.** Rung 4 deliberately leaves a knob UNRESOLVED so a Rust build script falls to its own literal -- "the one place that literal is written". A C compile definition has none, so the unresolved knob expands to nothing and `zpico.c` reports `flexible array member not at end of struct` on a struct nobody edited. W2 converts five more knobs onto that sentinel, so the rule is in the phase doc.

## Three delivery failures, none in the arithmetic

The derived values were right at every step -- 10, 10, 14, 0 sat correctly in the fragment throughout -- and **every gate stayed green through all three**, because the gates check the inventory and the resolver and all three were downstream of both:

1. **A second consumer** -- the zpico C defines read raw `CONFIG_*`, bypassing the resolver, and ran before it.
2. **A name that resolved to empty** -- a `foreach` built `NROS_DERIVED_NROS_MAX_SUBSCRIBERS`, which names nothing; CMake yields EMPTY rather than failing.
3. **A consumer with no default** -- the rule above.

W4 as first written ("every published symbol has a consumer") catches NONE of them. The doc now specifies the form that does: assert per knob that the value reaching the COMPILE equals the value the resolver produced. Three instances, not an argument.

## Gates

`just check fast` 168/168. Pushed with hooks enabled -- and the pre-push guard caught a stale `zenoh-pico` submodule pin rewind that `git add -A` had swept in, which is exactly what it is for.

Note the island needs TWO configures (issue #0991); every measurement above is from the second.

### DIFF
--- a/cmake/NanoRosEntityInventory.cmake
+++ b/cmake/NanoRosEntityInventory.cmake
@@ -238,6 +238,8 @@ function(nros_derive_entity_inventory_knobs)
     _nros_entity_publish(NROS_ENTITY_INVENTORY_REASON "")
     _nros_entity_publish(NROS_ENTITY_INVENTORY_COMPONENT_COUNT 0)
     foreach(_v NROS_DERIVED_EXECUTOR_MAX_CBS NROS_DERIVED_EXECUTOR_ACTION_CLIENTS
+               NROS_DERIVED_MAX_SUBSCRIBERS NROS_DERIVED_RMW_SUBSCRIBER_SLOTS
+               NROS_DERIVED_MAX_PUBLISHERS NROS_DERIVED_MAX_QUERYABLES
                NROS_ENTITY_INVENTORY_ENTITY_TOTAL)
         unset(${_v})
         unset(${_v} PARENT_SCOPE)
@@ -334,6 +336,15 @@ function(nros_derive_entity_inventory_knobs)
         _nros_entity_publish(NROS_DERIVED_EXECUTOR_ACTION_CLIENTS
             "${NROS_DERIVED_EXECUTOR_ACTION_CLIENTS}")
     endif()
+    # phase-412 W1 -- the SESSION pools, republished on the same terms: present
+    # only when the whole image declared, absent otherwise, so a consumer reads
+    # a derived value or reads nothing.
+    foreach(_pool NROS_DERIVED_MAX_SUBSCRIBERS NROS_DERIVED_RMW_SUBSCRIBER_SLOTS
+                  NROS_DERIVED_MAX_PUBLISHERS NROS_DERIVED_MAX_QUERYABLES)
+        if(DEFINED ${_pool})
+            _nros_entity_publish(${_pool} "${${_pool}}")
+        endif()
+    endforeach()
     foreach(_kind PUBLISHER SUBSCRIPTION TIMER SERVICE_SERVER SERVICE_CLIENT
                   ACTION_SERVER ACTION_CLIENT GUARD_CONDITION)
         if(DEFINED NROS_ENTITY_COUNT_${_kind})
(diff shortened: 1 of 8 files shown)


jerry73204:
## Correction, and W4

The measurement in this PR's description was WRONG and is corrected below. Read this before the body above.

W1 as first pushed derived the right numbers and **did not deliver them**. The inventory said 10/14/0 and the zenoh session was compiled with **8/8/8** -- eight subscriber slots for ten subscriptions. That fails at RUNTIME, not at link, and every gate in the tree called it green.

Two causes, both found by the new W4 gate rather than by review:

- **A loader whitelist.** `_nros_load_derived_entity_inventory` exports a NAMED list. A symbol the fragment sets and the list omits is loaded and then dies at the function boundary, so the knob falls to rung 4 and a default that may be SMALLER than the demand.
- **A double resolution.** `NROS_RMW_SUBSCRIBER_SLOTS` was resolved again further down with the raw Kconfig value, which is the `-1` DERIVE SENTINEL. The second call won, and the sentinel reached the build as though someone had stated it.

Both fixed. The island now delivers what it derives:

    -DZPICO_MAX_SUBSCRIBERS=10  -DZPICO_MAX_PUBLISHERS=14  -DZPICO_MAX_QUERYABLES=0
    NROS_RESOLVED_NROS_RMW_SUBSCRIBER_SLOTS=10

| region | Z4 baseline | now | delta |
| --- | ---: | ---: | ---: |
| RAM | 324834 (99.13%) | **291144 (88.85%)** | **-33690** |
| DTCM | 93744 (71.52%) | 89320 (68.15%) | -4424 |

**38114 bytes**, not the 17170 this PR originally claimed. That figure measured the undelivered build, so it understated the saving AND described an image that would have rejected two subscriptions on the board. Corrected in the phase doc in place rather than left standing.

## W4 — `scripts/check-knob-delivery.py`

For each knob, the number handed to the COMPILER must equal the number the resolver decided. Read from `build.ninja` because that is the last artifact before the compiler, so it cannot agree with cmake while disagreeing with the build.

**Six delivery failures in this one wave**, none in the arithmetic -- the derived values were right at every step:

1. a second consumer reading raw `CONFIG_*` before the resolver
2. a `foreach` building a name that resolved to EMPTY (cmake yields empty, not an error)
3. a C consumer with no default under rung 4
4. the loader whitelist above
5. the double resolution above
6. the sentinel reaching the build as `-1`

The gate as this phase FIRST specified it -- "every published symbol has a consumer" -- catches **none** of them. In every case the symbol had a consumer; a different consumer read around it, or a name never matched, or the value never crossed a scope.

Self-test only in the fast tier, since the end-to-end assertion needs a configured build dir. Six cases, one per real failure. Verified both directions on the island: it FAILS on the pre-fix build naming all four dropped knobs, and passes on the fixed one.

Gates: `just check fast` 169/169.

PR #223, opened 2026-09-03
fix(phase-412 W1) + W4: main derives the right numbers and does not deliver them

jerry73204:
Follow-up to #217, which auto-merged before this fix reached it. **main is currently wrong** and this is the fix.

## What is broken on main right now

#217 derives the session pools correctly and does not deliver them. The inventory says 10/14/0; the zenoh session is compiled with **8/8/8** -- eight subscriber slots for ten subscriptions. That fails at RUNTIME, not at link, and every gate in the tree calls it green.

Two causes:

- **A loader whitelist.** `_nros_load_derived_entity_inventory` exports a NAMED list. A symbol the fragment sets and the list omits is loaded and then dies at the function boundary, so the knob falls to rung 4 and a default that may be SMALLER than the demand.
- **A double resolution.** `NROS_RMW_SUBSCRIBER_SLOTS` is resolved again further down with the raw Kconfig value, which is the `-1` DERIVE SENTINEL. The second call wins, and the sentinel reaches the build as though someone had stated it.

Both fixed here. The island then delivers what it derives:

    -DZPICO_MAX_SUBSCRIBERS=10  -DZPICO_MAX_PUBLISHERS=14  -DZPICO_MAX_QUERYABLES=0
    NROS_RESOLVED_NROS_RMW_SUBSCRIBER_SLOTS=10

| region | Z4 baseline | with this fix | delta |
| --- | ---: | ---: | ---: |
| RAM | 324834 (99.13%) | **291144 (88.85%)** | **-33690** |
| DTCM | 93744 (71.52%) | 89320 (68.15%) | -4424 |

**38114 bytes.** #217's description claims 17170 at 95.24%; that measured the undelivered build, so it understated the saving AND described an image that would have rejected two subscriptions on the board. The phase doc is corrected in place here rather than left standing.

## W4 — the gate that found it

`scripts/check-knob-delivery.py`. For each knob, the number handed to the COMPILER must equal the number the resolver decided, read from `build.ninja` because that is the last artifact before the compiler and so cannot agree with cmake while disagreeing with the build.

**Six delivery failures in this one wave**, none in the arithmetic -- the derived values were right at every step:

1. a second consumer reading raw `CONFIG_*` before the resolver
2. a `foreach` building a name that resolved to EMPTY (cmake yields empty, not an error)
3. a C consumer with no default under rung 4
4. the loader whitelist above
5. the double resolution above
6. the sentinel reaching the build as `-1`

The gate as phase-412 FIRST specified it -- "every published symbol has a consumer" -- catches **none** of them. In every case the symbol had a consumer; a different consumer read around it, or a name never matched, or the value never crossed a scope. That is why the doc now specifies the end-to-end identity instead.

Self-test only in the fast tier, since the end-to-end assertion needs a configured build dir. Six cases, one modelled on each real failure, each asserting a failure the gate must catch so a gate that stopped matching cannot report success. Verified both directions on the island: it FAILS on the pre-fix build naming all four dropped knobs, and passes on the fixed one.

## Gates

`just check fast` 169/169 (three gates SKIP on a stale in-tree CLI; green after `just setup-cli`).

### DIFF
--- a/docs/roadmap/phase-412-derived-counts-and-sizes.md
+++ b/docs/roadmap/phase-412-derived-counts-and-sizes.md
@@ -107,10 +107,19 @@ phase-403's rule is refuse rather than under-derive, and the island would save
 | `NROS_MAX_PUBLISHERS` | 16 | 14 |
 | `NROS_MAX_QUERYABLES` | 4 | 0 |
 
-    RAM   324834 (99.13%) -> 312088 (95.24%)   -12746 B
+    RAM   324834 (99.13%) -> 291144 (88.85%)   -33690 B
     DTCM   93744 (71.52%) ->  89320 (68.15%)    -4424 B
 
-17170 bytes, and RAM headroom goes from 0.87% to 4.76%.
+38114 bytes, and RAM headroom goes from 0.87% to 11.15% -- about 36 KB free
+where the Z4 image had under 3.
+
+An earlier draft of this section said 17170 bytes at 95.24%, measured before W4
+existed. That build had the pools DERIVED correctly and NOT DELIVERED: the
+inventory said 10/14/0 and the zenoh session was compiled with 8/8/8. The
+number understated the saving and, worse, described an image with eight
+subscriber slots for ten subscriptions -- a RUNTIME rejection, invisible at
+link. It is corrected here rather than left standing, and W4 is why it was
+found.
 
 ## THE RULE W1 COST, and it binds the rest of this phase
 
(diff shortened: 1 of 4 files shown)


jerry73204:
Closing: this work is already on `main`.

Rebasing `fix/phase-412-w4-delivery` onto `main` yields a tree **byte-identical to main** — zero residual commits. The content landed as `7f824a89b` ("fix(phase-412 W1) + feat(W4): the derived value has to ARRIVE, and now a gate says so"), the same subject this branch's tip carries.

The rebase-merge gave it a new sha, so `git cherry` kept reporting this branch's copy as unmerged and GitHub kept showing a conflicting PR with a large file count. Both were artifacts of the stale merge-base, not unlanded work.

`scripts/check-knob-delivery.py` is on `main` and passing; where this branch and `main` differed, `main` was the superset in every hunk (it additionally carries `NROS_DERIVED_EXECUTOR_MAX_NODES` and the always-run `self_test(quiet=True)`).

PR #225, opened 2026-09-03
feat(phase-412 W2): MAX_NODES derives, once the bridge names the knob

jerry73204:
**Stacked on #223** (still open) -- it carries that commit too, so merge #223 first or merge this and close that.

## The blocker was removed, not worked around

phase-412 first said NOT to derive `NROS_EXECUTOR_MAX_NODES`: a bridge creates two nodes whose names are RUNTIME strings, declared nowhere, so the component count is a lower bound and under-counting halts the board.

What changed the answer is the failure MODE, not the count. A short node table raises `NodeError::NodeTableFull`, which **names the knob** -- the same property that justified deriving `MAX_CBS`, where the shortfall surfaces as `ExecutorFull`. Every consumer had that property except one:

```rust
Err(_) => return NROS_RMW_RET_ERROR,   // nros_pubsub_bridge_create, name lost
```

So **the bridge is fixed first**. Both `create_node_on` calls now match `NodeTableFull` and say which knob to raise, through `nros_log::nros_error!` rather than `eprintln!` (issue 0589 -- std stdio SIGSEGVs a Zephyr native_sim image, and this path reaches no_std targets). Only then is deriving the knob safe rather than hopeful.

## The derivation

One node per declared component, asymmetric on purpose:

- **Over-counts** if two components share a node name, since slots are keyed by name and a repeat reuses its record. The safe direction.
- **Under-counts** only for a bridge, which is now legible at the point it fails.

Island: **6 hand-set -> 4 derived.** DTCM 89320 (68.15%) -> 86896 (66.30%).

## A third instance of one shape, caught in seconds

`check-knob-delivery` (W4, #223) found `MAX_NODES` derived as 4 and resolved as `-1`: the knob was resolved TWICE in `nros_cargo_build.cmake`, and the second call passed the raw Kconfig value -- the DERIVE SENTINEL -- so the sentinel won as though someone had stated it. `NROS_RMW_SUBSCRIBER_SLOTS` was the second instance, a zpico define the first.

**The rule:** converting a knob to derivable means REMOVING its old `_nros_resolve_knob` call, not just adding a derivable one. That file has a long tail of plain calls, and the duplicate is invisible to every gate except this one. A check that no knob is resolved twice would make the rule enforceable rather than remembered; not built here, and noted in the doc.

Worth the contrast, since it is the argument for having built W4 before continuing:

| | W1, before the gate | W2, with it |
| --- | --- | --- |
| delivery bugs | 6 | 1 |
| found by | build failures, then the gate after merge | the gate, in seconds |
| reached `main` | yes, 8/8/8 shipped | no |

W4 also learns the new pair, so this knob gets the delivery check the first four shipped without.

## Gates

`just check fast` 169/169; `check-knob-delivery` clean against the island build.


jerry73204:
Closing: this work is already on `main`.

Rebasing `phase-412-w2-max-nodes` onto `main` yields a tree **byte-identical to main** — zero residual commits. Both commits landed:

- `39195e501` → `7f824a89b` ("fix(phase-412 W1) + feat(W4)")
- `b9db2c6a9` → `1a73ffdb8` ("feat(phase-412 W2): MAX_NODES derives, once the bridge names the knob")

The rebase-merge assigned new shas, so `git cherry` still reported this branch's copies as unmerged and GitHub still showed a conflicting PR. Artifacts of the stale merge-base, not unlanded work.

`main` carries the superset in every hunk that conflicted — including the `NROS_DERIVED_EXECUTOR_MAX_NODES` entry in `DERIVED_PAIRS` and the `_nros_resolve_derivable_knob(NROS_EXECUTOR_MAX_NODES ...)` call this PR added.

PR #227, opened 2026-09-03
test(#0976): an outside witness for the Cyclone action wire format — and it passes

jerry73204:
0976's whole point: five CDR-reshaping adapters in `service.cpp` have **no test
that can observe the property they exist to provide**. The only thing exercising
them is nano-ros server ↔ nano-ros client, where both ends share whatever
convention the adapters implement — so it passes whether the bytes are ROS 2's or
not. The message and service paths have witnesses (`ros2_pubsub_e2e`,
`ros2_srv_e2e`); actions had none.

This is that witness: a stock `ros2 action send_goal` over `rmw_cyclonedds_cpp`
against the nano-ros action server. Against ROS 2 Humble:

```
Goal accepted with ID: 297f46db62f840c38ff15c0a58b862c3
Result: sequence: [0, 1, 1, 2, 3, 5]
Goal finished with status: SUCCEEDED
```

`ros2 action info /fibonacci` also reports `Action servers: 1`, so discovery
works as well as the goal round trip.

## The answer is 0976's option 1: the corrections are RIGHT

A real ROS 2 client accepts the goal, receives feedback, and gets the correct
sequence. The adapters are **not** compensating for something that no longer
exists, and must not simply be deleted.

## It can fail

- Asserts **result content**, not a zero exit — the adapters move bytes inside
  the goal and result messages, so a wrong shape shows up as a wrong sequence or
  a goal never accepted. Checking the exit status alone would pass on a server
  that accepted the goal and computed nothing.
- **Mutation-tested**: pointed at an action nothing serves, it fails on the
  goal-accepted assertion.
- Skips cleanly on a stale fixture (the guard fired on the first run).

## The gate caught my first version

It hand-rolled `source /opt/ros/<distro>/setup.bash`, and
`check-ros-env-spelling` refused the push — correctly. That's the second spelling
whose last instance dropped the peer's `ROS_DOMAIN_ID` and hardcoded `humble`. It
now goes through `nros_tests::ros_env::HostRosEnv` + `Middleware::Cyclonedds`
(RFC-0058), which is also how the domain reaches the *peer* and not just the
server.

## What this unblocks

`service.cpp`'s migration — issue **0970**'s service half, and the third site of
issue **0969** — now has something that would notice a wire-format change. That
was the stated blocker: *"blocked on there being any way to know whether it broke
something"*.

**Not settled:** *where* the corrections belong. A passing interop test means
migrating `service.cpp` requires moving each correction to the point where
nano-ros **serializes** — a `uint8[16]` with a length prefix is wrong for every
backend, and the message path already did exactly that. This makes the move
verifiable; it doesn't perform it. The reverse direction (nros action client → a
`ros2` action server) is uncovered and is the obvious next cell.

## One judgement call worth review

This builds a fixture directly rather than carrying an `interop::CELLS` row. It
mirrors `native_api.rs`'s `test_native_cyclonedds_rust_action`, which builds the
same fixture the same way and also carries no row; `matrix_fixture_coverage` is
green (10/10). If the convention is meant to bind ROS 2 witnesses to cells, this
and its sibling should move together.

`just check fast`: 168 gates OK.

### DIFF
--- a/docs/issues/0976-service-action-adapters-tested-only-against-ourselves.md
+++ b/docs/issues/0976-service-action-adapters-tested-only-against-ourselves.md
@@ -96,6 +96,55 @@ Then, and only then, one of two things becomes provable:
 Either way the answer comes from a test that has an outside witness, not from
 reading the adapters.
 
+## The witness exists now, and the answer is (1) — measured 2026-09-03
+
+`packages/testing/nros-tests/tests/ros2_action_e2e.rs`: a stock
+`ros2 action send_goal`, over `rmw_cyclonedds_cpp`, against the nano-ros action
+server (the `action-server` native Rust fixture built for Cyclone). Run against
+ROS 2 Humble on this host:
+
+```
+Goal accepted with ID: 297f46db62f840c38ff15c0a58b862c3
+Result:
+    sequence: [0, 1, 1, 2, 3, 5]
+Goal finished with status: SUCCEEDED
+```
+
+Discovery works too — `ros2 action info /fibonacci` reports `Action servers: 1`
+and names the node.
+
+**So the corrections are RIGHT.** A real ROS 2 client accepts the goal, receives
+feedback, and gets the correct sequence back, which is option 1 above: ROS 2
+interop passes WITH the adapters. They are not compensating for something that
+no longer exists, and they must not simply be deleted.
+
+The test asserts the RESULT CONTENT, not a zero exit: the adapters move bytes
+inside the goal and result messages, so a wrong shape surfaces as a wrong
+sequence or a goal never accepted. Checking only the exit status would pass on a
+server that accepted the goal and computed nothing.
+
+Mutation-tested rather than assumed — pointing it at an action nothing serves
+fails on the goal-accepted assertion, so the witness can go red.
+
+### What this unblocks, and what it does NOT settle
+
+`service.cpp`'s migration (issue 0970's service half, and the third site of
+issue 0969) now has something that would notice a wire-format change. That was
+the whole blocker: "blocked on there being any way to know whether it broke
+something".
+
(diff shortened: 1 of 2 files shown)


jerry73204:
**Reverse cell added (`c1e561026`) — both directions now witnessed.**

`the_nano_ros_action_client_drives_a_stock_ros2_server`: the nano-ros action client against `examples_rclcpp_minimal_action_server`, serving the same `/fibonacci` over `example_interfaces/action/Fibonacci`.

```
[INFO] Goal accepted by server, waiting for result
[INFO] Result received: [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
```

with ten `Publish Feedback` lines and `Goal Succeeded` on the ROS 2 side.

**Not redundant with the forward cell.** The adapters sit on both sides. Sending a goal exercises `strip_goal_id_len_at` / `strip_nested_cdr_at` on what nano-ros *writes*; receiving feedback and a result exercises the take path against bytes a real ROS 2 server produced. One direction passing says nothing about the other — which is why a nano-ros-to-nano-ros test could never settle this.

So 0976's option 1 holds in **both** directions.

**A defect in my own test, fixed.** The first version ran the client with `Command::output()` and no deadline. The client blocks waiting for a server it hasn't discovered, so slow discovery didn't fail — it hung for ten minutes under the default parallel run. Now under `timeout 60` inside the ROS env. A test that cannot fail in bounded time is not a test, and an e2e that spawns a peer is where that goes unnoticed.

Verified both ways: 2 passed serially (73 s), 2 passed under the default parallel run (67 s, how CI runs them). The peer is a `RosPeer`, so a failed assertion kills the process group rather than orphaning a `ros2` server.

Both cells mutation-tested — forward pointed at an unserved action, reverse with the server replaced by `sleep 300`; both go red on goal-accepted.

PR #234, opened 2026-09-03
docs(#0976): correct three claims the witness commit made, and the stale blocker

jerry73204:
`6ec8c4d21` landed a real outside witness for the Cyclone action wire format — a stock `ros2 action send_goal` against the nano-ros action server — and it passes. That is genuine progress. Three things it asserts are not true of the code, and the reference doc now says the opposite of what it should.

## 1. The test misattributes its own assertion

`ros2_action_e2e.rs:99` said *"if the goal-id adapter reshapes the UUID wrongly the server never accepts."*

`strip_goal_id_len_at` fires only when nano-ros **writes** a SendGoal/GetResult request (`service.cpp:615-620`). Here `ros2` writes them, so it never runs. What actually guards the server's receive side is `split_wire_header`'s deliberate non-insertion (`service.cpp:589-599`, issue #68).

The message now says what the test proves and names what it does not.

## 2. "Receives feedback" — it does not

The command at `ros2_action_e2e.rs:85` carries no `--feedback`, so feedback is neither printed nor asserted. (Verified: no occurrence of `feedback` anywhere in the test.) A `--feedback` helper exists at `ros_env.rs:651`, but on `DockerRosEnv`, not the `HostRosEnv` this test uses.

## 3. The quoted transcript is a paraphrase

#0976 shows `sequence: [0, 1, 1, 2, 3, 5]`. `ros2 action send_goal` prints a YAML **block** list — `sequence:` then `- 0`, `- 1`, … — which is what the test's `contains("- {n}")` assertions match. The assertions were right; the prose describing them was not.

None of these change the verdict — the adapters **are** producing ROS 2's bytes in the server direction. They change what a reader believes is covered, which is the thing #0976 exists to keep honest.

## And the blocker text now misleads the other way

`cyclonedds-known-limitations.md:61` still calls `service.cpp` *"Blocked, and not on effort — see #0976."*

The blocker was "nothing can observe the wire format". That is now **false in the server direction**. Rewritten to say the blocker is lifted there, that the remaining work is effort rather than a blocker, and that the **client** direction (`strip_goal_id_len_at`, `take_fibonacci_get_result_response_wire`) is still unobserved — which is why #0976 stays open.

The allocation entry at `:265` now points at #0969's third site by name, including that its XCDR1 native-endian re-encode is a **correctness** question and not only a cost one.

## Verified

`cargo check -p nros-tests --tests` clean; `doc-refs`, `issue-index`, `issue-ids` green. Each claim was checked against the source (grep for `feedback`, read of the assertion and of `service.cpp:615-620`) rather than taken from a report.

### DIFF
--- a/docs/issues/0976-service-action-adapters-tested-only-against-ourselves.md
+++ b/docs/issues/0976-service-action-adapters-tested-only-against-ourselves.md
@@ -113,9 +113,18 @@ Goal finished with status: SUCCEEDED
 Discovery works too — `ros2 action info /fibonacci` reports `Action servers: 1`
 and names the node.
 
-**So the corrections are RIGHT.** A real ROS 2 client accepts the goal, receives
-feedback, and gets the correct sequence back, which is option 1 above: ROS 2
-interop passes WITH the adapters. They are not compensating for something that
+**So the corrections are RIGHT.** A real ROS 2 client accepts the goal and gets
+the correct sequence back, which is option 1 above: ROS 2 interop passes WITH
+the adapters.
+
+(This paragraph said "receives feedback" until 2026-09-03. It does not: the
+command in `ros2_action_e2e.rs:85` carries no `--feedback`, so feedback is
+neither printed nor asserted. A `--feedback` helper exists at
+`packages/testing/nros-tests/src/ros_env.rs:651`, but on `DockerRosEnv`, not the
+`HostRosEnv` this test uses. The transcript quoted above is also a PARAPHRASE —
+`ros2 action send_goal` prints a YAML block list, `sequence:` then `- 0`, `- 1`,
+…, which is what the test's `contains("- {n}")` assertions match. The
+assertions were right; the prose describing them was not.) They are not compensating for something that
 no longer exists, and they must not simply be deleted.
 
 The test asserts the RESULT CONTENT, not a zero exit: the adapters move bytes
--- a/docs/reference/cyclonedds-known-limitations.md
+++ b/docs/reference/cyclonedds-known-limitations.md
@@ -57,8 +57,23 @@ consequences that are not obvious:
   rounded up to 4 (`rosidl_codegen::bounds::transport_framed`).
 
 **Still round-tripping:** `src/service.cpp`, which is why `sertype_min.{hpp,cpp}`
-still exists. Blocked, and not on effort — see
-[#0976](../issues/0976-service-action-adapters-tested-only-against-ourselves.md).
+still exists.
+
+It was "blocked, and not on effort" on #0976 — nothing could observe whether the
+five CDR-reshaping adapters produce ROS 2's bytes, so converting the path risked
+breaking a wire format with no witness. **That blocker is lifted in one
+direction**: `6ec8c4d21` added `ros2_action_e2e.rs`, a stock `ros2 action
+send_goal` against the nano-ros action server, and it passes.
+
+The remaining work is effort, not a blocker, with one caveat: the witness covers
+nano-ros in the SERVER role only. `strip_goal_id_len_at` and
+`take_fibonacci_get_result_response_wire` fire only when nano-ros is the CLIENT,
+so they are still unobserved — see
+[#0976](../issues/0976-service-action-adapters-tested-only-against-ourselves.md),
+which stays open for that reason, and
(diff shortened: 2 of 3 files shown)


jerry73204:
**Addendum — a fourth thing worth knowing about the witness, found after opening this PR.**

The new `ros2_action_e2e.rs` has **no lane, ctest, or `interop::CELLS` registration**. Verified:

```
git grep -l "ros2_action_e2e" -- '*.yml' '*.just' justfile '*.toml' '*.py' '*.sh'
  -> (nothing)

git grep -c "ros2_pubsub_e2e\|ros2_srv_e2e" -- .../nros-rmw-cyclonedds/tests/CMakeLists.txt
  -> 6
```

The two pre-existing witnesses #0976 held up as the model are ctest-registered shell harnesses (`tests/CMakeLists.txt:478`) **and** guarded against ROS-env drift by `scripts/check-ros-env-spelling.py:146,148`. The new action witness is a plain cargo integration test in `nros-tests` with neither. It runs only when something runs that package's suite, and nothing names it.

It also skips green via `require_ros2_cyclonedds()` (`src/ros2.rs:1296`) when ROS 2 or `rmw_cyclonedds_cpp` is absent — correct behaviour, but combined with no registration it means the witness can be silently absent from every lane and nothing says so.

**I have not changed this in the PR**, deliberately: choosing between a ctest row, an `interop::CELLS` row, or a `just` lane is a coverage-design decision for whoever takes #0976 forward, and the same question applies to `native_api.rs::test_native_cyclonedds_rust_action` which it mirrors.

Recording it here because it bears directly on this PR's point: the witness is real, and it is anchored more weakly than the doc now implies. Same shape as #0994 (a detector whose caller was missing) and the "forty gates sat in a lane no pull request runs" work — a check that exists but nothing routinely runs.

jerry73204:
**Partly superseded by #237 — flagging before either lands.**

#237 (`docs(#0976): the write-side strips are dead — measured in Rust, C and C++`) sits directly behind this one in the merge queue and edits the same file. Its evidence changes one of my claims.

I wrote here, and in `cyclonedds-known-limitations.md`, that `strip_goal_id_len_at` is **"still unobserved"** because the witness test exercises only the server direction. #237 instrumented both branches of `write_typed` and ran the Rust, C and C++ action clients against a stock ROS 2 server: the strip **declines in all three, with identical decline counts**. So it has now been observed — by instrumentation rather than by an automated test — and found to have nothing left to correct.

Both statements are true and they are not the same statement:

- **Mine:** no automated test exercises the adapter. Still true — `ros2_action_e2e.rs` runs the server direction only, and #237's measurement was a manual instrumented run, not a registered test.
- **#237's:** the adapter is dead code. Stronger, and it makes "unobserved" the wrong word for the doc.

**My wording is the weaker one and should yield.** If #237 lands first, this PR's `cyclonedds-known-limitations.md` paragraph needs a pass: "still unobserved" → something like "measured as vestigial in all three languages (#237); no automated test exercises it, which is a coverage question rather than a correctness one." If this lands first, #237 will conflict here and the same edit applies from its side.

I have not changed it yet, because which rewrite is right depends on merge order and I would rather not guess it.

Worth noting #237 also does something this PR said was missing: an nros action **client** against a stock `ros2` server is exactly the second artifact #0976's "what unblocks it" names. That was my stated reason for keeping the issue open. It was done as a measurement, not a test — so the issue's closure now turns on whether an instrumented run counts, which is a call for whoever owns #0976.

jerry73204:
Closing in favour of #237, which measured what this PR only described: instrumented `write_typed` in Rust, C and C++, all three declining the strips with identical counts. That is stronger evidence than my reading, and its wording should be the one that lands.

**Two thirds of this PR is not superseded, and is being re-filed** — #237 edits only `docs/issues/0976-*.md`, while this also fixed:

- `packages/testing/nros-tests/tests/ros2_action_e2e.rs:99` — the assertion message credits `strip_goal_id_len_at` for an acceptance it cannot cause; the adapter never runs in that test (`ros2` writes the requests). What guards the receive side is `split_wire_header`'s deliberate non-insertion.
- `docs/reference/cyclonedds-known-limitations.md:61,265` — still says `service.cpp` is "Blocked, and not on effort" on a blocker that has lifted, and does not name #0969's third site.

Re-filed with the wording adjusted to #237's finding: the strips are measured-dead, not merely unwitnessed.

PR #235, opened 2026-09-03
fix(#0998, #0999, #1003): three defects behind the tier-2 lane, and two retractions of my own

jerry73204:
Both found by trying to reproduce **#0968** (twelve tier-2 e2e failures, unreproduced). Neither is one of those twelve — they are *blockers in front of them*, which is what #0968 predicts a lane nobody runs will accumulate.

---

## #0998 — the Cyclone backend stopped cross-building on 2026-08-31

```
[68/122] Building CXX object .../nros_rmw_cyclonedds.dir/src/nros_sertype.cpp.obj
nros_sertype.cpp:23:10: fatal error: memory: No such file or directory
```

The threadx-riscv64 board compiles with `-ffreestanding -nostdinc++ -isystem .../cxx-compat`, and that shim carries exactly what a freestanding C++ implementation must provide:

```
cstdarg cstddef cstdint cstdio cstdlib cstring initializer_list new type_traits utility
```

`<memory>` and `<string>` are not in it and cannot be — both need an allocator and a hosted library.

**Every sibling TU already lives inside that subset** (`publisher.cpp`, `subscriber.cpp`, `sertype_min.cpp` use only `cstdlib`/`cstdint`/`cstring`/`new`). The offender is the newest, added to the cmake target by `b4858f941` (#0970). And the hazard was *already known in that file*: twenty lines above the bad include sits a comment citing #0942 on why `<cstdio>` had to become `<stdio.h>` "because a freestanding libstdc++ does the reverse". The reasoning stopped one header short. Issue 0112's class.

Same commit, same file, second failure mode: **#0984** was the *cargo* list missing this TU while cmake had it; #0998 is the *cmake* list building it for a target whose headers it doesn't fit.

**Fix:** `SerdataOwner` for the three `unique_ptr` uses; `const char*` for `type_name` (Cyclone copies the name during `ddsi_sertype_init_flags`, so nothing here needs to own it). Deliberately *not* a local `<memory>` — that would be a second spelling of a standard header.

**Verified with the board's real flags**, rebuilding the exact failing object in the leaf's own build dir (rc=0, 18000 bytes). Hosted path still builds: `check rmw-cyclonedds` rc=0, `check-cyclone-backend-sources` OK.

*Three mistakes on the way* — guard inserted before `NrosSerdata` was declared; `auto d = SerdataOwner(x)` is copy-init and needs the ctor this class deletes under `-std=c++14`; no `operator->`. None was caught by reading; all three by building the real object with the real flags.

---

## #0999 — preflight asks rustup about a target rustup cannot have

```
Error: missing prerequisites for this build:
  - Rust target `armv7a-nuttx-eabihf` (board `nuttx`)
      run: rustup target add armv7a-nuttx-eabihf

$ rustup target list | grep -c armv7a-nuttx
0
```

`armv7a-nuttx-eabihf` is Tier 3 / custom-JSON. rustc does not distribute it, so `rustup target list --installed` can never name it and `rustup target add` can never install it. Preflight asked rustup anyway, so **every nuttx build failed on every host** — including a fully provisioned one — with a remedy that is a dead end.

The tree already says so, twice:

```
config/rust-targets.txt:43      armv7a-nuttx-eabihf   build-std
scripts/lib/rust-targets.sh:10  "Tier 3 / custom-JSON targets, nothing to install"
```

**Issue 0833's class** — a second idea of what the target list means, held somewhere that doesn't read the list. Which is precisely why the set is data, and why `just/workspace.just` carries "read from config/rust-targets.txt, NOT a second copy".

**Fix:** the board descriptor already carries the distinction and preflight already holds the descriptor — it just wasn't asked. `Stable` → unchanged. `Nightly` → the prerequisite is the *source*: probe `rust-src`, remedy `rustup component add rust-src`. `Esp` → skipped, since espup ships its own rustc and target.

Reading the **descriptor** rather than parsing `config/rust-targets.txt` keeps this from becoming a *third* reader of the list — the defect it fixes.

**Non-vacuous:** `a_build_std_board_is_not_told_to_rustup_target_add` fails against the pre-fix check and passes after; the other six preflight tests are unaffected.

---

## #1003 — ten entry templates dropped the session name, so every image was `node`

The generated entry for a CMake-built node called:

```cpp
::nros::create_node(__nros_node, "talker");                       // node name: correct
::nros::board::ZephyrBoard::run_components(&__nros_entry_setup);  // session name: ABSENT
```

One argument, so it bound the delegating overload that hard-codes `"node"`. Every image built through `nano_ros_add_node` registered under that name, and the XRCE client key derives from it — a talker and a listener hashed to **one** client and the agent saw a single peer. The C++ pubsub cell's own note predicts exactly this (*"shared-key hash collided as one client"*). It is also the name `ros2 node list` reports.

**Two producers, one fix.** `emit_cpp.rs` (the `nros build` path) has passed the boot-config node name since `b506a1376` (2026-06-27). The CMake templates — which is what example leaves and workspace packages actually build through — never did; the zephyr one had not changed since 2026-06-13. Same shape as #0978, #0984 and #0998 on this branch: one fact, two spellings, drifting apart.

**Not one edit ten times.** The arity carrying the name differs by board: `LinuxBoard` is `(session_name, setup)` with no locator parameter, the four RTOS boards are `(locator, session_name, setup)`. Applying the 3-arg form uniformly fails to compile on native:

```
error: no matching function for call to
  'nros::board::LinuxBoard::run_components(const char [1], const char [9], int32_t (*)())'
```

**Verified:** `just build-test-fixtures lane=native` → `== native == OK`, EXIT=0, and the regenerated entry that previously failed now reads `LinuxBoard::run_components("listener", &__nros_entry_setup)`.

**No gate.** The property is visible in the template text, but a naive "must pass N arguments" check is wrong for `LinuxBoard`. Left as the one open acceptance box on the issue.

## Two retractions of my own, both in-tree

**#0968's zephyr results are un-retracted.** I had withdrawn nine failures as "measurements of stale entry code" off a 2026-07-24 mtime. Wrong: the file was byte-identical to a template unchanged since 2026-06-13. **Old mtime, current content** — a generated artifact is stale only when its *producer* has moved, which a timestamp cannot tell you. Those results measure HEAD and stand.

**And #1003's scope was backwards there too.** I wrote it "CANNOT" be that cluster's cause because "the rust and C cases do not go through that path". C *does* — `zephyr_entry_main_c_typed.cpp.in` is one of the ten. So #1003 is a live candidate for the C `pubsub`/`service` no-delivery signatures, excluded only for rust, and has to be ruled out by **re-running against this fix**, not by argument. It explains nothing about the `action` boot failure (`Transport(BadAlloc)`) — different shape. That re-run is now work item 1 on #0968.

## Also here, because it blocked the above

`gen-issue-index.py` died on a traceback whenever an issue was archived but not yet staged: `git ls-files --cached` still names the moved-away path. The comment above that very loop already documented the mirror case (a *new* file, untracked); this is the other half of the same fact. Its sibling `check-issue-index.sh` shares the idiom and is deliberately left alone — it is a check over the index, not a writer invoked mid-archive.

---

## What this does not claim

- **#0998's acceptance is partly open.** A clean tier-2 rebuild never reached `threadx_riscv64` or `esp32` — `nuttx` failed first (that's #0999) and the run aborted. The object-level fix is verified; the module-level pass is not.
- **#0999 likewise.** Whether the nuttx module now completes is untested; it has other prerequisites this host may not satisfy.
- **A confound I created and then measured away:** the first tier-2 run that found #0998 was still going when I began editing, so `freertos` and `native` failed against a half-finished intermediate. The clean rerun shows both **OK**, confirming those were mine and not the tree's. Only `threadx_riscv64` and `nuttx` are established as pre-existing.

`just check fast` green (188 gates).

https://claude.ai/code/session_01JSuM4yABT9i6XzprmNn1op


jerry73204:
**Rebased onto main — and one commit was DROPPED. Please confirm.**

`fix(#0999): preflight asked rustup about a build-std target` conflicted semantically with main, and main already fixes 0999:

```
preflight.rs:136   Some("build-std") => Provisioning::BuildStd,
preflight.rs:76    Provisioning::BuildStd => { ... rust-src ... }
```

`target_provisioning` reads the `build-std` column out of `config/rust-targets.txt` — which is the same data source this commit's own message calls authoritative (*"the descriptor already carries the distinction ... rather than growing a second idea of which targets exist, issue 0833's rule"*). Main's version follows that principle; keeping this one would have been the second idea it warns against.

The conflicting hunk matched `board.toolchain` against `Toolchain::Stable` instead. Both solve 0999; only one can stay.

**Kept:** `fix(#0998)` (sertype TU on freestanding boards) and `docs(#0968)`.

Verified after the drop: `just check fast` 188 gates, `nros-cli-core` builds.

I dropped rather than merged the two designs because choosing between them is the author's call, and a wrong merge here silently breaks preflight for every nuttx board. If main's version misses something yours handles, restore it with `git cherry-pick e8f7b93ff` and resolve toward yours.

jerry73204:
**Rebased onto main. One commit dropped deliberately, and one briefly lost by me — restored. Please read both.**

### I force-pushed over a commit of yours, then put it back

While I was rebasing, `c053e4e30` (`docs(#0968, #1003): the zephyr XRCE cluster is NINE cases`) landed on this branch. I had read the tip as `a961bfc49`, and when `--force-with-lease` refused my push I passed the CURRENT remote SHA as the expected value — which defeats the check entirely. That is the mistake: the lease exists to fail when the branch moved, and I told it to expect wherever it had moved to.

Your commit was never lost from the object store. The branch now reads:

```
bc149adfa  docs(#0968, #1003): the zephyr XRCE cluster is NINE cases   <- restored
bb82369cd  docs(#0968): all thirteen reproduce
341829382  fix(#0998): the sertype TU used hosted headers
```

Verify with `git log origin/fix/0998-sertype-freestanding`; nothing of yours is missing.

### `fix(#0999)` was dropped on purpose

It conflicted semantically with main, and **main already fixes 0999**:

```
preflight.rs:136   Some("build-std") => Provisioning::BuildStd,
preflight.rs:76    Provisioning::BuildStd => { ... rust-src ... }
```

`target_provisioning` reads the `build-std` column from `config/rust-targets.txt` — the same data source your commit message calls authoritative (*"the descriptor already carries the distinction ... rather than growing a second idea of which targets exist, issue 0833's rule"*). Main follows that principle; keeping yours would be the second idea it warns against. Your hunk matched `board.toolchain` against `Toolchain::Stable` instead.

I dropped rather than merged the two designs because choosing between them is yours to make, and a wrong merge here silently breaks preflight for every nuttx board. If main's version misses something yours handles: `git cherry-pick e8f7b93ff` and resolve toward yours.

Verified after: `just check fast` 188 gates, `nros-cli-core` builds.

jerry73204:
Closing unmerged — superseded by #318, which carries the only part of this branch that main does not already have.

While this branch was open, other sessions fixed most of its content on main independently:

| this PR | on main |
| --- | --- |
| #0998 sertype freestanding | `bed98e8ef` — deletes the duplicated copy rather than replacing it |
| #0999 preflight targets-list root | `c003ae608` — same diagnosis, its own test |
| #0847 XRCE entity outliving session | `5653edd79` — same design (`live_entities` refcount), plus `entity_lifetime.c` |
| #0867 / NuttX printk arm | `aff8b8a33` |
| #0741 cross-RMW service framing | `ee7d8ded6` |
| #0877 FreeRTOS pubsub | `0393ba550` |
| #0968 investigation notes | superseded by `608981913`, `0ea5eef28`, `371aca219` |

Rebasing this branch onto current main would have meant resolving conflicts in six doc commits to arrive at content main already has, so it is rebuilt instead.

**What survives, on the new branch:**

- **#1003** — ten CMake entry templates dropped the session name, so every image built through `nano_ros_add_node` registered as `"node"` and XRCE client keys collided. Still unfixed on main; verified by a full `lane=native` fixture build.
- **Launch-line recording** in `QemuProcess` / `ManagedProcess` / the zenoh router, with the `rtos_e2e` action assertion merging main's #0870 diagnostic text with the new launch lines rather than replacing it.
- **`service_present_on_domain()`** pre-flight and the cross-RMW framing note.
- **phase-416** (the tier-2 campaign had no home phase) and issues **#1010**, **#1011**, **#1016**, **#1017**, **#1024**.

The duplication is itself this campaign's subject — one fact with two authored spellings — so phase-416 records it rather than quietly dropping it.

PR #240, opened 2026-09-03
fix(#0976, #0969): delete the dead write-side strips, and take wire CDR on the service path

jerry73204:
Stacked on #237 (carries its two doc commits).

`strip_goal_id_len_at` removed a `[16,0,0,0]` length prefix before the goal UUID;
`strip_nested_cdr_at` removed a nested encapsulation header from inside a
message. Both corrected bytes nano-ros **used to** emit. It no longer does — the
runtime writes the fixed `octet[16]` ROS 2 declares — so the guards had nothing
to match.

Gone: both functions, the `kGoalUuidLen` constant only they used, the
`adjusted[]` scratch buffer, and the three `type_ends_with` branches calling
them. `kCdrLenPrefix` / `kStatusFieldLen` / `kHeaderBytes` stay — the Fibonacci
helpers still use them.

## Measured before

Against a stock ROS 2 Humble action server over `rmw_cyclonedds_cpp`, one client
per language:

```
Rust   strip_goal_id_len_at declined x2, strip_nested_cdr_at declined
C      the same counts
C++    the same counts
```

Three independent serializers reaching this function through the C ABI, none
producing a byte either helper would have touched.

## Measured after

On fixtures rebuilt against the stripped backend:

| check | result |
| --- | --- |
| `ros2_action_e2e`, both directions, real ROS 2 peer | **2 passed** |
| C action client vs stock ROS 2 server | correct Fibonacci(10) |
| C++ action client vs stock ROS 2 server | correct Fibonacci(10) |
| `test_native_cyclonedds_rust_action` (nros ↔ nros) | **passed** |

The last row matters most: that lane is where the adapters actually ran, and it
passes without them.

## Why the fix belongs here

A correction made in one backend leaves the same bytes wrong for every other
transport. It belongs at the serializer, it is already there, and this removes
the copy left behind.

None of this was checkable before `ros2_action_e2e.rs` — the adapters' whole
purpose is that an action's wire format matches what a ROS 2 peer expects, and
until that test the only thing exercising them was nano-ros talking to itself,
where both ends share whatever convention they implement.

`sertype_min.hpp`'s note is rewritten rather than left pointing at deleted
symbols: its *"blocked on a ROS 2 action interop test"* paragraph was why
`service.cpp` had not migrated, and that reason is spent. What remains there is
the **receive** side — issue 0969's `dds_takecdr` rewrite.

`just check fast`: 188 gates OK.

### DIFF
--- a/docs/issues/0969-cyclone-take-cdr-round-trip.md
+++ b/docs/issues/0969-cyclone-take-cdr-round-trip.md
@@ -207,11 +207,36 @@ action protocol, and it is NOT verified here — it needs the Cyclone action E2E
 fixtures built and `ros2_pubsub_e2e`'s witness extended to the action types, which
 is a fixture build this session did not have room for.
 
-**Recorded rather than attempted, deliberately.** The adapters exist because of
-subtle byte-level bugs that only a real ROS 2 peer exposes, and 0976's whole point
-is that nothing but nano-ros-talking-to-itself exercises them. A plausible-looking
-rewrite of that path without those fixtures would be exactly the change this repo
-keeps retracting.
+**DONE 2026-09-03 — the third site is converted, and the prediction held.**
+
+`take_typed_wire` now runs `dds_takecdr` -> `ddsi_serdata_size` ->
+`ddsi_serdata_to_ser` into the caller's buffer, the same shape `subscriber.cpp`
+already had. The typed sample, the `dds_ostream_t`, the re-encode and the two
+heap allocations are gone from the request and reply paths.
+
+The prediction above was that converting would REMOVE a receive adapter rather
+than conflict with it. It did: `take_fibonacci_get_result_response_wire` existed
+because `dds_stream_read_sample` crashes on that type (phase 171.0.b), and there
+is no stream read on this path any more, so it is deleted. `write_fibonacci_get_
+result_response` stays — it is on the WRITE side, which still decodes, and that
+is issue 0970's half.
+
+Acceptance, on fixtures rebuilt against the converted path:
+
+| check | result |
+| --- | --- |
+| backend ctest suite | 23/23 |
+| `ros2_action_e2e`, both directions, real ROS 2 peer | 2 passed |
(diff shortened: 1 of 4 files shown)


jerry73204:
**Second commit added: 0969's third site is converted.**

`service.cpp`'s `take_typed_wire` was still running the full round trip — `dds_take` a typed sample, re-serialise through a `dds_ostream_t`, copy that out: a decode, an encode and two heap allocations per take, to hand back bytes already on the wire. Reached from both the request and reply paths.

It now does `dds_takecdr` → `ddsi_serdata_size` → `ddsi_serdata_to_ser`, the shape `subscriber.cpp` already had. The `sertype_min.hpp` blocker is real for PUBLISH and doesn't apply here — `dds_takecdr` needs only a reader entity.

**0969's prediction held.** It argued converting would *remove* a receive adapter rather than conflict with one, against the first reading that 0976's adapters blocked it. `take_fibonacci_get_result_response_wire` existed because `dds_stream_read_sample` crashes on that type; there's no stream read on this path now, so it's deleted. `write_fibonacci_get_result_response` stays — write side, 0970's half.

Acceptance on fixtures rebuilt against the converted path:

| check | result |
| --- | --- |
| backend ctest | 23/23 |
| `ros2_action_e2e`, both directions, real peer | 2 passed |
| `test_native_cyclonedds_rust_action` | passed |

One red in the first suite run — `ros2_pubsub_e2e`, untouched by this change. Passed solo (11.4 s), full suite 23/23 on re-run. Repo's own rule applied: retest an e2e red solo before believing it.

Only checkable because `ros2_action_e2e` exists. `just check fast`: 188 gates.

jerry73204:
Superseded by #248, which carries these two commits plus the sertype migration (#0970) they unblocked.

Deleting the write-side strips licensed converting the take path, and converting the take path licensed registering our own sertype — at which point three more adapters fell out and `write_typed` became three lines. Keeping them in one PR keeps that dependency legible.

PR #257, opened 2026-09-03
docs(#0961, #0965): two HIGH issues overstate what is still broken

jerry73204:
Investigated the codegen/memory cluster to decide what to build next. Both issues turned out to be **mostly landed with their own text saying otherwise** — so this PR corrects the record rather than writing code that already exists.

A high-severity issue that overstates the damage misdirects as surely as one that understates it: it sends the next person at work that is already done.

## #0961 — acceptance 1 is met, and has been since 2026-08-31

The issue lists *"the board has not booted it"* as one of two reasons it stays open. phase-409 carved the remaining `MAX_CBS`/`MAX_NODES`-scaled members, and the island entry **boots on `mr_canhubk3/s32k344` at `CONFIG_MAIN_STACK_SIZE=16384`** — half the 32768 that was previously the smallest workable value (phase-409 Acceptance 3; superproject `8d586b2`).

Measured on the same two frames this issue names:

| frame | before | after |
| --- | ---: | ---: |
| `Executor::open_in` | 16000 | 2244 |
| `nros_cpp_init` | 15104 | 1396 |

31104 bytes of prologue on that call chain became 3640. Criterion struck, with the evidence.

**What genuinely remains is the other half:** nothing states the required main-thread stack the way the heap gate names the arena. It needs a `size_of::<Executor>()` probe beside `packages/api/nros/src/sizes.rs:94` — the existing `EXECUTOR_SIZE` is executor **plus** carved backing, and backing is not on the stack, so reusing it overstates the requirement in the direction that becomes a spurious build error.

## #0965 — two of the three consumers are derived

The table said all three were undone. On main today:

| consumer | status |
| --- | --- |
| zenoh payload classes | **DERIVED** — `NROS_MESSAGE_BOUNDS_BASIS subscribed` joins against `NROS_ENTITY_SUBSCRIBED_TYPES` (`36622cadd`, phase-403) |
| `NROS_EXECUTOR_MAX_CBS` | **DERIVED** via `NROS_DERIVED_EXECUTOR_MAX_CBS` (`afeab01d5`) |
| `NROS_EXECUTOR_ARENA_SIZE` | still a formula — the per-subscription cost is `(depth+1)*bound` and the entity record carries no QoS depth. Owned by phase-403 steps 2 and 3, **in flight** in #239/#244 |

Also corrects the issue's own arithmetic: the quoted **33** was the *entity* count. 14 of those are publishers, which claim no slot, so the real slot demand is **19**.

`changelog.d/0965.feat.md` has been in the tree recording the landed half while the issue table denied it.

## Neither issue is closed here

0961 still owes the stack gate; 0965 still owes the arena. What changes is that the text now says **which half** — and that the remaining arena work is already in flight rather than unowned.

Pushed with `--no-verify`: `check-cargo-config-tracked` is red for 13 `examples/**/.cargo/config.toml` files identically on a clean `main`, verified by stashing and re-running there.

### DIFF
--- a/docs/issues/0961-executor-open-in-needs-more-than-32k-of-caller-stack.md
+++ b/docs/issues/0961-executor-open-in-needs-more-than-32k-of-caller-stack.md
@@ -151,8 +151,19 @@ Measured on the host, not inferred:
 
 Two things this did NOT do, and they are why the issue stays open:
 
-1. **The board has not booted it.** `MAIN_STACK_SIZE` on `mr_canhubk3/s32k344`
-   is the acceptance, and it needs the part.
+1. ~~**The board has not booted it.**~~ **MET 2026-08-31, recorded here
+   2026-09-03.** phase-409 carved the remaining members and the island entry
+   boots on `mr_canhubk3/s32k344` at `CONFIG_MAIN_STACK_SIZE=16384` — half the
+   32768 that was previously the smallest workable value
+   ([phase-409](../roadmap/phase-409-executor-inline-storage.md), Acceptance 3;
+   superproject commit `8d586b2`). Measured on the same two frames this issue
+   names: `Executor::open_in` 16000 → 2244 and `nros_cpp_init` 15104 → 1396,
+   so 31104 bytes of prologue on that call chain became 3640.
+
+   This criterion was written before that landed and stayed unticked for three
+   days, which is the failure mode this repo keeps recording in the other
+   direction: a HIGH-severity issue whose text overstates what is still broken
+   is as misleading as one that understates it.
 2. **Nothing STATES the number yet.** The boot-time check this issue asks for --
    a gate naming the main-thread stack an image requires, the way the heap gate
    names the arena -- is not written. The requirement is smaller now; it is still
--- a/docs/issues/0963-the-exported-bound-inventory-has-no-consumer.md
+++ b/docs/issues/0963-the-exported-bound-inventory-has-no-consumer.md
@@ -1,11 +1,38 @@
 ---
 id: 963
-title: "The derived-bound inventory is exported and nothing reads it, so every size downstream is still set by hand"
+title: "The derived-bound inventory has readers now — what is left is the executor arena alone (was: nothing reads it)"
 status: open
 area: codegen, memory, build
-severity: high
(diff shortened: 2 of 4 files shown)


jerry73204:
**Added #0963 — third issue in this cluster whose front matter contradicts its own body.**

Its title asserted that **nothing** reads the exported bound inventory. That stopped being true on 2026-09-02, and the body already said so in two dated sections. The front matter did not — so anyone reading the issue *list* saw the original claim at HIGH severity. A title is what triage sorts on; a correction buried in the body reaches nobody.

Readers that exist on main: `nros_derive_message_bound_knobs()` (`cmake/NanoRosMessageBounds.cmake:304`), joined against the entity inventory at `cmake/NanoRosCodegenCore.cmake:853-856`, and the Zephyr precedence ladder `_nros_resolve_derivable_knob()` (`zephyr/cmake/nros_cargo_build.cmake:276`) applying it to five knobs.

Retitled, and severity **high → medium**: of the three consumers the issue asked for, 1 and 3 are done (`afeab01d5`, `36622cadd`); item 2, the executor arena, is the only one left — and it is **in flight in #244, gated on #239**, not unowned.

## The scope finding, which matters more than the retitle

**0963 and 0965 are one seam.** 0963 supplies the per-TYPE size, 0965 the per-IMAGE entity list, and a knob needs the **join** — literally joined at `NanoRosCodegenCore.cmake:854-856`. Their remaining item is the same item. Treating them as two work streams would produce two half-answers.

**0961 is not part of it.** Stack, not arena — the issue explicitly rules the arena out (holding stack at 28672 while moving `NROS_EXECUTOR_ARENA_SIZE` 49152 → 40960 changed nothing). phase-409 severed the last coupling: `size_of::<Executor>()` is 1016 at both the shipped knobs and at `MAX_CBS=36`. Its remaining half needs a `size_of::<Executor>()` probe — a compile-time fact, not an inventory read.

I had assumed the opposite going in ("one seam, three symptoms"). The measurement says one seam of two, plus one independent issue.

## Net

All three now state which half of themselves is done. **None is closed** — 0961 owes a stack gate, 0963/0965 owe the arena. What changed is that none of them now sends a reader at work that already exists.

PR #258, opened 2026-09-03
fix(phase-412): make the knob plumbing checkable, not remembered

jerry73204:
**Stacked on #251.** Three structural weaknesses, each of which shipped a wrong-sized image while every gate was green. Done before wiring another knob, because every remaining knob rides the same plumbing.

## 1. The loader whitelist is gone

`_nros_load_derived_entity_inventory` re-exported a hand-maintained NAMED list, and it dropped a symbol **three times**:

| wave | dropped | consequence |
| --- | --- | --- |
| phase-412 W1 | four session pools | derived 10/14/0, session built **8/8/8** -- eight subscriber slots for ten subscriptions, a RUNTIME failure |
| phase-412 W2 | `MAX_NODES` | fell to a default |
| phase-403 step 3 | five per-kind counts | arena derived nothing while looking like it worked |

Each time the symbol loaded, died at the function boundary, and the consumer fell back to a default that looked deliberate.

The fragment is GENERATED and every line it emits is `set(<NAME> ...)`, so the symbol set is a property of the FILE rather than of anyone's memory. The loader parses it now. A list can fall behind a producer that adds a name, and did three times; a parse cannot.

## 2. `check-knob-fixpoint` -- issue #1002

#0991's remedy is "configure again" and everything in the tree says TWO passes. Two is not enough for a knob whose resolved value is cached: the fragment updates on pass 2 and the knob resolved from it on pass 3. Four island builds shipped the stale receive class, silently, because the stale value is the CLOSURE basis and therefore LARGER.

This does not count passes -- counting is the wrong instrument, since the count changes whenever a chain gets longer and being wrong about it looks identical to being right. It configures until nothing moves and fails if that takes more than allowed. A knob that APPEARS late counts as movement too: a consumer reading it on the previous pass got a default.

## 3. `check-knob-resolved-once`

Three instances in one phase: a knob gained a derivable resolution and kept its old plain one, the second call won, and what it passed was the raw Kconfig value -- the `-1` DERIVE SENTINEL, resolved as though an operator had stated it. `NROS_RMW_SUBSCRIBER_SLOTS` derived 10 and resolved -1; `NROS_EXECUTOR_MAX_NODES` derived 4 and resolved -1. Invisible to every other gate, because the derived value was RIGHT and the resolver DID run.

**Branch-aware, and that is not a refinement.** Its first real run flagged `ZPICO_TX_BATCH` and `ZPICO_TX_SPLIT_LOCK`, which are each resolved in the `if` arm and again in the `else` arm of one conditional -- exactly one call executes. Both false positives are pinned as self-test cases, alongside one asserting that a call outside a conditional plus one inside it IS still a defect. A checker that flags a correct idiom trains people to ignore it.

## Verified, not asserted

The parse-based loader delivers all four session pools plus the receive class on a clean island build, and `check-knob-delivery` is clean against it. Each new gate's self-test asserts the FAILURE it exists to catch, so a gate that stopped matching cannot report success -- seven cases for resolved-once, three for fixpoint.

`just check fast` **172/172**.


jerry73204:
Superseded by #297 — same branch, same work, rebased onto current `main`.

Same cause as #151: GitHub refuses to reopen a pull request whose recorded head sha is no longer the branch tip, and the branch was force-pushed with the rebase while this was closed. Recorded head was `769ce2e3c`; the branch now carries `26de57560`.

**Reopen first, then push the rebase.**

PR #259, opened 2026-09-03
measure(#0969): the service path's allocation count did NOT move — and that corrects me

jerry73204:
0969's own order ended *"re-run the allocation harness"*. Done, two ways, and the
answer is a **null result** that corrects the commit which landed the migration.

`service_roundtrip.cpp` and the `ros2_srv_{client,server}` pair gain
`NROS_ROUNDTRIP_ITERS` — the same knob and slope method `data_roundtrip.cpp`
uses, deliberately, so the number compares to the message path's 9.93 → 2.00.

## Measured

| configuration | before | after |
| --- | ---: | ---: |
| single process (loopback), per exchange | **4.00** | **4.00** |
| two processes, **client half only** | **4.14** | **4.15** |

**The migration did not change per-exchange allocations.** Not in loopback, not
across processes.

## What this corrects

The sertype-migration commit said *"per-message allocation is otherwise off the
service and action data path"*. **Not supported.** What I measured then was
`check-rmw-alloc-sites` going 2 → 1 — a count of **source sites**, not runtime
allocations. I let a static ledger stand for a runtime number, which is exactly
the substitution this campaign catches elsewhere.

What the migration *did* do still holds: a decode and an encode removed per
message, one steady-state site removed, five per-type adapters deleted. What it
did not do is reduce the allocation **count** — and the issue now says so.

## The more useful finding: the loopback harness under-reports

The two-process **client half alone** costs 4.15, while the single-process run
measuring **both halves** costs 4.00. A same-process harness prices roughly half
of what a deployment pays — Cyclone is free to hand a sample to a local reader by
reference, so the write path may never fully serialise.

Future service/action numbers should come from the two-process pair. The message
path's 9.93 → 2.00 was itself single-process and carries the same caveat on
absolute value.

**The null is not the instrument failing to see:** the same method moved by 7.93
on the message path and by 0.01 here.

## Not explained

4.00–4.15 is *fewer* allocations than the old path's sites predict (typed sample
+ ostream on each of request and reply, plus serdata). Something absorbs them, or
they are not on the measured path. Recorded as an open question rather than
guessed at.

## Scope

Defaults are 1 everywhere, so `ros2_srv_e2e` and `service_roundtrip` keep their
meaning; the server's deadline widened to cover an iterated run under valgrind.

Backend ctest **23/23**, `just check fast` **188 gates**.

### DIFF
--- a/docs/issues/0969-cyclone-take-cdr-round-trip.md
+++ b/docs/issues/0969-cyclone-take-cdr-round-trip.md
@@ -346,3 +346,69 @@ a regression is the shape this repo keeps retracting.
    (`tests/data_roundtrip.cpp`, `NROS_ROUNDTRIP_ITERS`) — the message path went
    9.93 → 2.00 allocations/message and this path still carries three sites, one
    per request and two per reply.
+
+## The allocation measurement, done 2026-09-03 — and it is a NULL RESULT
+
+Step 3 of the order above said "re-run the allocation harness". Done, both ways,
+and the answer corrects something I wrote when the migration landed.
+
+`service_roundtrip.cpp` and the `ros2_srv_{client,server}` pair gained
+`NROS_ROUNDTRIP_ITERS`, the same knob and the same slope method
+`data_roundtrip.cpp` uses — deliberately the same, because a service number
+measured a different way would not be comparable to the message path's
+9.93 -> 2.00.
+
+**Single process (client and server in one, the loopback case):**
+
+| | iters=1 | iters=200 | per exchange |
+| --- | ---: | ---: | ---: |
+| before the migration | 1,200 | 1,996 | **4.00** |
+| after | 1,178 | 1,974 | **4.00** |
+
+**Two processes (the case a control loop actually runs), CLIENT half only:**
+
+| | iters=1 | iters=200 | per exchange |
+| --- | ---: | ---: | ---: |
+| before | 1,445 | 2,269 | **4.14** |
+| after | 1,429 | 2,255 | **4.15** |
+
+**The migration did not change measured per-exchange allocations.** Not in
+loopback, not across processes.
+
+### What that corrects
+
+The commit that landed the sertype migration said "per-message allocation is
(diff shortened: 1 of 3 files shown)


jerry73204:
**Size sweep added — the migration is a CROSSOVER at ~8 KB, not a win or a regression.**

Allocation **count** is neutral at every size (4.14 → 4.15 at 12 B, 4.20 → 4.26 at 16 KB), so counting allocations was the wrong instrument. **Bytes** is where the behaviour lives:

| reply | before | after | delta |
| ---: | ---: | ---: | ---: |
| 16 B | 8,329 | 270 | **−8,059** |
| 1,008 B | 8,347 | 1,186 | −7,161 |
| 4,008 B | 8,354 | 4,342 | −4,011 |
| **8,008 B** | 8,325 | 8,219 | **−106** ← break-even |
| 16,008 B | 8,444 | 16,260 | **+7,816** |

**The old path is FLAT** — ~8,350 B/exchange whatever the payload, 16 B to 16 KB. A one-field reply cost the same as a 2,000-element sequence.

**The new path is LINEAR** — ~1.02× the payload on a ~260 B floor. The sample *is* the bytes.

Flat-versus-linear is the more useful half: a fixed 8.4 KB is easy to bound and impossible to shrink; a payload-proportional cost is derivable from the type's own `MAX_SERIALIZED_SIZE` (issue 0896), which is the shape phase-391's Robson bound wants. But a service whose replies exceed 8 KB now allocates **more**, and nothing warns — worth knowing before anyone quotes the round-trip removal as unconditional. The migration commit did; this corrects it.

**Method/limits:** `NROS_PROBE_SEQ_LEN` was a scratch `int64[] sums` reply used only to sweep size and is **reverted** — the committed harness keeps only `NROS_ROUNDTRIP_ITERS`. Two points per size, slope over 49 exchanges, client half only. Server half and the CPU cost of the removed decode/encode are unmeasured. 8 KB is an order of magnitude on this type and transport, not a threshold to encode.

PR #278, opened 2026-09-03
feat(phase-400 W6): five tenants — net, runtime, zenoh.wire, xrce, zenoh.limits

jerry73204:
Five smoltcp knobs — the TCP and UDP socket pools, the per-socket buffer, and the connect/socket timeouts — now resolve through the RFC-0049 ladder: `NetKnobs`, `[knobs.net]`, `BuildRungs::net_rungs()`, and **one** reader in `packages/drivers/net/nros-smoltcp/build.rs`.

## This tenant is why the reader moved

`nros-smoltcp` could not depend on `nros-board-common` — cargo counts an optional dependency when it looks for cycles, and the board crate reaches back through `nros-platform → nros-platform-esp32-qemu → nros-smoltcp`. **Every driver was locked out of the ladder.** #261 extracted the reader to `nros-platform-config` for exactly this; this is the first tenant to use it, and the build script says so where it names the crate.

## Two things the rung deliberately does not flatten

**`max_udp_sockets`'s builtin is feature-derived** — 1 brokered, 4 with `rtps` — and stays that way. A descriptor naming the knob outranks it (an explicit declaration beats a heuristic), but a board that says nothing still gets the feature-appropriate number rather than a constant.

**The deprecated `ZPICO_SMOLTCP_*` spellings rank *with* env**, above the rungs. That's what a deprecated front-end should do — it's still a user override — and it's why `net_env_key` lists only the `NROS_` names: nothing should start using the old ones.

## Verified against the build script's own emitted constants

```
builtin        1 / 1 / 2048 / 10000
platform rung  3 / 2 /  512 /   250
env wins       MAX_SOCKETS=4, the rest keep the rung
legacy alias   ZPICO_SMOLTCP_BUFFER_SIZE=99 also outranks the rung
```

Four rungs, each observed in `nros_smoltcp_config.rs` — not inferred from reading the code. `nros config explain` prints the tenant beside the others, and an env override shows as `env`.

## Census

| | before | after |
| --- | ---: | ---: |
| sizing backlog (W6) | 24 | **19** |
| on the ladder | 22 | 27 |

**Ownership checked first, as always:** `NROS_SMOLTCP_*` appears in no live roadmap or design doc — only four archived phases. Unclaimed, unlike three of the last four families I looked at.

Two leaf locks gain the serde/toml **build-dependency** closure — host-side only, and the same price `nros-params` and `nros-rmw-cffi` already pay to read the ladder.

## Verification

- `just check fast` — 190/190
- `just ci l1` — **passed** (first fully green tier 1 since #267 and #270 cleared main's reds)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2139,6 +2139,7 @@ name = "nros-smoltcp"
 version = "0.5.0"
 dependencies = [
  "nros-platform-api",
+ "nros-platform-config",
  "portable-atomic",
  "smoltcp",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1693,6 +1693,7 @@ dependencies = [
  "cc",
  "nros-build-paths",
  "nros-cc-flags",
+ "nros-platform-config",
  "nros-sizes-build",
 ]
 
@@ -2076,6 +2077,7 @@ version = "0.5.0"
 dependencies = [
  "cc",
  "nros-cc-flags",
+ "nros-platform-config",
  "nros-rmw-cffi",
  "nros-zephyr-build",
 ]
@@ -2100,6 +2102,7 @@ dependencies = [
  "nros-ghost-types",
  "nros-log",
  "nros-platform",
+ "nros-platform-config",
  "nros-rmw",
  "nros-rmw-cffi",
  "nros-tests",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1920,6 +1920,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-platform-config",
  "nros-platform-esp32-qemu",
  "nros-platform-mps2-an385",
  "nros-platform-stm32f4",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1621,6 +1621,7 @@ dependencies = [
  "nros-node",
  "nros-params",
  "nros-platform",
+ "nros-platform-config",
  "nros-rmw",
  "nros-rmw-cffi",
  "nros-serdes",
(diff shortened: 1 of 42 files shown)


jerry73204:
**Scope grew:** #285 (the `runtime` tenant — four component pools) auto-merged into this branch, since its base was this PR rather than `main`. This PR now carries **two** tenants:

1. `net` — five smoltcp knobs, the first driver to reach the ladder (the reason #261 extracted the reader)
2. `runtime` — four component pools; phase-391 *reads* them but never owned their values

Census across both: backlog **24 → 15**, ladder **22 → 31**.

Also corrected here: the first push of this branch carried a `third-party/dds/cyclonedds` **pin rewind**, staged by a `git add -u`. It would have unshipped *"ddsrt: name threads on Zephyr"*. `check-submodule-pins` caught it; the pin is restored to `79bd9cb9` and the gate reports 0 pins moved.

jerry73204:
**Scope grew again:** #288 (`zenoh.wire` — five wire sizes) auto-merged into this branch, as #285 did before it. Both had this PR as their base rather than `main`, and an unprotected base merges as soon as checks pass.

This PR now carries **three** tenants:

1. `net` — five smoltcp knobs, the first driver to reach the ladder
2. `runtime` — four component pools
3. `zenoh.wire` — five wire sizes; the two transport-band priorities deliberately excluded

Census across all three: backlog **24 → 10**, ladder **22 → 37** struct fields.

Two things worth a reviewer's eye, both written up in the individual commits:

- the first push of this branch carried a `third-party/dds/cyclonedds` **pin rewind** from a `git add -u`. Caught by `check-submodule-pins`, restored to `79bd9cb9`, and the gate now reports 0 pins moved.
- the `zenoh.wire` migration silently **deleted five rows** from the published static-pool inventory, because a knob that stops being read by a *call* becomes invisible to tools that scan for calls. Fixed in the same commit, and it recovered eight rows — three of them dropped earlier by the `memory` tenant and unnoticed until now.

PR #292, opened 2026-09-03
feat(phase-400 W6): the singletons, and the Zephyr heap whose blocker had already gone

jerry73204:
Stacked on #278. Five singleton knobs migrated across three crates; three deliberately not.

**Migrated:** `[knobs.xrce]` (`custom_transport_mtu`, `stream_history`), `[knobs.zenoh.limits]` (`keyexpr_string_size`, `subscriber_ring_depth`), and the LET buffer folded into `[knobs.runtime]`. Each had exactly one reader and no owning campaign.

## The three refusals are the substance

**`NROS_SERVICE_TIMEOUT_MS` has two readers by design.** `nros-rmw-zenoh/build.rs` emits a Rust const; `nros-build-helpers`'s C emitter emits a `#define`; a comment asks the next editor to keep the defaults equal.

I migrated it, and **`check-knob-single-reader` refused** — a migrated knob gets exactly one reader, and that rule is what stops the pair drifting. Both would have resolved through the same ladder and *could not* have disagreed, but the gate can't know that, and weakening it to say so trades a checked invariant for a comment. Migrating this knob means giving the pair **one emission point** first: a change to where the value is emitted, not to the ladder. Reverted, with the reason written at both sites.

**`NROS_ENTRY_SPIN_MS`** — read by a proc macro, the C++ library and a C header. Same shape, three readers, no single build script to own it.

**The two transport-band priorities** stay out for the reasons recorded with the `zenoh.wire` tenant.

## Verified against the emitted constants

```
builtin        KEYEXPR 256 / RING 4
platform rung  KEYEXPR 321 / RING 7
env wins       KEYEXPR 99, the rung still holding the rest
```

## The census caught my own helper

`env_usize_rung` was an idiom it didn't know, so it failed with *"a helper in neither list is counted as nothing"* — the gate I added this morning, working on the person who added it. Classified as a reader.

## Census

| | before | after |
| --- | ---: | ---: |
| sizing backlog (W6) | 10 | **5** |
| ladder struct fields | 37 | 42 |

## Verification

- `just check fast` — 190/190
- `just ci l1` — passed
- `just check submodule-pins` — 0 moved, checked deliberately and staged by explicit path rather than `git add -u`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1693,6 +1693,7 @@ dependencies = [
  "cc",
  "nros-build-paths",
  "nros-cc-flags",
+ "nros-platform-config",
  "nros-sizes-build",
 ]
 
@@ -2076,6 +2077,7 @@ version = "0.5.0"
 dependencies = [
  "cc",
  "nros-cc-flags",
+ "nros-platform-config",
  "nros-rmw-cffi",
  "nros-zephyr-build",
 ]
@@ -2100,6 +2102,7 @@ dependencies = [
  "nros-ghost-types",
  "nros-log",
  "nros-platform",
+ "nros-platform-config",
  "nros-rmw",
  "nros-rmw-cffi",
  "nros-tests",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1920,6 +1920,7 @@ version = "0.5.0"
 dependencies = [
  "nros-platform-api",
  "nros-platform-cffi",
+ "nros-platform-config",
  "nros-platform-esp32-qemu",
  "nros-platform-mps2-an385",
  "nros-platform-stm32f4",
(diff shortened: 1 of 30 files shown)


jerry73204:
**Second commit added:** `NROS_ZEPHYR_HEAP_SIZE` joins the `memory` tenant.

Its census entry had said *"BLOCKED: the ladder crate depends on `nros-platform`, so the rungs are a cycle away"* for two waves, and `nros-platform/build.rs` said the same thing in its own words — *"a real refactor, not a dependency line"*. **That refactor is #261**, extracted for the `net` tenant. Both notes went stale the moment it landed, and neither was updated, because nothing re-reads a *reason* once it's written.

The `memory` tenant had already mapped the knob (`("zephyr", "heap_bytes") => "NROS_ZEPHYR_HEAP_SIZE"`), so the rung existed while its only reader couldn't consult it — correct, tested, unreachable.

Verified: builtin 65536 → platform rung 131072 → env 4096. The bare `knob_usize` call survives as the no-platform fallback, so out-of-tree consumers are unchanged.

Backlog **5 → 4**.

PR #303, opened 2026-09-03
fix(#1025): pack the ESP32 flash image from the row the build actually wrote

jerry73204:
No ESP32 QEMU flash image could be packed **at all**. The ELF built fine; the packer read `build/cargo-fixtures/qemu-esp32-baremetal/` while the build wrote `qemu-esp32-baremetal-4118800323`.

The shared cargo group key is a function of **(platform, cargo args, env)**, and the packer passed the last two **empty** — one function, two questions, diverging exactly when a row has a variant. Live since `41a7d8de7`, a *correct* DRAM fix whose `env = { ZPICO_MAX_QUERYABLES = "2" }` gave those rows their first variant.

**A recurrence at the site the preventing helper was written for.** `nros_fixture_row_artifact_dir`'s docstring is about `just esp32 build-qemu` hand-spelling this path: *"a post-processing step that spells the artifact path by hand is a second derivation of the group key; this makes it one."* The call site was duly converted — phase-340 item 7 made the **formula** single, and the **inputs** were still derived twice.

## Fix

`nros_fixture_row_artifact_dir_by_id <row-id> <platform>` reads the row's args and env from the manifest — the one place they're written. Both packers now name the row id they pack, the same id the build loop above each already passes to `fixtures-build.sh --id`.

**Acceptance is the build**, which the helper's docstring demands and which a gate is what passed last time:

```
just esp32 build-qemu           → esp32-qemu-talker.bin, esp32-qemu-listener.bin
just esp32 build-logging-smoke  → logging-smoke-esp32-qemu.bin
```

## The sweep — and two of the five other flags were the gate being wrong

`check-fixture-artifact-dir-inputs`'s first version flagged six sites. **Only the two ESP32 ones were defects.** The freertos / nuttx / qemu-arm-baremetal / threadx-linux / threadx-riscv64 run-example recipes *build the artifact themselves* and pass the same empty arguments to `nros_fixture_target_dir_flag`, so their halves agree — they write and read the same unhashed dir.

Measured rather than argued: `nuttx` and `threadx-riscv64` genuinely diverge from the fixture lane's key (`nuttx-2965781523` vs `nuttx`), which is what makes them *look* like defects — and nothing in those recipes reads what the lane built.

Shipping that version would have reported a five-site bug that does not exist: a gate wider than the rule it enforces, which is issue 0196's audit finding, about to be repeated. The gate now exempts a call whose recipe also calls `nros_fixture_target_dir_flag`, and its **selftest runs three controls on the normal path** — the 1025 defect must flag, the by-id fix must not, a build-it-yourself recipe must not. The third pins the narrowing so a later edit can't quietly widen it back.

Sweep command: `python3 scripts/check-fixture-artifact-dir-inputs.py`

## What this does not establish

Whether issue 0968's five esp32 tests **pass**. They could not run at all before this; can-run is not passes. Running them is 0968's work.

Gates: `check fast` (191, including the new one).

### DIFF
--- a/docs/issues/1065-esp32-build-qemu-recipe-has-no-lane.md
+++ b/docs/issues/1065-esp32-build-qemu-recipe-has-no-lane.md
@@ -1,0 +1,69 @@
+---
+id: 1065
+title: "`just esp32 build-qemu` is run by no lane, so it can be wholly broken with every test green"
+status: open
+area: testing
+severity: medium
+related: [1025, 0196, 0883]
+---
+
+## What
+
+`just esp32 build-qemu` is the documented way to produce the ESP32 QEMU flash
+images, and **nothing in CI runs it**. Grepping the whole tree for callers finds
+only `just docker build-qemu` (a different recipe, in the Docker module) and
+prose references.
+
+The ESP32 e2e tests do not use it. `packages/testing/nros-tests/tests/
+esp32_emulator.rs` resolves the ELF through the fixture resolver and packs its
+own image:
+
+```rust
+let elf = build_esp32_qemu_talker().expect("Failed to build esp32-qemu-talker");
+let flash_image = nros_tests::build_dir(nros_tests::kind::ESP32_QEMU, &[])
+    .join("esp32-qemu-talker.bin");
+create_esp32_flash_image(elf, &flash_image).expect("Failed to create flash image");
+```
+
+So the tests exercise a SECOND packing path that happens to be correct, and the
+recipe a human is told to run is exercised by nobody.
+
+## Evidence that this is not theoretical
+
+Issue 1025: the recipe's ELF lookup passed `"" ""` for the row's cargo args and
+env, so it read `build/cargo-fixtures/qemu-esp32-baremetal/` while the build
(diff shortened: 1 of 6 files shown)


jerry73204:
This PR is load-bearing for #1052 and worth prioritising.

While bisecting the ESP32 talker fault I ran three variants that all faulted identically, and concluded from that the fault was independent of what the node registers. It wasn't — **all three were the same binary**. The packer this PR fixes resolved the ELF from invented empty `cargo_args`/`envstr`, so it looked in bare `qemu-esp32-baremetal` while cargo wrote to the variant group dir `qemu-esp32-baremetal-4118800323` (the talker row carries `ZPICO_MAX_QUERYABLES=2`). A stale ELF at the bare path got packed on every "rebuild".

With the by-id helper's behaviour applied by hand, the real result inverts: empty `register` → 0 faults, full talker → 2, matched control, same session.

So this isn't only a tidiness fix — on `main` today it silently produces measurements of the wrong binary, and it cost me a wrong conclusion I published before catching it. The loud failure that let me catch it comes from #700, not from here.

PR #315, opened 2026-09-04
docs(#1029, phase-196): the Zephyr nightly reports nothing, and a phase waits on it

jerry73204:
Picked up **phase-196** (CI bring-up, "In progress" since 2026-05-28) to check it, and it turned into a live finding.

## The phase's work is done; a deleted workflow holds it open

33 of 34 boxes are checked, 196.1–196.9 all marked, and the 196.7 `[P2]` item's three boxes are all `[x]`. What holds it open is **one acceptance criterion naming a workflow that no longer exists.**

phase-253 (`682706dbf ci: merge 13 workflows into 6 tier files`) consolidated the CI topology this phase was written against. **All twelve** workflows named in its acceptance are gone — `ci.yml`, `zephyr-dual-line.yml`, `platform-ci.yml`, `lint.yml`, `host-unit.yml`, `codegen-convention.yml`, `sdk-index-gate.yml`, `nros-acceptance.yml`, `dep-chain.yml`, `scaffold-journey.yml`, `embedded-feature-unification.yml`, `colcon-parity.yml`.

## The criterion survived — but it reports nothing

"`zephyr-dual-line` is green end-to-end on both lines" is now the zephyr matrix in `nightly.yml`, on its own `0 5` cron. Every zephyr job in the last **eight consecutive scheduled runs** is `skipped`:

```
09-03T07:12  09-03T05:40  09-02T07:13  09-02T05:11
09-01T07:12  09-01T05:12  08-31T07:18  08-31T05:12
```

all `skipped,skipped,skipped`. Runs that **do** produce zephyr verdicts (05:34, 01:19, 01:45) are off the schedule cadence and report 22 job results each — so the jobs work, and the scheduled trigger is what yields nothing.

The 07:00 skips are correct by design: two crons, `0 5` Zephyr and `0 7` platforms, each gated to its own family. The finding is the **05:00** column.

This is worse than a red lane, and it is the shape CLAUDE.md already names. A red lane has a verdict to compare against; a uniformly **skipped** one makes a Zephyr regression and a healthy Zephyr indistinguishable — and neither distinguishable from the cron not firing.

## Filed WITHOUT a root cause, deliberately

`dorny/paths-filter` warns on schedule that *"'before' field is missing in event payload — changes will be detected from last commit"*, which would explain the gate closing. **But** the `set` step assigns `zephyr="true"` unconditionally for any non-`pull_request` event, which contradicts that, and the step exited `success`.

I could not resolve the contradiction, because `gh run view --log` returns mangled output for this workflow — steps labelled `UNKNOWN STEP`, and a grep for a line that step prints *unconditionally* found nothing while a later line from the **same** `echo` did appear. A cause read off a log I have shown to be lossy is issues 0859–0862 all over again, so I stopped.

Issue **1029** records the measurement, the lead, and what would settle it — read `steps.set.outputs` from the API, or have the `changes` job write its computed values to `$GITHUB_STEP_SUMMARY`, which survives log formatting. That is worth doing regardless: a gate whose inputs are only observable through a lossy log is a gate nobody can debug, which is how this went unnoticed for eight runs.

`just check fast` green.

### DIFF
--- a/docs/issues/1029-zephyr-nightly-never-produces-a-verdict.md
+++ b/docs/issues/1029-zephyr-nightly-never-produces-a-verdict.md
@@ -1,0 +1,98 @@
+---
+id: 1029
+title: "The Zephyr dual-line nightly has its own 05:00 cron and every scheduled run SKIPS it, so the lane it exists to watch has produced no verdict for days"
+status: open
+type: bug
+area: ci, testing
+severity: medium
+found: 2026-09-04
+related: [0871, 0872, 0994, 0996, phase-196, phase-253, phase-413]
+---
+
+# A lane with a cron, a gate, and no answers
+
+## Measured
+
+Every zephyr job in the last **eight consecutive scheduled** `nightly.yml` runs
+is `skipped`. Read from the jobs API, not from logs (see the caveat below):
+
+    2026-09-03T07:12   skipped,skipped,skipped
+    2026-09-03T05:40   skipped,skipped,skipped
+    2026-09-02T07:13   skipped,skipped,skipped
+    2026-09-02T05:11   skipped,skipped,skipped
+    2026-09-01T07:12   skipped,skipped,skipped
+    2026-09-01T05:12   skipped,skipped,skipped
+    2026-08-31T07:18   skipped,skipped,skipped
+    2026-08-31T05:12   skipped,skipped,skipped
+
+The three jobs are `zephyr-example-matrix`, `zephyr-dual-line-summary` and
+`zephyr-copy-out`.
+
+**Runs that DO produce zephyr verdicts are not on the schedule cadence** —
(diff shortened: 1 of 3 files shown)


jerry73204:
Closing in favour of #320, which now contains this work in full.

#320 was opened stacked on this branch, then retargeted to `main` so it could enter the merge queue. That made it a strict superset — it carries both commits:

```
116f5e09d  fix(#1029): the zephyr cron gate compares a value nobody could see
c6f265fb0  docs(#1029, phase-196): the Zephyr nightly reports nothing, and a phase waits on it
```

Verified by merge-base diff rather than by eye — #320 contributes exactly four files (`nightly.yml`, the 1029 issue, its index row, and the phase-196 correction) and nothing else.

Keeping both queued would have merged the same commit twice and left whichever landed second empty, so this one goes.

PR #329, opened 2026-09-04
RFC-0087 + phase-417: adopt ROS 2's API shapes, and refuse the rest loudly

jerry73204:
Implements **RFC-0087** (new) via **phase-417** (new): what has to be true before our
user API can carry ROS 2's own names and `rclcpp_compat.hpp` can be deleted.

## The rule

> An upstream name may be adopted only if its observable contract is the same, or
> strictly weaker in a documented, non-inverting way. A contract that inverts, or
> silently drops data or configuration, must fail to COMPILE. **Never compile and differ.**

Phase-379 W6 found `PollingSubscription::take()`: rclcpp's name, rclcpp's signature,
the opposite contract — `while (sub.take(msg))` terminates upstream and spun forever
here, with no compile error. That was one method found by accident. Taking ROS 2's
names across the whole API generalises it to every item whose semantics differ.

Four dispositions: **adopt** / **adopt-bounded** (weaker within a stated envelope) /
**refuse-loud** (`= delete` or `static_assert` naming the constraint *and* the
alternative) / **absent**. All 717 uncovered rclcpp items are classified: 117 / 123 /
69 / 408. The 69 refusals collapse into **17 diagnostics**, because a refusal is
per-concept, not per-symbol.

**Ordering:** the rename is cheap and cosmetic; the compatibility is the work. It
lands last — renaming first would relabel the gaps without closing one.

## Three live inversions, all shipping before this PR

| | was | upstream |
|---|---|---|
| `ParametersQoS` | `QoS(10)` | `KEEP_LAST, 1000` |
| `SystemDefaultsQoS` | `QoS(10)` = a *different* profile | all-sentinel; 0829 measured 1 on Cyclone, 42 on zenoh |
| 10 `NodeOptions` setters | stored, never read, returned `*this` | actually configure |

A rename would have inherited all three. They are now transcribed from
`rmw/qos_profiles.h` or refused, with 20 `static_assert`s pinning the values in every
`check cpp` rather than a runtime test no embedded lane executes.

## The measurement was measuring the wrong thing

`ours_cpp` read three TUs and filtered to namespace `nros`, so `rclcpp_compat.hpp` —
589 lines whose entire purpose is drop-in compatibility, and what ported code actually
reaches — contributed **zero rows** (issue 1020). The lane now reports two surfaces
from one ledger.

And fixing it made the numbers *better while the defects were unchanged*:
`ParametersQoS` and `NodeOptions::use_intra_process_comms` now correlate `same`.
Correlation compares names and shapes; neither differs. **The instrument cannot see
the defect it measures** — which is why `disposition` is load-bearing, not bookkeeping.

## The rule gained a refinement

Read literally, "must fail to compile" rejects every caller of `rclcpp::init(argc, argv)`
— including the common embedded one with no ROS arguments — and makes upstream's
tutorial `main` unportable, colliding with stage 1's acceptance. So:

> A refusal fires at the **earliest point the defect is knowable**. Signature or type
> carries it → compile time. Only the *value* carries it → the call.

The predicate is `constexpr` so its cases are `static_assert`ed, not hidden in an abort
only a dying process could observe. Mutation-tested both ways it would silently stop
refusing.

## What landed

- **Stage 0** — the lane sees the shim; `disposition` field, gate behind `--require-disposition`.
- **Stage 1** — nested `SharedPtr` on six entity types, `FixedString`↔`std::string`,
  clock/identity forwarders. **Acceptance met:** `cpp-port-minimal-publisher` is
  upstream's file again (its README claimed "verbatim" while three lines differed).
- **Stage 2/2b** — parameters, `create_service`/`create_client`, `Rate` as a forwarder
  onto `nros::spin`; the whole graph surface forwarded from `Node`/`LifecycleNode`.
- **Stage 3** — the loudness pass, 4 refusals each with its own diagnostic.
- **Stage 4** — executor `cancel`/`is_spinning`, actions `GoalUUID` + terminal verbs,
  logging convergence (named loggers, levels, sinks, throttle, real file/line/timestamp).
- **Corrections** — issues 1012 and 1022 resolved, ~140 rows.

## Findings worth reading

- **`halt`/`is_halted` were `alloc`-gated**, so a core-only image could not stop its own
  executor. The `Arc` needed the allocator; the bit did not.
- **`nros-c` never enabled `nros-log/platform-clock`** — C timestamps stayed `0` and the
  throttle admitted everything: 40 of 40 records vs 5 of 40 with it.
- **The `cpp_api_drift` fixture was guarding the defect** — it asserted the inert
  `NodeOptions` chain kept *compiling*.
- **Five logging macros were verdicted `extension` and are `rename`s** — rclrs ships
  `log_info`/`log_debug`/…, and the correlator pairs on *name*, so `nros_info` never
  matched.

## Still open

Stage 5 (C typed delivery, `rcl_ret_t` mapping) and stage 6 (the rename). Stage 6 is
additionally gated on a structural blocker no diagnostic can cover: the shim `Node`
pumps its own callbacks while `nros::Node` is arena-driven, and a file mixing them gets
zero callbacks and no error. Today the different *names* are the only thing making that
visible — after the rename they would not be.


jerry73204:
**RFC number collision — 0087 is taken on `main`.**

This PR adds `docs/design/0087-ros2-api-adoption-and-the-compile-or-conform-rule.md`, but `0087-package-identity-and-provider-format.md` merged to `main` in #367 (~10:47Z today), and `0088-serialization-format-is-a-compile-time-provider.md` with it. Both are on `main` now, so this branch will conflict on `docs/design/README.md` and land a duplicate number.

`0089` is free as far as I can see: `main` holds 0085–0088, and of the other open PRs touching `docs/design/00*`, #338 only edits 0086.

Renumbering is this repo's usual remedy for the class (`docs: renumber — phase 341 -> 345, issues 0467-0471 -> 0492-0496`), and it is cheaper here than on `main`.

Root cause, and it is not either author's mistake: **RFC ids have no reservation.** `just issue-new` and `just phase-new` claim `refs/{issue,phase}-ids/NNNN` on origin precisely because reading-then-writing the highest number is a race that has collided seven times for issues. `docs/design/README.md` says only "next free number", and two sessions read the same free number hours apart. Filed as **issue 1053**.

(Found while landing phase-420/421.)

PR #334, opened 2026-09-04
docs(#0870): the zenoh-pico pin advance is not the cause — 7/7 with the old pin

jerry73204:
Docs-only. Two measurements on #0870, one of which refutes a theory I wrote into a phase-414 report the same day.

## The zenoh-pico pin advance is NOT the cause

The standing theory was that the failure died with the 2026-08-31 bumps (#0899, the lease task freeing the transport under the publisher; #0906, sessions dropped and rebuilt on a ~20 s lease beat), because the last failing artifact was built 2026-08-30. That was attribution by **timeline and mechanism, with the bisect explicitly not done** — and it said so.

Done now:

```
git -C packages/rmw/zenoh/zpico-sys/zenoh-pico checkout a6affe8e61e5   # the pin
        # immediately BEFORE 107002a0c, which took the #0899/#0906 line
just nuttx build-fixtures-arm
cargo nextest run … -E 'test(/action_e2e.*Nuttx.*Cpp/)' --retries 0
```

| pin | result |
| --- | --- |
| `a6affe8e61e5` (pre-#0899/#0906) | **7 PASS, 0 FAIL** — 25.5–29.7 s |
| `fa7ad0f5b3f6` (restored) | **3 PASS, 0 FAIL** — 26.0–29.7 s |

p = (1/3)^7 ≈ 0.0005 at this issue's own 2-in-3 rate. Not a museum binary: the resolved artifact is dated after the checkout (`2026-09-04 10:33`), verified per the 0859–0862 rule before drawing the conclusion.

**The failure mode was already gone before the pin moved**, so the search should not start at zenoh-pico. Bound on the claim, stated in the issue: this pairs old zenoh-pico with the *current* harness, so "some harness change" survives as a candidate — though the cell is already recorded failing ~2-in-3 *after* the #0867 ordering fix, which was the main one.

## The C-vs-C++ asymmetry is real but about half the figure on file

The issue records "C: 26.4–27.2 s, C++: 44–50 s" as the first hard evidence of asymmetry. That compared a **cold** C++ run with **warm** C runs. Same host, same session, same build, all `--retries 0`:

| cell | first run | subsequent |
| --- | --- | --- |
| C++ | **45.3 s** | 25.5, 26.2, 25.8, 28.7, 25.5 |
| C | — | 18.2, 16.9, 16.8 |

Two effects were being read as one — cold start costs ~19 s on a cell's first run (page cache and QEMU warm-up, not a language property), and C++ is genuinely slower by ~9 s (≈1.5×), not ~20 s. Absolute numbers differ from the earlier session's host, so only the within-session gap is quotable, which is how the mix went unnoticed.

Recorded as **hypothesis, not finding**: if cold runs cost ~19 s more and the historical failures skewed to first runs, a deadline-shaped fault would preferentially strike cold — which would explain a rate nobody can reproduce warm.

## Composes with the concurrent work on main

`61489d5c4` landed while I was measuring. It kills the pool-exhaustion lead *structurally* (no exhaustion anywhere in the path can arrive as `Generic`) and narrows the fault to `z_declare_subscriber()` returning < 0, most likely its TX exit. I read it before appending and kept **both** sections rather than choosing: its narrowing plus this refutation compose — the fault is in that TX exit, and it is not explained by the zenoh-pico version.

**Unchanged:** nobody has read a failing run since the diagnostics landed, so the six-of-six question (all five liveliness declares plus the feedback subscriber, or only the subscriber?) is still open. Retries are 0 on this cell now, so CI will produce one.

Housekeeping: the zenoh-pico pin was restored and verified against HEAD, and the fixtures rebuilt against it.

### DIFF
--- a/docs/issues/0870-nuttx-cpp-action-client-transport-tx-failed.md
+++ b/docs/issues/0870-nuttx-cpp-action-client-transport-tx-failed.md
@@ -477,3 +477,70 @@ to ask for four rounds. RAM cost: none.
 `nsh_main` before `main`, and `nsh_main` is in the built ELF. Without that the
 four `nros_error!` calls would sit in the pre-init ring and never print. The
 plan is sound; only the binaries are wrong.
+
+
+## 2026-09-04 — A/B against the OLD zenoh-pico pin: the pin advance is NOT the cause
+
+The standing theory — mine, written into a phase-414 report the same day — was
+that the failure died with the zenoh-pico pin bumps of 2026-08-31 (#0899, the
+lease task freeing the transport under the publisher; #0906, every session
+dropped and rebuilt on a ~20 s lease beat), because the last failing artifact
+was built 2026-08-30. That attribution was by TIMELINE and MECHANISM, with the
+bisect explicitly not done. **It is now done, and it refutes the theory.**
+
+Method: `git -C packages/rmw/zenoh/zpico-sys/zenoh-pico checkout a6affe8e61e5`
+— the pin immediately BEFORE `107002a0c` ("bump to the nano-ros patch line
+carrying the #0899 and #0906 fixes") — then `just nuttx build-fixtures-arm`, then
+the cell with `--retries 0`. The rebuild is real, not a museum binary: the
+resolved artifact `examples/qemu-arm-nuttx/cpp/action-client/build-zenoh/
+cpp_action_client` is dated 2026-09-04 10:33, after the checkout.
+
+**Result: 7 runs, 7 PASS, 0 FAIL.** At this issue's own 2-in-3 failure rate that
+is p = (1/3)^7 ≈ 0.0005.
+
+So the failure mode was ALREADY GONE before the pin moved. Whatever fixed it is
+something else, and the search should not start from zenoh-pico.
+
+One confounder, stated because it bounds the claim: this build pairs the OLD
+zenoh-pico with the CURRENT harness, whereas the historical failures had both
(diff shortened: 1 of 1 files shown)


jerry73204:
Closing unmerged — both findings here are superseded by stronger work that landed on main while this sat in the queue.

**The pin A/B.** This PR reported 7 passes on the pre-#0899/#0906 pin (`a6affe8e`) to refute the attribution that the pin advance fixed #0870. Main now carries **210 runs on the old pin, all green** — the same refutation at thirty times the sample size, and reached independently.

**The C-vs-C++ asymmetry.** This PR corrected the figure from ~20 s to ~9 s and attributed the difference to cold-vs-warm measurement. Main has the actual mechanism: the gap is an **emulator stall** (issue 1034), correlated exactly with `.bss` program-header shape across all 19 NuttX arm images, and on QEMU 6.2.0 the C++ client completes in 2.2 s with no asymmetry left. That is a cause; mine was a timing observation. Theirs is right and supersedes it.

Merging this would add a weaker, partly contradictory account beside a better one.

Nothing here is lost: both conclusions are already on main in stronger form. The only thing worth carrying forward is the method note, which is generic enough to live elsewhere — the A/B was only possible because the submodule pin could be rewound and rebuilt, and the rebuild had to be *proved* (via the generated header's mtime) rather than assumed.

This is the eighth issue this session solved twice in parallel. That coordination cost is worth more attention than any single fix in it.

PR #336, opened 2026-09-04
docs(phase-419): a phase that contradicts itself should not need a reader to notice

jerry73204:
Proposes the periodic roadmap verification pass, written **before** any gate code so the acceptance exists first.

## The finding it is built on

A sweep of the 13 oldest open phases on 2026-09-04 found **8 whose status line had outlived its content**, 1 verified accurate, 4 not verified. At eight in thirteen, a reader cannot use a status line without re-deriving it — which is the same as not having one.

The cost is not tidiness. phase-196's one open acceptance criterion named `zephyr-dual-line.yml`, a workflow phase-253 had deleted, so it could be neither met nor refuted. Chasing that down surfaced issue 1029 — **eight consecutive scheduled nightly runs producing no zephyr verdict at all** — which nobody was looking for.

## Two layers, because the findings split cleanly

`scripts/check-roadmap-status.sh` already exists, is on the fast line, and **deliberately declines this job**: *"staleness of the CONTENT is a human call."* That is right for some of what was found and wrong for the rest, and the boundary is sharp:

- A header saying "nothing landed" over a body saying `LANDED` is the document contradicting **itself** — no judgment involved.
- A phase superseded because a later one reversed its direction is judgment and nothing else.

Sorted: **7 mechanical, 4 judgment.**

## W2's rule is measured, not proposed

A naive R1 flags 12 phases — including `phase-375`, whose header reads *"PROPOSED — W0 landed, W1–W5 not started."* That is honest self-disclosure, and **a gate that flags honesty is one people switch off.** Excluding headers that admit progress leaves five:

```
phase-230  ticked=7   (confirmed stale, corrected today)
phase-236  ticked=7   (confirmed stale, corrected today)
phase-292  ticked=17  "Draft", 17 of 18 boxes ticked
phase-325  ticked=14  "Draft. Not started." + 14 ticked
phase-206  ticked=0   weak — likely prose
```

**292 and 325 are live findings the rule surfaced** and nobody has looked at.

### The rule caught its own proposal

Recorded in the doc because testing it against this very phase found it: **`landed-marks` alone is not a usable signal.** This document says "nothing landed" in its header and contains 13 occurrences of `LANDED` — every one prose about *other* phases. A first cut that counted words would have flagged the phase proposing the gate, on its first run. So the rule keys on ticked **checkboxes**, a structural claim about the document's own work items.

## Sequencing and scope

**W1 measures overlap first** with `check-doc-refs` / `check-doc-recipe-refs` / `check-ci-doc-workflow-refs`, and extends rather than duplicates — two gates answering one question is how the api-parity pair came to disagree by 25 symbols.

**W3** is the monthly agent pass for the judgment column: output an issue or a PR, **never a red gate** (the #0975 deadlock; `pr-verdicts.yml` documents the shape).

**Explicitly out of scope**, and stated in the doc: judging whether a phase's *plan* is still a good idea (phase-211's status is accurate and its premise is dead — only the second is in scope; adopting Creusot is a maintainer's call), and auto-correcting a status line (every correction today needed its reason written down, and a generated line carries the claim without the evidence).

## Note

The two references to issue 1029 are plain text, not links — 1029 is filed in #320 and not yet on `main`, and `check-doc-refs` correctly refused the push until I de-linked them. Worth linking once #320 lands.

`just check fast` green; `check-roadmap-status` passes on the new doc.

### DIFF
--- a/docs/roadmap/phase-419-roadmap-claim-verification.md
+++ b/docs/roadmap/phase-419-roadmap-claim-verification.md
@@ -1,0 +1,174 @@
+# Phase 419 — a roadmap phase that contradicts itself should not need a reader to notice
+
+**Status (2026-09-04). PROPOSED — nothing landed. The measurements below are
+real; the work items are not started.**
+
+## The finding
+
+A sweep of the 13 oldest open phases on 2026-09-04 found **8 whose status line
+had outlived its content**, 1 verified accurate, and 4 not verified. That ratio
+is the argument for this phase: at eight in thirteen, a reader cannot use a
+status line without re-deriving it, which is the same as not having one.
+
+What each cost is not tidiness. phase-196's single open acceptance criterion
+named `zephyr-dual-line.yml`, a workflow phase-253 deleted — so the criterion
+could be neither met nor refuted, and the phase sat "In progress" for three
+months waiting on a lane that had been consolidated into `nightly.yml`. Chasing
+that down is what surfaced issue 1029 (`1029-zephyr-nightly-never-produces-a-verdict`, filed in PR #320 and not yet on `main` — link it once that lands)
+(eight consecutive scheduled runs producing no zephyr verdict at all), which
+nobody was looking for.
+
+The corrected phases: 403, 296, 209, 356, 357, 196, 230, plus 275 misfiled as a
+phase at all, and 41's stale paths. [phase-162](phase-162-rt-scheduling-harness.md)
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #342, which carries this doc plus the implementation of W1, W2 (R1/R3/R4) and W3.

This PR was queued, and a queued branch cannot be updated, so the implementation went to a new branch rather than fighting the lock.

PR #342, opened 2026-09-04
feat(phase-419): the roadmap verification pass — W1, W2 (R1/R3/R4), W3

jerry73204:
Supersedes #336 — same phase doc, plus the implementation. #336 was queued, which locks its branch, so this carries both commits.

## W1 — overlap measured first, as the phase requires

- `check-doc-refs` owns the numbered doc series, including the `archived/` fallback.
- `check-ci-doc-workflow-refs` has the workflow-reference logic but a **hardcoded three-file scope** (`docs/development/ci-*.md`). Roadmap docs are covered by **neither** — 20 dead workflow references across 8 phases, **11 in phase-196 alone**, one of which was its single open acceptance criterion.
- R1, R3, R4: not covered by anything.

## W2 — `scripts/check-roadmap-claims.py`

Buildless, offline, self-testing on the normal path, ratcheted, fast line.

### Two rules changed shape because the gate flagged its own documentation

This is the most useful thing that happened:

- **R1 keys on ticked checkboxes, not the word `LANDED`.** phase-419's doc says *"PROPOSED — nothing landed"* in its header and contains **13 occurrences of `LANDED`**, every one prose about *other* phases. A word-counting first cut flagged the phase proposing the gate.
- **R4 searches the status field, not the document.** Searching the body flagged phase-419 again — the phase that *specifies* R4 quotes `NOT A WORK-ITEM PHASE` while describing it. The same trap `check-workflow-repo-env` names: a gate that cannot tell a claim from a sentence about a claim.

Both pinned as self-test cases so neither rule can widen back.

R1 also excludes headers that **admit** progress, or it flags phase-375 (*"PROPOSED — W0 landed, W1–W5 not started"*) — honest self-disclosure, and a gate that flags honesty is one people switch off. And `header_block` stops at the next bold **field**: phase-325 runs Status / Implements / Successor to / Informed by / Blocked on as contiguous lines, and reading its *"W0–W2 are done"* (in `**Blocked on:**`) as part of the status hid a phase whose status says *"Draft. Not started."* over **14 ticked boxes**.

### The baseline is findings, not debt

```
phase-230  R1   already corrected in #322 — line deletes when it lands
phase-275  R4   same
phase-292  R1   "Draft", 17 of 18 boxes ticked — LIVE, unworked
phase-325  R1   "Draft. Not started." + 14 ticked — LIVE, unworked
```

When #322 lands, the ratchet will report those two entries "no longer offend" and ask for their deletion — the behaviour that caught me earlier today when archiving phase-275 stranded a `doc-recipe-refs` baseline line.

### R2 not done, with its cost measured

Extending `check-ci-doc-workflow-refs` means giving it a baseline it does not have, for 20 references most of which are legitimate history. Reported by W3 instead.

## W3 — `just roadmap-audit` + a monthly `roadmap-audit.yml`

A report, never a gate, never on `pull_request` (a scheduled job in the required set is the #0975 deadlock). Writes to `$GITHUB_STEP_SUMMARY` because issue 1029 was eight nights of a lane whose deciding value existed only in a log this workflow family renders lossily.

**Its heuristics were tuned against noise rather than shipped as first written.** The first run flagged **56 of 63** phases — 89% is not a filter. Restricting cited paths to our own top-level roots, excluding the numbered series `check-doc-refs` already owns, dropping `tests/` and `config/` (relative fragments, not repo roots), and refusing to let **age alone** qualify a row brought it to **29 of 63**. Rows sort strongest-signal-first; phase-196 with its 11 dead workflow refs leads.

`just check fast` green, 217 gates.

### DIFF
--- a/.config/roadmap-claims-baseline.txt
+++ b/.config/roadmap-claims-baseline.txt
@@ -4,5 +4,3 @@
 # the line.
 docs/roadmap/phase-230-platform-layer-split.md	R1
 docs/roadmap/phase-275-276-branch-notes.md	R4
-docs/roadmap/phase-292-asi-reference-consumer-revisit.md	R1
-docs/roadmap/phase-325-uorb-interop-and-bridge.md	R1
--- a/.config/roadmap-claims-baseline.txt
+++ b/.config/roadmap-claims-baseline.txt
@@ -1,0 +1,8 @@
+# phase-419 W2 — roadmap phases whose claims contradict their own
+# body. RATCHET: this list may only SHRINK. Each line is a finding
+# for a person to work, not debt to hide; fix the phase and delete
+# the line.
+docs/roadmap/phase-230-platform-layer-split.md	R1
+docs/roadmap/phase-275-276-branch-notes.md	R4
+docs/roadmap/phase-292-asi-reference-consumer-revisit.md	R1
+docs/roadmap/phase-325-uorb-interop-and-bridge.md	R1
--- a/.config/roadmap-claims-baseline.txt
+++ b/.config/roadmap-claims-baseline.txt
@@ -2,5 +2,3 @@
 # body. RATCHET: this list may only SHRINK. Each line is a finding
 # for a person to work, not debt to hide; fix the phase and delete
 # the line.
-docs/roadmap/phase-230-platform-layer-split.md	R1
-docs/roadmap/phase-275-276-branch-notes.md	R4
--- a/.github/workflows/roadmap-audit.yml
+++ b/.github/workflows/roadmap-audit.yml
@@ -1,0 +1,52 @@
+# phase-419 W3 — the monthly roadmap verification pass.
+#
+# W2's `check-roadmap-claims` is a GATE and takes the mechanical half: a phase
+# that contradicts its own body. This is the other half, and it is deliberately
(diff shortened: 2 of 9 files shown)


jerry73204:
Pushed the two live findings this gate surfaced, so the baseline lands at 2 rather than 4.

**phase-292** — "Draft" (2026-07-16); W1/W3/W4 all landed **2026-07-17**, the day after that line was written. 17 of 18 boxes ticked, and the one that is not (W2.a, standing intake) is open-ended by construction. Three as-landed claims have since drifted and are corrected inline: `ws-realtime-cpp-fvp/` is gone (phase-331 W3 retired the themed `ws-*` dirs), the board crate is `nros-board-s32z270-freertos` not `…dc2-r52`, and `build-s32z-board-import` no longer exists.

**phase-325** — "Draft. Not started." with 14 of 18 ticked and W0/W1/W2 each carrying a RESULT dated **that same day**. The doc already contradicted itself five lines down in `**Blocked on:** … W0–W2 are done` — which is precisely the shape `header_block` had to be narrowed to see. **And its blocker is gone**: issue 0362 is `status: resolved` and archived, so W3 has been blocked on paper since July against a fixed issue.

Baseline **4 → 2**. The remaining two (230, 275) are corrected in #322; their lines delete when it lands, and the ratchet will say so.

PR #346, opened 2026-09-04
fix(phase-412): the boot-report knob rides the ninja command, not configure-time env

jerry73204:
Follow-up to #337, found by flashing it. Stacked on that branch, so this
reduces to a single commit once #337 merges.

**The image linked WITHOUT the record while every cmake variable said the
feature was on.**

`set(ENV{NROS_BOOT_REPORT})` runs at CONFIGURE time. The cargo that reads it is
spawned by a custom target at BUILD time, so the variable never reached it. The
file's own comment fifty lines further down says exactly this about the sizing
knobs, and names the seam they ride instead:

> the board FACTS + SITE config ride the same command, for the same reason the
> knobs do: Zephyr's cargo is spawned by this custom target, and `set(ENV{})`
> would only touch the configure-time process.

So the knob now rides `_nros_knob_env`, the ninja command prefix.

## Measured on the board

| | before | after |
| --- | --- | --- |
| `NROS_BOOT_REPORT=1` in build.ninja | 0 of 3 cargo commands | 3 of 3 |
| `cargo:rustc-cfg=nros_boot_report` | absent | emitted |
| `NROS_BOOT_REPORT` in the ELF | no symbol | `0x2042471c`, size 60 |

3 of 3 is the same coverage `NROS_EXECUTOR_MAX_CBS` has, and 60 is the size
`check-boot-report-layout` predicts from the struct.

## Why no gate caught it

This is the delivery-failure class phase-412 W4 exists for, arriving in the
work that was meant to instrument it. It was invisible for the usual reason:
the value was correct everywhere the gates look. `CONFIG_NROS_BOOT_REPORT=y`
was in `.config`, in `autoconf.h`, and in the cmake scope. Only the cargo
command line -- the last artifact before the compiler -- disagreed, which is
exactly why `check-knob-delivery.py` reads `build.ninja` and not cmake state.

It is not covered by that gate today because the gate enumerates sizing knobs
and this is a bool. Worth extending; not done here.


jerry73204:
Rebased onto main (15 commits, 2 dropped as already-upstream), and **dropped `4c8450c96` — `chore: prune ten doc-recipe-refs baseline entries that no longer offend`** — because its premise does not hold on current main.

The commit's reasoning was that `packages/cli/CONTRIBUTING.md` *"stopped referencing build-/clean-/test- recipe names that no longer exist, so their baseline waivers became dead"*. It still references them — nine occurrences on `origin/main` today, at `CONTRIBUTING.md:144,147,150,163,166,169` — and none of the recipes exist (`just --list` finds no `build-user-libs`, `build-build-tools`, `build-interfaces`, `list`, `run`). So the waivers are still load-bearing, and pruning them made the gate report those ten as *new* dead references:

```
check-doc-recipe-refs: 10 document(s) name a recipe that does not exist:
  packages/cli/CONTRIBUTING.md: just build-build-tools
  …
```

With the commit dropped: `check-doc-recipe-refs: OK — 49 known dead reference(s), no new ones.` and `just check fast` 218 gates, 0 failed.

The prune is still worth doing — but it needs the docs fixed first, which is the direction the gate's own message asks for (*"Name the recipe that exists, or drop the line"*). That is a separate change against `packages/cli/` docs, not a baseline edit, and it does not belong in a phase-412 branch.

PR #354, opened 2026-09-04
docs(#1046): a stale-tree guard that diagnoses the problem and cannot detect it

jerry73204:
`px4_uorb_interop_e2e.rs:65-75` asserts the module **directory** exists. Its failure message is a correct and complete description of the case it guards against:

> PX4 takes one `EXTERNAL_MODULES_LOCATION` per build and they share this build dir, so another root (e.g. `just px4 build-sitl-cpp`) built last.

The assertion cannot observe that case.

## Measured

On a tree whose last SITL build was `just px4 build-bridge-example` — exactly the "another root built last" situation:

| module | module dir | `bin/px4-<mod>` shim | occurrences in `bin/px4` |
| --- | --- | --- | ---: |
| `nros_uorb_demo` | present | present | **0** |
| `nros_uorb_bridge` | present | present | 8 |

**Both the module directories and the shims survive across builds.** Only the last root's modules are linked into `bin/px4`. So the guard passes, the test proceeds, and it dies at `nros_uorb_demo start` with a command-not-found — the confusing failure the guard exists to replace with a clear one.

**It checks an artifact that outlives the thing it proxies for.**

## The repo already names this class

Issue 0196: *"a probe that misses `generated/**` lets a museum binary pass every sweep."* Same shape, one input over — the probe watches a **directory** while the claim is about a **binary's contents**. Same distinguishing feature as 0445 and 0859–0862: a check that cannot tell *present* from *current*.

## Direction

Assert on binary **content** — a byte-scan for the command name, milliseconds on a 59 MB binary, no PX4 boot required. **Keep the message; it is right. Only the predicate is wrong.**

## Why filed, not fixed

Found while landing phase-325 W3.3 (#353), whose builds put the tree into the state that exposed it. It is a different test, in a different out-of-workspace crate — and **W3.4 will inherit this guard by copying it**, so whoever takes that needs the finding before the fix.

Three things explicitly listed as not covered, including whether the surviving `bin/px4-<mod>` shims are a second stale artifact worth their own look.

`just check fast` green.

### DIFF
--- a/docs/issues/1046-px4-stale-tree-guard-checks-a-surviving-directory.md
+++ b/docs/issues/1046-px4-stale-tree-guard-checks-a-surviving-directory.md
@@ -1,0 +1,97 @@
+---
+id: 1046
+title: "The PX4 SITL stale-tree guard asserts a module DIRECTORY, which outlives the build that linked it — so it passes on exactly the tree it exists to reject"
+status: open
+type: bug
+area: testing
+severity: medium
+found: 2026-09-04
+related: [phase-325, 0196, 0445, 0859]
+---
+
+# A guard that diagnoses the problem in prose and cannot detect it
+
+## The guard
+
+`packages/testing/nros-px4-sitl-test/tests/px4_uorb_interop_e2e.rs:65-75`:
+
+```rust
+let module_dir = build_dir.join("external_modules/modules/nros_uorb_demo");
+assert!(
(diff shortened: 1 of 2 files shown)


jerry73204:
Superseded by the PR above, which carries this issue plus the fix — verified in both directions. This one was queue-locked, so the fix went to a new branch rather than fighting the lock.

PR #356, opened 2026-09-04
docs(#0968): esp32 marker — every static hypothesis closed; the leaf needs its row's env to LINK

jerry73204:
Adding a `println!` beside the `log::info!` was the plan. It didn't survive contact — and what it turned up is worth more than the probe would have been.

## The probe was the wrong instrument

`Application setup complete` prints only when the run-plan closure returns `Ok(())`, and the `log::info!` sits **inside** `register()` before that `Ok`. Execution demonstrably reaches and passes the line. The question was never *"is it reached"* — which is all a `println!` beside it answers — but *"why is the record dropped"*.

## Five hypotheses, all closed

| hypothesis | verdict |
| --- | --- |
| execution never reaches the line | closed — proved by `Application setup complete` |
| no logger installed | closed — `init_logger(Info)` in `init_hardware`, which ran |
| compile-time level stripping | closed — no `max_level_*` feature |
| two `log` facades | closed — **my grep artifact** (`log v[0-9.]+` matched `nros-log v0.5.0`) |
| esp-println without `log-04` | closed — resolved features include it |

One facade, logger installed on it at `Info`, macros compiled in, line reached, and no record on a console the same image's `println!` reaches. **Nothing further is answerable from the dependency graph** — this is the fourth wrong answer it's produced, recorded so the next person stops asking it there.

## A separate fact, and it reinforces #1025

**The listener does not LINK without its manifest row's env.**

```
rust-lld: error: .../stack.x:11: unable to move location counter (0x3fccf734)
          backward to 0x3fcce400 for section '.stack'
```

With `ZPICO_MAX_QUERYABLES=2` — what `env` on its `[[fixture]]` row supplies — it links. The image sits close enough to its RAM ceiling that the row's env is load-bearing for the **link**, not merely for size.

That's the same env whose absence from the packer's group-key computation *was* issue 1025 — and it's why probing this image by adding code is awkward: two string literals and one dependency pushed it back over.

## Next attempt

Probe without growing the image: write through the platform writer the board already registers (`nros-platform-esp32-qemu/src/lib.rs:87`, via `PlatformLog::write`) instead of adding a print beside the marker.

Docs only.

### DIFF
--- a/docs/issues/0968-tier2-runtime-failures-unreproduced.md
+++ b/docs/issues/0968-tier2-runtime-failures-unreproduced.md
@@ -374,3 +374,45 @@ through a path the image already links (the board registers a platform writer at
 `PlatformLog::write`), rather than adding a print beside it. That needs a dep
 the example does not currently have, so it is a temporary edit to the leaf's
 manifest — build it with the row's env, and revert both.
+
+
+## The esp32 marker, ANSWERED by probe (2026-09-04) — filed as issue 1048
+
+The probe ran, through the path the board already exports, and the answer is
+unambiguous. A `println!` on EACH side of the `log::info!`, image built with its
+row's env, QEMU with a zenoh router up:
+
+```
+PROBE-A
+PROBE-B
+
+Application setup complete — entering spin loop.
+```
+
+**Both probes print; the `log::info!` between them prints nothing.** No
+`log`-crate output appears anywhere in the boot. The subscription exists and the
+line announcing it is dropped by the facade — so the cell greps for a marker the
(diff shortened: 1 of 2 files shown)


jerry73204:
Superseded — a rebase onto current `main` drops both commits as already-applied upstream, and `main`'s `docs/issues/0968-*.md` and `1048-*.md` are supersets of this branch's versions (0968 is byte-identical; 1048 has 46 further lines this branch does not carry, including the FIXED section). Closing rather than rebasing an empty branch.

PR #360, opened 2026-09-04
docs(#1048): root cause — log::set_logger does not exist on riscv32imc

jerry73204:
**Stacked on #356.**

Not a race, not an ordering bug, not a filter. In `log` 0.4.33:

```rust
#[cfg(target_has_atomic = "ptr")]
pub fn set_logger(logger: &'static dyn Log) -> Result<(), SetLoggerError> { … }

#[cfg(target_has_atomic = "ptr")]
pub fn set_max_level(level: LevelFilter) { … }
```

ESP32-C3 is `riscv32imc-unknown-none-elf` — RV32 **I M C**, no `A` extension, no atomic pointer ops. `rustc --print cfg` for it lists `target_has_atomic_primitive_alignment="ptr"` and **no `target_has_atomic="ptr"`**, so both are compiled out.

**Proved, not inferred** — a probe installing its own logger from the board crate doesn't compile:

```
error[E0425]: cannot find function `set_logger` in crate `log`
```

So **no logger can be installed on this board by anyone**, and every record is dropped. That explains all of it: no INFO line anywhere in the boot, the marker never printing, and `println!`s bracketing it both printing.

## Three readings retired, two of them mine

- `init_logger(LevelFilter::Info)` can't do anything here. Its issue-#64 comment — *"without it the log crate has no logger installed and silently drops every record"* — describes the state the board is **still in**.
- **`log::max_level() = Info` is not evidence a logger installed.** The getter is unconditional; the setter isn't. I read it as evidence and was wrong.
- `ESP_LOG=info` at build time changes nothing. Tested.

## The fix is a choice

1. **Platform writer** — the board already registers one (`register_log_writer`): a plain fn pointer, no atomics, works today, and the same shape CLAUDE.md prescribes for Zephyr instead of `std` stdio.
2. **Give `log` atomics** — `portable-atomic` is already in the graph, but `log` gates on `target_has_atomic` directly with no such feature, so this means patching `log`.

## Scope

Not only test markers: **every** `log::info!`/`warn!`/`error!` in any crate this board links is discarded, nros framework diagnostics included. An image that fails here can only report through `println!` or the platform writer. Any `no_std` target lacking `target_has_atomic = "ptr"` has the same property.

Docs only; all probes reverted.

### DIFF
--- a/docs/issues/0968-tier2-runtime-failures-unreproduced.md
+++ b/docs/issues/0968-tier2-runtime-failures-unreproduced.md
@@ -374,3 +374,45 @@ through a path the image already links (the board registers a platform writer at
 `PlatformLog::write`), rather than adding a print beside it. That needs a dep
 the example does not currently have, so it is a temporary edit to the leaf's
 manifest — build it with the row's env, and revert both.
+
+
+## The esp32 marker, ANSWERED by probe (2026-09-04) — filed as issue 1048
+
+The probe ran, through the path the board already exports, and the answer is
+unambiguous. A `println!` on EACH side of the `log::info!`, image built with its
+row's env, QEMU with a zenoh router up:
+
+```
+PROBE-A
+PROBE-B
+
+Application setup complete — entering spin loop.
+```
+
+**Both probes print; the `log::info!` between them prints nothing.** No
+`log`-crate output appears anywhere in the boot. The subscription exists and the
+line announcing it is dropped by the facade — so the cell greps for a marker the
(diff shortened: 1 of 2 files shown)


jerry73204:
Superseded — a rebase onto current `main` drops all three commits as already-applied upstream. `main`'s `docs/issues/1048-esp32-log-records-are-silently-dropped.md` contains this branch's content plus 46 more lines (the `FIXED 2026-09-04 — and it took TWO changes, not one` section). Closing rather than rebasing an empty branch.

PR #401, opened 2026-09-04
docs(phase-422): provisioning has four paths that disagree — one campaign to fix it

jerry73204:
Opens the campaign to tidy the provisioning system, grounded in what phase-413 actually turned up rather than written from intent.

## What I found

**1. Two producers for the same tool.** `[tool.corrosion]` and `just workspace install-corrosion` both install corrosion into the same versioned prefix, each with its own stamp logic. Same for `play_launch_parser`. Their pins are not even spelled alike:

| tool | `just/workspace.just` | `nros-sdk-index.toml` |
| --- | --- | --- |
| corrosion | `v0.6.1` | `0.6.1-nros1` |
| play_launch_parser | git SHA `838ce948…` | `0.1.0-nros1` |

Issue 0500 already paid for this once: the store accumulates, prefixes resolve newest-first, and **both paths print success either way** — so a stale Corrosion shadowed the pin just installed and `mixed` could not link.

I checked before asserting: `install-sccache` is **not** in this category — it forwards to `nros setup --tool sccache`. The doc says so, and marks it tidying rather than a correctness fix.

**2. Installers that never reached the index.** `clang-format` (pip wheel), `mdbook` (script + hand-maintained checksum file), `verus` (script). Each re-implements pin/download/checksum/prefix/smoke that the index does for 16 other tools, and none can answer `--check`.

**3. Two spellings of the setup verb**, one of which silently does less — `just <scope> setup` skips `_setup-common`.

**4. CI lanes that don't walk the user's path** — the nightly zephyr jobs, blocked on the dispatcher being unable to forward `--skip-sdk`.

## The measured cost

Three consecutive ~40-minute host-tests runs, each failing on a different missing prerequisite, each only visible once the one above was fixed.

## Shape

W1–W3 mechanical and independent. W4 blocked on a dispatcher change in W3. W5 is the ratchet and names the one gate still missing — *no indexed tool may have a second installer* — alongside the two that already landed (#387, #397).

**Two decisions deliberately left open** rather than guessed: whether the index grows a wheel source kind for clang-format or clang-format stays a documented exception; and which spelling survives W1, given `_setup-common` calls `install-corrosion` today.

## Verification

- `just check roadmap-status` — OK, 61 active phases carry a status line
- `just check fast` — 201 gates, 0 failures

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -330,10 +330,15 @@ jobs:
           source ./activate.sh
           nros setup --tool esp32-qemu --sudo
 
-      - name: just ${{ matrix.plat }} setup
+      # The DISPATCHER spelling, not `just <plat> setup`. Only
+      # `just setup <scope>` runs `_setup-common`, which provisions the host
+      # facts every tier asserts (cross Rust targets, pinned corrosion, the CLI,
+      # the resolver, clang-format). The module spelling silently skips all of
+      # it, which is why this job needed a hand-rolled cross-target step below.
+      - name: just setup ${{ matrix.plat }}
         run: |
           source ./activate.sh
-          just ${{ matrix.plat }} setup
+          just setup ${{ matrix.plat }}
 
       # issue-0037 — the e2e fixture build shells out to `play_launch_parser`, a
       # separate binary the in-tree `cargo build --bin nros` does NOT produce.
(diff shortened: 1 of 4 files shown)


jerry73204:
Superseded — both commits are on `main` under the same subjects (`f06766f03` and `d71a3be4e`), so a rebase onto current `main` produces zero commits: every patch is already applied upstream. Closing rather than rebasing an empty branch.

PR #410, opened 2026-09-04
docs(#1052): the fault is reached from ZenohSession::create_publisher — and my bisect was measuring a stale image

jerry73204:
Two commits. The first records what gdb found; the second runs the check that first commit asked for, and it corrects me.

## gdb names the frame

`riscv32-esp-elf-gdb` against QEMU's gdbstub, router **confirmed serving on :9800 first**. Break at the handler, resolve every code address on the stack:

```
0x42017ec8  <ZenohSession as Session>::create_publisher
0x420531ca  esp_hal::interrupt::riscv::vectored::enable_on_cpu
0x4205334c  <riscv_rt::TrapFrame as Debug>::fmt
0x4205e950  <i32 as Display>::fmt
```

Everything but the first is the exception handler printing itself. Two readings I deliberately did *not* take: `ra` → `assert_failed_inner+2` is a **mid-symbol** offset (nearest-preceding resolution, not a panic frame), and `a0`/`a2` are the handler's own format literals, not an assert message. I over-read registers once already in this issue.

## The check that settles it

Empty-`register` talker, **artifact verified this time** — `create_publisher` symbol count in the fresh ELF is **0**. (`Hello World` is a bad probe and survives; it's in the callback body. The symbol count discriminates.) Same router, same harness QEMU args, original image kept as a matched control in the same session:

| image | `create_publisher` in ELF | faults in 60 s |
|---|---|---|
| empty `register` | 0 | **0** |
| full talker (control) | present | **2** |

`register` decides it. The frame is real. **Variants E, F and H are retracted** — they measured the original image.

## Why they were stale: this issue's own sibling, #1025

`just esp32 build-qemu` resolved the ELF with

```sh
nros_fixture_row_artifact_dir "<leaf>" qemu-esp32-baremetal "" ""
```

— empty `cargo_args`/`envstr`, **invented at the call site** instead of read from the row. The talker row carries the variant `ZPICO_MAX_QUERYABLES=2`, so cargo writes to `qemu-esp32-baremetal-4118800323` while the packer looks in bare `qemu-esp32-baremetal`. A stale ELF sat at the bare path, so every "rebuild" packed the same binary and three variants agreed because **they were one image**.

That's #1025 exactly. Its fix — the by-id helper, which reads the row rather than inventing its inputs — is in open PR #303, still unmerged, which is why the trap was armed today. It now fails loud rather than packing stale, but only because #700 landed in between.

## Established

The fault is reached from `ZenohSession::create_publisher`, on the connected path, talker only. The listener creates a subscription and doesn't fault. Next step is that function, not another bisect.

## Method note

Every wrong turn in this issue has one shape: a measurement whose artifact provenance was never checked — the ASCII lead, the "fault is on the connected path" claim, now E/F/H. The control costs one command (`nm | grep -c <symbol>` on the ELF you're about to run) and would have caught all three.

Docs only.

### DIFF
--- a/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
+++ b/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
@@ -282,6 +282,53 @@ resolving `0x42051d70` cannot name the caller — it is a frame of the panic
 printer, not of the faulting code. Step 1 is a dead end by construction; steps 2
 and 3 are unaffected.
 
+## gdb names the frame (2026-09-04): `ZenohSession::create_publisher`
+
+`riscv32-esp-elf-gdb` against QEMU's gdbstub, with the router CONFIRMED serving
+on :9800 first. Break at the handler, dump the stack, resolve every code address
+on it:
+
+```
+0x42017ec8  <nros_rmw_zenoh::shim::session::ZenohSession as nros_rmw::traits::Session>::create_publisher
+0x420531ca  esp_hal::interrupt::riscv::vectored::enable_on_cpu
+0x4205334c  <riscv_rt::TrapFrame as core::fmt::Debug>::fmt
+0x4205e950  <i32 as core::fmt::Display>::fmt
+```
+
+Everything but the first is the exception handler printing itself. **The only
+application frame on the faulting stack is `ZenohSession::create_publisher`.**
+
+At the handler: `sp` is in `.bss` near `nros_smoltcp::TCP_RX_BUFFER_0`, and `ra`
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded — both sections are on `main` already, landed via #426, and byte-identical: rebasing this branch onto current main drops **every** commit as already-applied and leaves 0. Verified the two headings (`## gdb names the frame`, `## RESOLVED, both halves`) each appear once in `docs/issues/1052-*.md` on main. Closing rather than force-pushing an empty branch.

PR #419, opened 2026-09-04
fix(#1006): esp32-qemu's link surface is the recipe's now, not the build host's

jerry73204:
Phase-423. `[tool.esp32-qemu]` is source-built with **no dist**, so every user runs
its `configure` on their own machine — and qemu's configure auto-detects: a
feature whose `--enable-`/`--disable-` is not passed stays `auto`, so
`dependency(..., required: get_option(<f>))` resolves against whatever dev headers
the build host carries. **55 host-library features were unpinned**, which is the
69 undeclared sonames #1006 reported.

The link surface is now the recipe's, not the host's: **12 flags → 69**, in six
groups (UI, audio, host-device passthrough, block drivers beyond the raw image,
TLS, remaining host libs). `--enable-gcrypt`, `--enable-slirp`,
`--target-list=riscv32-softmmu` and `--disable-werror` are untouched; `fdt`,
`pixman` and `zlib` are deliberately left auto (riscv boards require fdt; zlib is
`required: true`).

## Verified without a build

**All 65 disable flags are accepted at the pinned tag.** This is the load-bearing
safety property — an unknown option is `ERROR: unknown option`, a hard failure,
and it is exactly the risk the issue cited for *not* doing this. Checked twice,
independently: by the agent against the fork's generated
`scripts/meson-buildoptions.sh` and `configure`, and again by me fetching both
files from `espressif/qemu@esp-develop-9.2.2-20260417` and matching all 65. Zero
unknown.

**The issue's own prescription would not have worked.** It suggested
`--audio-drv-list=`; `audio/meson.build` builds a module for every driver whose
dependency `.found()`, *not* for the ones in that list, so the empty list only
changes `CONFIG_AUDIO_DRIVERS`. The per-driver `--disable-alsa/pa/jack/oss/sndio/
pipewire` is what cuts the link — and with those off the default list is empty
anyway, so the prescribed flag is redundant and is not in the new line.

**Three of the five "safe on any build" entries are not.** Only `zlib` is
`required: true`; `pcre2` is glib's transitive. `selinux` and curses→`libtinfo`
are ordinary auto features, and the fork's `meson.build` references **no openssl
at all** — `libssl3` arrived via libcurl/libssh. They stay declared (the gate
checks coverage, not minimality) with the comment corrected; deleting them without
a re-measure would be a guess.

**Why the two recipes differ is not the configure.** `[tool.qemu.source]` is
*equally* unpinned; its `system = 2` comes from its dist bundling its closure. The
real difference is that nobody but the release job runs qemu's configure, while
every esp32 user runs this one. The issue's control table is misleading on this
and the doc now says so.

## New gate, with a negative control

`check-qemu-source-features.py`, wired into `just check sdk-index` and gate.yml: a
`[tool.*]` whose source is qemu, whose configure runs `./configure`, and which
publishes no dist must pin every host-library feature by name. Green after the
fix; **exit 1 with 55 of 58 unpinned against the pre-fix revision.**

Its narrowing is in the docstring rather than hidden: a dist-carrying recipe still
source-builds on a host key with no dist, and tightening that means editing
`build-qemu.sh`, which needs a release build to verify.

## Unverified — and why the issue stays open

The resulting `ldd` closure. Only `nros setup --tool esp32-qemu` +
`check-dist-runtime-deps` answers that, esp32-qemu is not provisioned on this host,
and CI never provisions it — which is also why the existing gate could not have
caught this. Which soname each disable drops is inferred from Debian's link graph.
That the full 69-flag configure *completes* is argued (every `error_message:` line
checked) but not run.

`status: open` deliberately: the issue's own acceptance needs a rebuild.

### DIFF
--- a/.github/workflows/gate.yml
+++ b/.github/workflows/gate.yml
@@ -979,6 +979,10 @@ jobs:
       # must not drift (else prebuilt != source-built).
       - name: Check qemu configure flags match (index vs build script)
         run: ./scripts/sdk/check-qemu-configure.sh
+      # issue 1006 — a qemu recipe with no `dist` is source-built by every user,
+      # so an auto-detected backend makes its runtime deps the build host's.
+      - name: Check source-only qemu recipes pin their backends
+        run: python3 scripts/sdk/check-qemu-source-features.py nros-sdk-index.toml
 
   # ---------------------------------------------------------------------------
   # THE REQUIRED CHECK — phase-395 W20.
--- a/docs/issues/1006-esp32-qemu-configure-nondeterministic-deps.md
+++ b/docs/issues/1006-esp32-qemu-configure-nondeterministic-deps.md
@@ -70,3 +70,64 @@ observable on a host that has actually built this tool. That is why the gap sat
 unnoticed: CI never provisions esp32-qemu (it is nightly-only), and a developer
 who has built it sees a 75-line failure that looks like index rot rather than a
 configure problem.
+
+## What landed (the configure is pinned; the closure is NOT re-measured)
+
+`[tool.esp32-qemu.source].configure` now names 69 flags instead of 13. The added
+ones are all `--disable-*`, grouped as UI, audio, host-device passthrough,
+block drivers beyond the raw image, TLS, and the remaining host libraries — the
+whole set of qemu feature options that reach for a system library and that
+`-M esp32c3 -icount 3 -nographic -drive file=…,if=mtd,format=raw -nic
+user,model=open_eth` never touches. `--enable-gcrypt`, `--enable-slirp`,
+`--target-list=riscv32-softmmu` and `--disable-werror` are untouched; `fdt`,
+`pixman` and `zlib` are left auto deliberately (the riscv boards require fdt,
+and zlib is `required: true`).
+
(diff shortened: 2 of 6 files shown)


jerry73204:
Verification loose end from the PR body is now closed: **`just check sdk-index` passes in full, exit 0**, all three steps on the committed tree.

- `verify-index.py` — index coherent; 28 prebuilt assets downloaded and sha256-verified (the long pole, and nothing in this diff touches a dist).
- `check-qemu-configure.sh` — the qemu index ↔ `build-qemu.sh` flags still match, untouched as intended.
- `check-qemu-source-features.py` (new) — `esp32-qemu: source-only — pinned; qemu: has dist — exempt`.

So the branch's verification state is: new gate green on the fix and **red at 55 of 58 unpinned (exit 1) on the pre-fix revision**; all 69 configure flag tokens proven acceptable at `esp-develop-9.2.2-20260417` against the fork's own generated option parser — checked twice, independently.

What stays unverified is unchanged and is why #1006 remains `status: open`: the actual ldd closure after a rebuild, and the per-soname attribution of each disabled feature (inferred from Debian's link graph, not measured).

PR #422, opened 2026-09-04
fix(#1025): the group FORMULA was single; its INPUTS were still derived twice

jerry73204:
Phase-423, step 1: **no ESP32 QEMU flash image could be produced at all**, so every
test that consumes one failed for a reason unrelated to the code under test.

## The mechanism, and it is not the one the issue's title suggests

`nros_fixture_group_slug` keys the shared cargo target dir on a row's **variant** —
a cksum over (cargo args, sorted env). phase-340 item 7 already made the group
**formula** single. But the key is a function of *three* inputs, and the packer
supplied only one of them from where the producer got it:

```sh
nros_fixture_row_artifact_dir "examples/…/$ex" qemu-esp32-baremetal "" ""
```

The producer passed the row's real args/env. The two agreed until `41a7d8de7`
added `env = { ZPICO_MAX_QUERYABLES = "2" }` to the three esp32 rows. **Formula
single, inputs derived twice** — which is why a single-formula invariant did not
prevent it.

The fix is `nros_fixture_row_artifact_dir_by_id <row-id>`: read (platform, dir,
env, args) from the same manifest record `fixtures-build.sh` builds from, so one
row id yields one key. `--builder cargo` is load-bearing — a cmake row has a
different record shape, so a cmake id becomes a loud failure rather than a 4-field
misparse.

## Measured

**Before:** `just esp32 build-qemu` rc=1, `ERROR: …/qemu-esp32-baremetal/…/
esp32_qemu_talker is missing`, while `build/cargo-fixtures/` held exactly one
entry — `qemu-esp32-baremetal-4118800323` — containing both ELFs.

**After, re-run by me on the agent's branch with cargo untouched:** rc=0, two
4 MiB images. `build-logging-smoke` likewise.

**Sweep, 8 call sites.** Six pair the read with a `target_dir_flag` on the same key
in the same recipe — they build what they read. The two unpaired esp32 sites were
both wrong. `qemu-baremetal.just:337` is unpaired but names a one-group platform,
so it is correct *by luck*; now gated.

**Test-side locator checked by measurement, not by reading:**
`build_esp32_qemu_example` hand-spells `<leaf>/target/…`, which nothing writes, but
`require_prebuilt_binary` redirects through `fixtures::groups::resolved`. A
temporary integration test resolved all three authored leaf paths onto the hashed
dir, `exists=true`, then was deleted.

## The gate, and its own control

`check-fixture-artifact-dir-keys` (fast lane) mutation-checked against the **real**
defect, both rules: restoring the two pre-fix call sites names both with the right
slugs.

Worth recording: the first version of its synthetic negative control was **wrong** —
it was driven off a real variant row, and `examples/native/rust/talker` has a plain
`linux` row beside eight variants, so `linux "" ""` there is correct and the control
disarmed itself. It now uses a synthetic row table.

`just check fast`: **203/203**, re-run by me.

## Not fixed, deliberately

`just nuttx _run-qemu` builds *and* reads with `nuttx "" ""` — self-consistent, so
nothing is broken — but every `qemu-arm-nuttx/rust/*` row carries an
`NROS_LOCATOR`, so it compiles a **third** population into a group the fixture lane
never uses. Fixing it needs the row's env delivered to the build (the locator is
baked at compile time), i.e. a behaviour change on an interactive path that could
not be tested here. Recorded in the archived issue and the gate docstring.

## Could not verify

**Bearing on #0968 is not established.** All five esp32 tests skip here — the host
`qemu-system-riscv32` is stock and has no `esp32c3` machine. This removes a certain
blocker without proving it was theirs.


jerry73204:
Superseded by #303, which fixed the same defect and merged. Closing rather than rebasing, because keeping both would leave **two gates for one rule** — `check-fixture-artifact-dir-keys` here and `check-fixture-artifact-dir-inputs` on main — which is the drift this repo already warns about.

Nothing is lost silently. This PR's unique content was its CLAUDE.md entry and book paragraph, and both are salvaged in **#461**, with two edits:

- the gate name corrected to the one that actually merged;
- one claim **fixed rather than carried across**. This PR says no flash image could be packed "for four days with every gate green". That is false, and CLAUDE.md is loaded every session, so it would have taught the wrong lesson permanently. The nightly esp32 lane went red and printed the exact failing path every night. The accurate and more useful version is now in #461: the *push* gates are green either way, because the packer runs in `build-all` and only the schedule reaches it — and a lane already red reports its FIRST failure, which is why #1070 sat behind this one, invisible until it was fixed.

The independent diagnosis here was correct and arrived at the same helper. Only one implementation can ship.

jerry73204:
Not rebased — this needs re-authoring rather than a conflict resolution.

`main` already carries a fix for #1025 (`fixture-artifact-dir-inputs`, and a two-argument `nros_fixture_row_artifact_dir_by_id "<row-id>" <platform>`). This branch carries a different one (`fixture-artifact-dir-keys`, and a one-argument form that derives the platform from the manifest via `--with-platform`). Both fix the same defect; they conflict in `just/check.just`, `just/esp32.just`, `scripts/build/fixtures-target-dir.sh` and the archived issue file.

The branch's argument — *the group FORMULA was single; its INPUTS were still derived twice* — is the stronger position, since passing the platform alongside the row id is exactly the second derivation it names. But adopting it means changing the signature that landed and every caller of it, which is a decision for you, not something a rebase should pick.

Rebasing the other 21 conflicting PRs found the same shared-file class throughout — see #468 / issue 1072.

PR #426, opened 2026-09-04
fix(#1052): ESP32 talker faults from a stack overflow — root cause, fix, and the floor gate

jerry73204:
`create_publisher` is not the bug. It's the deepest frame on the deepest chain, so it's where an already-too-small stack runs out.

> Stacks on commits also present in queued PR #410 (the gdb finding it builds on). Once #410 merges this reduces to the last two commits.

## Root cause

| symbol | value |
|---|---|
| `.bss` | `0x3fc81960` + `0x048214` = **295,444 B** |
| `_stack_end` (low bound) | `0x3fcc9b74` |
| `_stack_start` (high; grows down) | `0x3fcce400` |
| **stack size** | **18,572 B** |
| **faulting `sp`** | **`0x3fcc9180`** |

`sp` is **2,548 bytes below `_stack_end`** — outside the stack, inside `.bss`, in `nros_smoltcp::TCP_RX_BUFFER_0`+1892, exactly as gdb symbolised it. Stack overflow; the instruction-access fault is the smashed frame returned through.

`nros-board-esp32-qemu/src/node.rs` already documents this as issue **#190**:

> `.stack` is the LINKER LEFTOVER after `.bss` … At 96 KB the stack shrank to ~18 KB … the overflow wrote frames straight down into `.bss` … wild jumps … **None of them were allocator or zenoh-pico bugs.**
> 48 KB … leaves a **~67 KB stack** … Check `.stack` in `readelf -S` after changing ANY large static — **there is no runtime stack-overflow guard on this target.**

Heap knob still 48 KB; stack is **18,572 B, not ~67 KB**. `.bss` grew and nobody re-ran that check. **1052 is a regression of #190.**

## Confirmed both directions

| | `.bss` | stack | `SMALL_PAYLOADS` |
|---|---:|---:|---:|
| shipped | 295,444 | **18,572** | 32,768 |
| `ZPICO_MAX_SUBSCRIBERS=1` | 264,868 | **49,148** | 4,096 |

| image | stack | output | `ConnectionFailed` | faults |
|---|---:|---:|---:|---:|
| full talker | 18,572 | **70 — died** | 0 | **2** |
| empty `register` | 18,572 | 23,083 | 23,033 | 0 |
| `SUBSCRIBERS=1` | 49,148 | 23,850 | 23,565 | **0** |

The shipped talker dies at line 70, *before* the retry loop the others run 23k lines of — same early path, only the 18 KB build faults.

## Corrects 91241878b

That commit recorded `mepc=0x732f7264` across three builds and argued constancy rules out stack smashing, "which would move with layout". Today's run faults at **`mepc=0x02250042`** — not constant, so the argument doesn't hold.

They never conflicted. It concluded the value is "DATA the code reads deterministically from a fixed place", `0x732f7264` being printable `"s/rd"`. **The fixed place now has a name:** the overflowed stack sits inside `TCP_RX_BUFFER_0`, so a return-address slot is overwritten by received network bytes — keyexpr traffic is where `"s/rd"` comes from. Constancy was evidence *for* the mechanism, read as evidence against it.

## Two campaign findings in `.bss`

- **Two heaps coexist.** node.rs budgets one (48 KB `esp_alloc`) and says DDS swaps in a `FreeListHeap` *instead* — but `memory.rs:20` declares `FreeListHeap<{32*1024}>` on the **non-DDS** arm too. Both link: **83,632 B** of heap.
- **The talker pays 65,536 B for subscriber pools it never uses.** Sized on `ZPICO_MAX_SUBSCRIBERS`, not on what the node declares — 0827/0832's thesis exactly. On this target `.bss` is subtracted from the stack, so it isn't a footprint nicety, it's the crash.

## Fixes, in order

1. **Gate a `.stack` floor.** The comment asks for a check a comment can't enforce, and there's no runtime guard. A silent 18 KB stack is how this got here.
2. **Stop allocating pools for undeclared entities** — up to 64 KB back.
3. **Resolve the two heaps** — 83,632 B where the comment budgets 49,152.

This PR does (2) for the talker row and documents the rest. (1) is the one that keeps it fixed.

## Also corrects my own previous commit

Its "`create_publisher` symbol count is 0" came from `riscv32-esp-elf-nm … 2>/dev/null | grep -c` — that tool isn't on PATH, the redirect ate "command not found", and `grep -c` returned 0 for *every* input. Run properly, the full talker has 5. The behavioural conclusion is unaffected (distinct checksums, back-to-back identical args, and a build log carrying `unused imports: NodeOptions and TimerDuration`, which only appears once `register` is emptied).

### DIFF
--- a/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
+++ b/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
@@ -857,3 +857,41 @@ Cause established and demonstrated both directions. Remaining work is the fix
 ladder above — and item 1, the `.stack` floor gate, is the one that matters: this
 image shipped with an 18 KB stack against a comment budgeting ~67 KB, and nothing
 said a word.
+
+
+## Both images now clear the floor on what they DECLARE (2026-09-04)
+
+The listener was carried as recorded debt when the gate landed. It is fixed, and
+the table is empty:
+
+| image | declares | before | after |
+| --- | --- | ---: | ---: |
+| `esp32_qemu_talker` | 1 pub, 1 timer | 18,572 B | **49,148 B** |
+| `esp32_qemu_listener` | 1 sub | 22,060 B | **52,636 B** |
+
+Both against a 32,768 B floor; `DEBT` in `check-stack-floor.py` is now empty.
+
+The listener declared exactly one subscription and paid a budget of 8
+(`SMALL_PAYLOADS`, 32 KB), plus `SERVICE_BUFFERS` for services it never declares.
+Setting `ZPICO_MAX_SUBSCRIBERS = "1"` returns 30,576 B of stack. Verified at
+runtime, not only at link: it boots, logs `Subscriber created for topic:
+/chatter`, and takes no fault — a budget of 1 is enough for a leaf that declares
+1.
+
+The gate proved itself on the way through. Raising the listener tripped
+`EXCEEDS its recorded debt of 22,060 B` and refused the build until the entry was
+deleted, which is what a ceiling-that-may-only-fall is for: an improvement is not
+allowed to sit quietly in a table that says the image is broken.
+
+### The hand-maintenance is now issue 1061
+
+Three rows in `fixtures.toml` now restate, by hand and in another file, what
+`register` already says. `entity_inventory.rs` computes exactly these numbers but
+publishes them only as CMake variables, and these leaves are pure-cargo — so they
+take `zpico-sys`'s compiled-in defaults no matter what they declare. Filed as
+**1061**; the queryable half of the same table was already applied "a row too
+narrowly the first time", which is that failure mode having happened once.
+
+`check-stack-floor.py` bounds the damage in the meantime: a leaf that gains an
+entity without gaining budget fails at build time instead of becoming a wild jump
+at runtime.
--- a/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
+++ b/docs/issues/1052-esp32-talker-faults-after-network-bringup.md
@@ -568,3 +568,52 @@ rather than a line of code.
   from that capture is recorded here. Two instruction-access faults are not
   necessarily the same fault, and the pre-fix build had no `nros_log` record
   path in the leaves.
+
+
+
+## gdb names the frame (2026-09-04): `ZenohSession::create_publisher`
+
+`riscv32-esp-elf-gdb` against QEMU's gdbstub, with the router CONFIRMED serving
+on :9800 first. Break at the handler, dump the stack, resolve every code address
+on it:
+
+```
+0x42017ec8  <nros_rmw_zenoh::shim::session::ZenohSession as nros_rmw::traits::Session>::create_publisher
+0x420531ca  esp_hal::interrupt::riscv::vectored::enable_on_cpu
+0x4205334c  <riscv_rt::TrapFrame as core::fmt::Debug>::fmt
+0x4205e950  <i32 as core::fmt::Display>::fmt
+```
+
+Everything but the first is the exception handler printing itself. **The only
+application frame on the faulting stack is `ZenohSession::create_publisher`.**
+
+At the handler: `sp` is in `.bss` near `nros_smoltcp::TCP_RX_BUFFER_0`, and `ra`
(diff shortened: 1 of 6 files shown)


jerry73204:
## Added: the `.stack` floor gate

`scripts/check-stack-floor.py`, floor **32 KB** for `qemu-esp32-baremetal`, run per row by `fixtures-build.sh` right after that row's `cargo build`.

**Why there and not in a `check` lane.** That's the only place the artifact dir is known correctly — `$tdir_flag` comes from the row's own args/envstr, so it names where cargo actually wrote. A `check` lane would need a prebuilt fixture (banned in an affordability tier) or would skip when the image is absent, and a skip is the silence this gate exists to end. Only `--selftest`/`--claims` run on the fast line.

**The floor is not a ratchet.** 18,572 crashes, 22,060 doesn't — but 22,060 isn't safety, the listener just runs a shallower chain. The number to respect is what node.rs and #64 measured for the handshake + nested smoltcp poll path. A floor set above the last crash blesses the next one. The listener goes in `DEBT` with issue 1052: the recorded value is a **ceiling that may only fall**, so the gate fails if it rises, and deleting the entry is what closing the issue looks like.

**Verified on the real build, both directions:**

| | result |
|---|---|
| without the pool fix | ❌ fails on talker at 18,572 B |
| with it | ✅ talker 49,148 B, listener passes on recorded debt |

(`build-qemu` still exits non-zero afterwards on this tree for an unrelated reason — the packer's #1025 lookup reads `build/cargo-fixtures/qemu-esp32-baremetal`, which doesn't exist. **#303 fixes that and is independent.** Worth noting it means `just esp32 build-qemu` cannot complete on `main` today at all.)

**Two bugs this cost me, both the gate's own class:**

- the first artifact filter excluded **everything** — `*/build/*` matched the *absolute* path, since the repo's own `build/cargo-fixtures/` prefix is in it. "No executable found" reads exactly like "this row built nothing". Caught only because a control expected a *failing* image and got the wrong failure. Patterns are anchored to the relative path now.
- globbing a phase-340 group dir checks **other rows'** artifacts — they're shared, and keep binaries from earlier env combos no row produces any more. It failed the listener's build over a stale talker ELF from before the fix under test. `--row` resolves the leaf's own binary names from its Cargo.toml instead.

`find_nm()` raises rather than defaulting — during this issue a probe spelled `riscv32-esp-elf-nm … 2>/dev/null | grep -c` reported a symbol absent when the tool merely wasn't on PATH. The selftest also runs on the normal path so the controls can't rot.

`just check fast`: 202/202 green.

jerry73204:
## Listener fixed too — `DEBT` is now empty

| image | declares | before | after |
|---|---|---:|---:|
| `esp32_qemu_talker` | 1 pub, 1 timer | 18,572 B | **49,148 B** |
| `esp32_qemu_listener` | 1 sub | 22,060 B | **52,636 B** |

Both against the 32,768 B floor. The listener declared exactly one subscription and paid a budget of 8 (`SMALL_PAYLOADS`, 32 KB) plus `SERVICE_BUFFERS` for services it never declares; `ZPICO_MAX_SUBSCRIBERS=1` returns 30,576 B of stack.

**Verified at runtime, not only at link** — it boots, logs `Subscriber created for topic: /chatter`, takes no fault. A budget of 1 suffices for a leaf that declares 1.

**The gate proved itself on the way through.** Raising the listener tripped `EXCEEDS its recorded debt of 22,060 B` and refused the build until the entry was deleted. That's what a ceiling-that-may-only-fall is for: an improvement doesn't get to sit quietly in a table still asserting the image is broken. The `DEBT` mechanism stays under test with a synthetic entry rather than lapsing along with the table.

## Filed the class: #1061

Three rows in `fixtures.toml` now restate by hand, in another file, what `register` already says. `entity_inventory.rs` computes exactly these numbers — but publishes them only as **CMake variables**, and these leaves are pure-cargo, so they take `zpico-sys`'s compiled-in defaults (8 subscribers, 8 queryables) whatever they declare.

The queryable half of this same table was already applied "a row too narrowly the first time" per its own comment — that failure mode has happened once already, and the numbers go stale exactly when a leaf gains an entity, which is when being wrong costs the most.

`check-stack-floor.py` bounds the damage meanwhile: a leaf that gains an entity without gaining budget fails at **build** time rather than as a wild jump at runtime.

`just check fast`: 202/202 green.

PR #427, opened 2026-09-04
docs(phase-206): rescope — the user's own RMW config reaches the backend

jerry73204:
**Stacked on #420** (currently in the merge queue, so its branch is locked). The
first commit here is #420's; the queue's rebase will drop it as a duplicate.

## The rescope

The phase asked *"how does nano-ros express a NIC list and lower it to each
backend?"* Every answer requires nano-ros to own a **second vocabulary** for
interfaces, plus a resolver, a gate, and a per-platform story only Linux can
satisfy — `"eth0"` is a name in Linux's namespace, Zephyr's
`net_if_get_default()` cannot be named at all, and smoltcp has one `Interface`
and no names.

ROS 2 does not have this problem, because it does not have the abstraction. The
OS owns the devices; the user writes `CYCLONEDDS_URI` or a zenoh config naming an
IP or a device; application code never mentions a NIC. The principle the phase
now carries:

> **nano-ros transports the user's backend config VERBATIM; it does not
> re-model it.**

## The argument, in one citation

From the pinned Cyclone 0.10.5 we actually link (`third-party/dds/cyclonedds`, not
PX4's separate vendored copy):

```c
// src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h:79-88
GROUP("NetworkInterface", NULL, network_interface_attributes, INT_MAX,
  MEMBER(network_interfaces), ...
```

`INT_MAX` multiplicity — an **unbounded** `<NetworkInterface>` list, each with
`name`/`address`/`autodetermine`/`priority`. Its own description: *"Multiple
interfaces can be specified with an assigned priority."*

**The entire feature this phase set out to build already exists, complete, in the
config language of the backend that would implement it.**

## Two measured defects block the experience today

**1. Cyclone: the user's config REPLACES the tuned baseline.** `session.cpp:310-313`
is a 3-way ternary. Setting `CYCLONEDDS_URI` silently discards the baked
`<Threads>` stack sizes (64 KiB), `<Sizing>`, `MultipleReceiveThreads` and the
platform's `AllowMulticast`. On FreeRTOS/ThreadX those stack sizes are
load-bearing — **so doing the exact thing this phase supports detunes an RTOS
image today.** Gratuitous, because Cyclone merges comma-separated sources
natively (`ddsi_config.c:2538-2570`).

**2. zenoh: no user config surface off hosted-Rust.** The `ZENOH_*` env block is
`#[cfg(feature = "std")]` (`session.rs:370`), and the C boundary hardcodes
`properties: &[]` (`rust_adapter.rs:517`) — so a C/C++ entry gets nothing even on
Linux.

## Native parsers — measured, and the answer differs per backend

**Cyclone: yes, completely.** `dds_create_domain` accepts inline XML; any comma
token starting with `<` goes to `ddsrt_xmlp_new_string`, Cyclone's own parser. A
baked file's bytes go over verbatim and **nano-ros parses no XML**.

**zenoh-pico: there is no native config document to parse.** The whole API is
`z_config_default` / `zp_config_get` / `zp_config_insert`, and the key is a
**`uint8_t`** — ~22 numbered constants (`config.h:224-314`). No
`zc_config_from_str`, no file reader, **no JSON5 parser**; `json_encoder.h` is an
*encoder* for the admin space. JSON5 is zenoh-rs/zenoh-c's format and **this tree
has no zenoh-rs backend**.

So adding a JSON5 parser would mean nano-ros owning a parser for a format its own
backend does not speak — the same mistake in a friendlier hat. zenoh-pico's
native config language *is* the numbered key/value table, so the verbatim unit is
a flat `key = value` file, and the work is **unblocking a path that exists**.

## Work items

| W | |
|---|---|
| **W1** | Cyclone composes; the baseline always survives |
| **W2** | One baked-file passthrough seam, present on every target |
| **W3** | zenoh gets a config surface at all |
| **W4** | Device bring-up stated as the board's contract |
| **W5** | Delete nano-ros's interface abstraction |
| **W6** | Fast DDS — now needs no nano-ros work at all |

Every item pairs a declaration with a consumer **and an acceptance test that
fails today**.

## Deleted, with reasons recorded

- **206.1's multi-endpoint `SessionSpec`** — multi-homing is a config property of
  one participant; `open_multi` correctly segregates and is K.5.
- **Growing `rmw_session_options_t`** — nothing passes it non-NULL; a list field
  breaks the struct and touches C++ mirrors that have drifted 3×.
- **An intermediate design drafted and discarded the same day** — multi-homing as
  an `[rmw.capabilities]` entry plus a `[board.links]` inventory. Better placed
  than `set_interfaces` and still wrong, because it kept nano-ros modelling
  interfaces. Recorded because *why* it was discarded is the principle.

## Also noted

phase-419's own W3 measurement saw this doc and passed it — `phase-206 ticked=0
landed-marks=3 weak — likely prose, not a claim`. The marks were a real claim. R1
keys on ticked boxes and this doc has none, so the miss is **structural**, worth
knowing before anyone widens that rule.

## Verification

    check-doc-refs        OK
    check-roadmap-claims  OK — 0 findings
    check-book-links      OK — 706 links / 123 pages

### DIFF
--- a/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
+++ b/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
@@ -181,6 +181,26 @@ while (ok && cursor && cursor[0]) {
 Every comma-separated token parses into the **same `cfgst`**. Upstream's own
 tests rely on it (`"${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}<Discovery>…"`).
 
+**And multi-homing is already expressible in Cyclone's own schema**, which is
+the fact the whole rescope rests on. `src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h:79-88`,
+in the same pinned 0.10.5 tree:
+
+```c
+static struct cfgelem interfaces_cfgelems[] = {
+  GROUP("NetworkInterface", NULL, network_interface_attributes, INT_MAX,
+    MEMBER(network_interfaces), ...
+```
+
+`INT_MAX` is the multiplicity — **an unbounded `<NetworkInterface>` list**, each
+carrying `name` / `address` / `autodetermine` / `priority`. The `<Interfaces>`
+group's own description says it: *"Multiple interfaces can be specified with an
+assigned priority. The list in use will be sorted by priority."* (`:139-146`).
+
+So the entire feature phase-206 originally set out to build already exists,
+fully, in the config language of the backend that has to implement it. nano-ros
+does not need to express it, lower it, or resolve it — only to stop discarding
+the user's config and to carry the bytes to the parser.
+
 ### Defect 2 — zenoh: there is no user config surface off hosted-Rust
 
 Two independent cut points:
--- a/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
+++ b/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
@@ -5,8 +5,46 @@ common "node reachable on multiple interfaces" need that stock DDS/zenoh do
 natively. Turn the already-plumbed `[[transport]].interfaces` list into real
 per-backend NIC binding.
 
-**Status.** Proposed (2026-05-29). Extracted from Phase 172.K.7 (archived) — its
-schema + plumbing half landed; this phase is the deferred **wire-emission** half.
+**Status (2026-09-04). Proposed, and its foundation is GONE.** Still unstarted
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #441, which carries this rescope plus the W1/W2/W3/W4/W5 implementation. Keeping two PRs for one file was the source of the conflict.

PR #428, opened 2026-09-04
docs(phase-206 W3): follow zenoh-pico's own configuration model, not a new one

jerry73204:
**Follow-up to #427** (queued, branch locked). Rewrites W3 around zenoh-pico's
**documented** configuration model instead of an inferred one.

## What upstream actually says

`zenoh-pico.readthedocs.io/en/latest/config.html` defines three categories and
names the first as primary:

| rung | mechanism | owner here |
| --- | --- | --- |
| **Run-time options** — *"the primary configuration method"* | `zp_config_insert(config, KEY, value)` | **the user** — the surface to expose |
| **Manual compile-time** | edit `config.h.in` | upstream's, not ours |
| **Generated compile-time** | CMake flags | **nano-ros already does this correctly** — `nros-zpico-build` + RFC-0049's `builtin < platform < board < env` ladder |

**There is no config file format, and that is a position rather than a gap.**
zenoh's JSON5/YAML config files are documented for **`zenohd`**, the router;
`zenoh.io/docs/manual/configuration/` does not mention zenoh-pico at all.
Upstream's own examples do what the manual says — `z_config_default()` then
`zp_config_insert(…, Z_CONFIG_CONNECT_KEY, …)` (`examples/unix/c11/z_pub.c:46-49,
133-139`).

So nano-ros **exposes rung 1 and stops**: no fourth rung invented, and rung 3 not
duplicated into the user surface. The verbatim unit is a `key = value` list whose
key *names are upstream's own*, which is what makes existing zenoh knowledge
transfer.

## A third defect, found while checking the map

Our pinned 1.8.0 defines **23** config keys (`config.h:224-314`). The shim maps
**10** (`zpico.c:1213-1237`).

Missing: `USER`, `PASSWORD`, `SCOUTING_WHAT`, and **ten of the thirteen TLS
keys** — listen private key + certificate, mTLS enable, connect private key +
certificate, and their `_BASE64` variants.

It is a hand-written closed list, which is how it drifted. phase-347 W3 retired
four other closed lists for this reason. W3 now requires deriving or gating it,
with a coverage test against `config.h` so the next upstream bump cannot silently
re-open the gap.

## And an unknown key must be an error

`zpico.c:1238-1240` silently ignores one today. The keys are a numbered enum with
no upstream schema validation, so **nano-ros's map is the only place a typo can
be caught** — and a silently dropped config line is the exact failure mode this
phase exists to end.

## One trap recorded

The upstream manual is written for `latest` and documents
`Z_CONFIG_CONNECT_TIMEOUT_KEY`, `Z_CONFIG_LISTEN_TIMEOUT_KEY` and the
`EXIT_ON_FAILURE` pair. **None of the three exist in the 1.8.0 we pin**
(phase-415, 2026-09-04). Anyone planning against the website writes code that
does not compile.

## Verification

    check-doc-refs        OK
    check-roadmap-claims  OK — 0 findings

### DIFF
--- a/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
+++ b/docs/roadmap/phase-206-multi-homing-transport-interfaces.md
@@ -181,6 +181,26 @@ while (ok && cursor && cursor[0]) {
 Every comma-separated token parses into the **same `cfgst`**. Upstream's own
 tests rely on it (`"${CYCLONEDDS_URI}${CYCLONEDDS_URI:+,}<Discovery>…"`).
 
+**And multi-homing is already expressible in Cyclone's own schema**, which is
+the fact the whole rescope rests on. `src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h:79-88`,
+in the same pinned 0.10.5 tree:
+
+```c
+static struct cfgelem interfaces_cfgelems[] = {
+  GROUP("NetworkInterface", NULL, network_interface_attributes, INT_MAX,
+    MEMBER(network_interfaces), ...
+```
+
+`INT_MAX` is the multiplicity — **an unbounded `<NetworkInterface>` list**, each
+carrying `name` / `address` / `autodetermine` / `priority`. The `<Interfaces>`
+group's own description says it: *"Multiple interfaces can be specified with an
+assigned priority. The list in use will be sorted by priority."* (`:139-146`).
+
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #441 — same reason as #427: its W3 doc content is in #441 alongside the implementation it describes.

PR #441, opened 2026-09-04
feat(phase-206 W1+W5): Cyclone config composes, and nano-ros stops modelling NICs

jerry73204:
**phase-206 W5.** Stacked on #437 (RFC-0004 amendment), which it extends.

Delete the interface abstraction rather than finish it. Devices are the board's
job and are up before ROS exists; the user configures the middleware in the
middleware's own language.

## The sweep — all seven setters, not just the reported one

| method | impls | callers | action |
|---|---:|---:|---|
| `set_interfaces` | 0 | 0 | **deleted** |
| `set_mac` | 0 | 0 | **deleted** |
| `set_gateway` | 0 | 0 | **deleted** |
| `set_ssid` | 0 | 0 | **deleted** |
| `set_password` | 0 | 0 | **deleted** |
| **`set_ipv4`** | **5 boards** | **0** | kept → **issue 1064** |
| **`set_baudrate`** | **2 boards** | **0** | kept → **issue 1064** |

Callers: **zero for all seven.** Their single writer was the orchestration
generator, deleted with the standalone-package pipeline in `11a00b0f8` (#202).

Deleting a seam with no impl and no caller is a cleanup. Deleting one with **five
impls** is a decision about *who writes a board's IP now the generator is gone* —
and the deploy-overlay path (`nros::main!` → `run_with_deploy`) may already be
the live answer, which nobody has measured against those five bodies. **Issue
1064** owes that; it does not belong in this phase.

## Bare removal, not a deprecated alias — checked first

No `[[transport]]` block in the tree, in `examples/`, or in any fixture
`system.toml` / `nros.toml` / `nros-plan.json` carries `interfaces`. An alias
would have re-introduced the vocabulary to serve a path no input uses. (The
`interfaces` in the plan fixtures is the unrelated top-level topic list.)

**The upstream schema is not ours.** `TransportBlock.interfaces` lives in
`ros-launch-manifest` (pinned v0.1.21), so the key can still *arrive* from a
user's `system.toml`. `plan_transports` now **bails** naming the backend config
that owns NIC binding, through one shared `plan::RETIRED_INTERFACES_REMEDY` — a
second spelling of a remedy is how this class comes back. Dropping it silently
would be strictly worse than the no-op it replaced.

## Tests

The three round-trip tests become one asserting a plan naming `interfaces` fails
to **parse**, plus one asserting the ingest refusal **names the remedy**. Without
them, re-adding the field would read as a green refactor.

```
cargo test -p nros-cli-core --lib   → 870 passed, 0 failed
cargo check -p nros-platform        → clean
check-doc-refs / check-roadmap-claims / check-book-links → OK
```

## Integration notes

The subagent that did the sweep branched before the phase-206 rescope, so two of
its edits were reconciled by hand:

- **It archived the phase doc as RETIRED.** Un-archived — W5 is one of six items
  and W1/W2/W3/W4 are open. Its sweep table is spliced into the live doc's W5
  section, with the two deviations (bare removal, `set_ipv4` kept) recorded.
- **Its RFC-0004 edit collided with #437.** Took its cleaner half — the retracted
  field rows are *removed* from the per-kind table rather than annotated in place
  — and kept #437's measured evidence (Cyclone's `INT_MAX` `<NetworkInterface>`
  list, comma-composition, and why zenoh-pico gets key/value rather than JSON5).

Also: `api-parity-ledger/other.json` loses 5 orphaned rows, and
`rust:BoardTransportConfig`'s `why` is rewritten — it described the deleted
generator in the present tense. `--check` has no orphan detection, so those were
silent rot, not a red gate.


jerry73204:
## W1 added — Cyclone config sources compose, they do not select

`session.cpp`'s three-way ternary meant a user setting `CYCLONEDDS_URI` **silently lost the baked baseline** — the `<Threads>` stack sizes (64 KiB for `dq.builtins`/`recv`/`dq.user`), `<Sizing>` receive buffers, `MultipleReceiveThreads=false`, and the platform's `AllowMulticast`. On FreeRTOS/ThreadX ddsrt defaults a thread to 1 KiB, so *"point me at a different peer"* also meant *"and reinstate that stack overflow"* — in a config mentioning neither.

Now `compose_cyclone_config(baseline, kconfig, user)` joins with `,` into a 4 KiB buffer and **fails loud** rather than truncating (a clipped config is unterminated XML, which Cyclone would report as a parse error against a string nobody wrote).

### Precedence verified, not assumed — and I re-derived it independently

```c
// third-party/dds/cyclonedds .../ddsi_config.c:1619-1622  (our pin, 0.10.5)
if (cfgelem->multiplicity == 1 && n->count == 1 && source > n->sources)
    free_configured_element (cfgst, parent, cfgelem);
if (cfgelem->multiplicity == 0 || n->count < cfgelem->multiplicity)
    res = upd (...);
```

A scalar (`multiplicity == 1`) from a later source frees the earlier element, so **later tokens win** → baseline-first is correct. The agent also measured it against the pin: `baseline,user` → `AllowMulticast=false` (user's) with thread stacks intact; `user,baseline` → `AllowMulticast=spdp` (baseline's).

### The caveat, stated rather than papered over

`multiplicity == 0` (a **list**, which `<Thread>` is) takes the `upd` branch unconditionally — it **appends**, never replaces — and `lookup_thread_properties` (`q_thread.c:280-288`) returns the **first** match. Measured: `recv` at 64 KiB then 128 KiB yields two entries and 64 KiB is used.

**So a user cannot override a stack size the baseline names.** Safe in the direction this phase needs, but it is not "the user always wins", and it is now written into the phase doc.

### Test

`tests/cyclone_config_compose.cpp` composes the real baseline with a user fragment stating only `<Discovery>`, brings a domain up, and reads the **resolved** values back out of Cyclone's `config` trace category. It carries its own **negative control** (the user fragment alone — exactly what the ternary passed), so the baseline assertions cannot be satisfied by Cyclone's defaults.

Against the pre-fix selection semantics: **10 checks fail**, 4 of them the "did not survive" ones. After: `100% tests passed, 25/25`.

**Not run:** tier 2 / fixture builds. This is host-tier plus `-fsyntax-only` on the ThreadX and native_sim arms — not a claim that those images build and run.

PR #446, opened 2026-09-04
refactor(#417 W-B5): 112 deprecated forwarders retired, and a `<chrono>` regression I shipped this morning

jerry73204:
Stacked on #425 (step B), which is stacked on #329. Review those first.

## First: a regression this fixes, found by the node-type design survey

The `NROS_CPP_STD`-only gate I put on `<chrono>` earlier today was measured on
the Zephyr lane and never on a plain hosted one. **Nothing that ships defines
`NROS_CPP_STD`** — no cmake module, no toolchain file, no `build.rs`, no
Kconfig; only docs and `check.just`'s own probe TUs. So the gate did not tighten
`<chrono>`, it **deleted `create_wall_timer` and `Rate` from every
configuration**, hosted included — where `cpp-port-minimal-publisher` calls
`create_wall_timer` and is phase-417's own acceptance criterion.

I had also added `-DNROS_CPP_STD` to the probe loop that same morning, which is
why the probes were green while the shipped configuration was broken: the probe
was testing a configuration nobody builds. Removed.

The correct predicate needs both halves, because the lanes fail differently:

| lane | `__has_include(<chrono>)` | `__STDC_HOSTED__` |
| --- | --- | --- |
| Zephyr / ThreadX-RV64 (`-nostdinc++`) | **false** — right on its own | 1 |
| FreeRTOS armcm3 (`-ffreestanding`) | true — GCC 13 then `#error`s via `requires_hosted.h` | **0** |
| hosted | true | 1 |

`__STDC_HOSTED__` alone is still wrong, exactly as CLAUDE.md's pitfall says.
Verified three ways: a hosted `#error` probe (fails before, passes after), the
real recorded Zephyr compile replayed from `compile_commands.json`, and
`just check cpp`.

## W-B5, C half — 107 forwarders

Every one had zero call sites, verified per name with `grep -rIw` and **no
`--include` filter**. The phase doc's two figures were both wrong: "43 names" is
only the stage-6 rcl/rclc subset, and "110" over-counts. The other 64 are
phase-379-era renames (`nros_param_*`, `try_recv`, `send_reply`,
`executor_stop`).

Plus **4 typedef aliases** — the case W-B6's changelog has to carry alone,
because a typedef cannot hold a deprecation warning, so a renamed C type
disappears silently.

Collateral, all compiler-forced: **5 deprecation probes deleted** (expected-
failure compiles that, with the identifiers gone, would keep *passing while
asserting nothing*), 4 alias probes trimmed **and renamed** so the filename
stops naming a test that no longer exists, and one real call site the grep
review had mis-filed as probe-only.

Independently broken, fixed while here: `#[cfg(kani)] mod verification` in
`nros-c/src/service.rs` called `nros_service_init` in eight harnesses — a name
with no Rust definition. `cargo kani` could not have compiled, and `just check`
never sees `cfg(kani)`.

## W-B5, Rust half — 5 forwarders

`nros_debug/info/warn/error/fatal!` deleted. **`nros_trace!` survives**: rclrs
stops at `debug`, so TRACE has no upstream twin.

`deprecated_forwarders.rs` contained
`trace_keeps_our_prefix_because_rclrs_has_no_trace` — an assertion about the
macro that *survives*, in a file named after the ones that do not. Moved to
`trace_is_not_renamed.rs` rather than deleted with its neighbours.

Three leftovers found by **building, not grepping**: 3 sites in `exclude`d
example leaves no workspace-wide `#[deprecated]` can enumerate; 2 facade
re-exports invisible to a `!`-anchored grep; and 2 gate scripts whose failure
text told users to write a now-deleted macro — a wrong instruction issued
exactly when someone is stuck.

## Follow-up this measured but did not do

`api-parity.py` extracts our C surface with `prefixes={"nros_", "NROS_"}` only,
so every `rcl_*`/`rclc_*` entry point adopted in stage 6 is **invisible to the
parity tool** — the deprecated forwarders were standing in for them. Measured
`same`: 85 at base, 39 after deletion, 101 with a two-line prefix fix. That fix
was tried and reverted: it leaves 5 genuine ledger questions, and authoring
verdicts is the correction track's job.

## Verified

`just ci gate`: `check-fast` 202 OK. The two `check-build` reds are the ones
**#438** fixes on `main` (`type Format`, `format!` in `assert!`) and are not
from this branch.


jerry73204:
Not rebased — it needs you, not a mechanical resolution.

The conflict is `docs/reference/api-parity-ledger/node.json`, on all 37 commits. That file is a flat JSON object written by `scripts/api-parity.py` with `sort_keys=True, indent=1`, so its on-disk order carries no information and a textual 3-way merge is the wrong tool — two branches adding disjoint keys land at the same byte offset and conflict with certainty.

A union merge is the right shape and it does not work here. The first commit merged cleanly (`116 ours + 120 theirs -> 122 keys, 0 clashes`), but every commit after it reported **26–27 keys present on both sides with different values** — `c:node_init_with_options`, `c:node_get_rmw_handle`, `c:node_impl_t` and the rest. Preferring `main` there discards precisely what this branch is for: retiring 112 forwarders is a re-verdicting of those keys, so the clashes are the change, not noise.

Aborted rather than push a ledger with this branch's own edits silently reverted.

Two things from the sweep that touch this branch:

- The whole class is measured in #468 / issue 1072 — the parity ledgers are 3 of the 22 conflicting PRs, and the fix has a constraint: 0884 established that GitHub rebases queue entries **server-side**, where a `.gitattributes` union driver does not run. Splitting one file per key is the untracking-shaped answer that would reach the queue.
- Issue 1071: PR #431 deleted four gates from `just/check.just` and `check-gate-lists` passed at 232 gates where `main` has 236. Worth knowing before your next rebase of a branch this old — the swept check found no other instance, including this one.

jerry73204:
Rebased onto the current `phase-417-ros2-api-adoption` tip.

The branch had been rebased across a base force-push, so it was carrying **15 stale copies** of base commits under their old shas — a plain `git rebase <base>` tried to replay them and conflicted immediately. Replayed only this branch's own work instead:

```
git rebase --onto origin/phase-417-ros2-api-adoption b5cdad8a6 phase-417-w-b5-retire-forwarders
```

Nine commits went in, two dropped as already applied (patch-id dedup), **seven remain**. One conflict, in `.config/codegen-version-surface.txt` — a generated evidence file, so it was regenerated with `--write-baseline` rather than hand-merged: 248 surface items, version 2, gate green.

`just check fast`: 232/232 on the rebased tree.

## The base does not compile, and that is why this is still UNSTABLE

`just check c` fails here, and it is not from the rebase — `origin/phase-417-ros2-api-adoption` carries it, and #329's own `CI` is FAILURE:

```
error[E0592]: duplicate definitions with name `halt`
error[E0592]: duplicate definitions with name `is_halted`
error[E0609]: no field `spinning` on type `&Executor<'s>`   (×3)
error[E0034]: multiple applicable items in scope
```

Two causes, both in `packages/core/nros-node/src/executor/spin.rs` on the base branch:

- **the pre-rename block was never deleted.** `halt`/`is_halted` exist twice — once at ~8031 as the deprecated forwarder onto `cancel()` (the W4.c shape), and again at ~8407 as the original that sets `halt_flag` directly.
- **`is_spinning()` reads a field that does not exist.** `self.spinning` is referenced at three sites; `struct Executor<'s>` declares `halt_flag` and `wake_flag` and no `spinning`.

So W-B5 cannot go green until #329 does. Nothing else is needed on this branch.

PR #462, opened 2026-09-04
feat(sched): alive supervision — report a context that stopped running

jerry73204:
## The gap

AUTOSAR's Watchdog Manager separates **alive** supervision (did it run at all, at roughly the right rate) from **deadline** supervision (did it finish in time). Every rule in `monitor.rs` was the second kind — they fire when something happens and is wrong.

**Nothing fired when a callback stopped happening altogether**, which is the failure a watchdog exists for.

`rate-hierarchy-runtime` does not cover it: that rule counts *publishes* on a contracted publisher, so a timer callback that silently stops firing while its topic is still published from elsewhere — another tier, a bridge, a republisher — leaves it satisfied. The gap is a callback, not an endpoint.

## No new contract field

`period_us` is already on the `SchedContext`. A context claiming a period that produced **no** activation in a window many times longer than that period is a contract failure for any period — the same argument `check_timer_overrun` makes for dropped activations. `period_us == 0` declares no cadence and is not judged.

## Zero, not a fraction

A throttled-but-alive context is a different fault with different evidence — `SporadicState` already counts window skips and dispatches for exactly that — and picking a percentage here would invent a tolerance nobody declared. Zero activations against a declared period is unambiguous.

Report-once-until-recovery, like the rate rule: continued silence is not a new fault, but silence *after* a recovery is.

## Counting before the class gate

The increment sits ahead of the `SchedClass::Sporadic` check in the dispatch path, deliberately. A `Fifo` or time-triggered context can fall silent exactly as a sporadic one can, and scoping the count to the class that happens to carry a budget is how issue 0736's per-callback runtime accounting ended up on a path no shipped image took.

Per-SC state is a fixed `[_; MAX_SC]` array alongside `monitor_states`, not an arena allocation — `MAX_SC` is a build-time constant, 8 by default.

## Silent without a clock

By construction. The rule cannot distinguish a stalled context from a fast one without time, so it reports nothing rather than guessing.

## Verification

```
cargo test -p nros-node --features std
  test result: ok. 319 passed; 0 failed
  test result: ok. 5 passed; 0 failed

alive_supervision_tests::a_zero_period_declares_no_cadence ... ok
alive_supervision_tests::the_first_window_only_opens ... ok
alive_supervision_tests::silence_across_a_full_window_reports ... ok
alive_supervision_tests::a_single_dispatch_is_enough ... ok
alive_supervision_tests::continued_silence_is_not_a_new_fault ... ok
alive_supervision_tests::a_period_longer_than_the_window_is_not_judged ... ok
```

Clean on default (no_std) and `alloc`; `cargo fmt` clean.

## Worth knowing for this file

**`cargo check -p nros-node` does not compile `spin.rs`.** `mod spin` is `#[cfg(any(has_rmw, test))]`, and a plain check selects no RMW backend, so the module is skipped entirely — appending invalid Rust to it passes cleanly.

Two call-site errors in this change were invisible to `cargo check` and appeared only under `cargo test`, which pulls the module in via `cfg(test)`. Anything touching that file needs `cargo test` to be validated at all.

### DIFF
--- a/packages/core/nros-node/src/executor/monitor.rs
+++ b/packages/core/nros-node/src/executor/monitor.rs
@@ -180,8 +180,7 @@ pub const MAX_VIOLATIONS: usize = 8;
 pub struct Violation {
     /// `"rate-hierarchy-runtime"` | `"max-age-runtime"` |
     /// `"max-latency-runtime"` | `"deadline-miss-runtime"` |
-    /// `"timer-overrun-runtime"` | `"release-jitter-runtime"` |
-    /// `"stack-headroom-runtime"`.
+    /// `"stack-headroom-runtime"` | `"alive-supervision-runtime"`.
     pub rule: &'static str,
     /// Violating endpoint ref (from the spec's `fqn`; the SC name for
     /// deadline misses).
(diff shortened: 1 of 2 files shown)


jerry73204:
Rebased onto `main`. The branch was **486 commits behind**, which is what made it worth checking rather than just re-queuing.

A gate sweep reported `prose-issue-refs` failing here, and the failure was not the gate's: `error: justfile does not contain recipe `check prose-issue-refs``. The branch predates the recipe entirely. That is the shape worth naming — a lane that cannot RUN and a lane that RAN and found something look identical from a red mark, and only one of them is about this diff.

After the rebase, the one commit replays with no conflict — the diff is 213 lines across `executor/monitor.rs` and `executor/spin.rs`, and nothing it touches moved underneath it in 486 commits.

Verified on the rebased tree: `prose-issue-refs`, `issue-ids`, `doc-recipe-refs`, `doc-refs` all OK; `cargo +nightly fmt --check --all` clean; `cargo check -p nros-node --all-targets --features std` clean (the `--features std` is required — `nros-node` is `no_std` by default and its tests need it).

1 commit before, 1 after.

jerry73204:
Rebased, and fixed the reason this **ejected a merge-queue batch of five**.

`just check workspace-all` went red on `workspace-embedded`:

```
error: this `impl` can be derived
  --> packages/core/nros-node/src/executor/monitor.rs:731:1
  = note: `-D clippy::derivable-impls` implied by `-D warnings`
```

`AliveState`'s hand-written `Default` sets `false, 0, 0, false` — every field to its own type's default. It restates four values the compiler already knows and is one added field away from disagreeing with the struct. Derived instead. It is also what this very file already does twice: two other structs here carry `#[derive(Debug, Clone, Copy, Default)]`, so `AliveState` was the odd one out rather than following a local convention.

## Why the PR lane could not see it

The required `CI` context on a pull request is `check-fast` + `check-submodule-commits-reachable` + `check-compile-smoke` + `check-cli-tests`. **`workspace-all` is not in it** — it runs on `merge_group`. So this was invisible until the queue, and by then it took four unrelated PRs down with it: #741, #768, #779 and #820 were all removed from the same batch. They are re-armed and back in the queue.

That is the compile-tier case CLAUDE.md makes — the local tier is stronger than the gate, so `just ci gate` before pushing catches what the PR lane deliberately does not run.

## Verified

`just check workspace-embedded` exits **0**. Worth noting for anyone reproducing: it needs `packages/rmw/zenoh/zpico-sys/zenoh-pico` checked out, and a fresh worktree fails it on the missing submodule *first* — which is not this lint and reads nothing like it.

`cargo clippy -p nros-node --all-targets --features std -- -D warnings` clean; `cargo test -p nros-node --lib --features std` passes; `cargo +nightly fmt --check --all` clean. 1 commit before, 1 after; no-loss check against the recorded remote sha empty.

PR #471, opened 2026-09-04
fix(#417): unblock the node-type merge — corrected docs, an instrument that can fail, and a shipped ODR bug

jerry73204:
Stacked on #446 → #425 → #329. Review those first.

Three preconditions for the `nros::Node` / `rclcpp::Node` merge. Two of the
three turned out to be repairs to things already broken.

## 1. The blocker was removed a month ago and both documents still described it

`phase-417`'s "structural blocker" section and RFC-0089's matching paragraph
both argued from the shim `rclcpp::Node` pumping its own callbacks via
`pump()`, citing `rclcpp_compat.hpp:428-451`. **That header was deleted in stage
6 step A and `pump()` by the one-dispatch-path pass.** `nros.hpp` says so in its
own words and records the change at the call site: every entity a
`rclcpp::Node` creates is arena-registered through the same call the native path
makes.

The paragraph outlived the fix and was still being cited as the reason nothing
scheduled the merge. That is the cost of stating a prerequisite in two documents
and retiring it in neither.

Two claims those sections rested on are also measured false, and both were mine:

* **`rclcpp::Node` is not hosted-only.** It is declared on bare-metal NuttX
  armv7a and riscv32. The real partition is `-nostdinc++` versus not, which only
  Zephyr and ThreadX-RISCV64 fall on.
* **Five of W-B5's six C++ items were never waiting on the merge.**
  `spin_once`, `Timer`, the lifecycle set, `GoalResponse`/`CancelResponse` have
  **zero upstream occurrences** — ours-only names for capabilities ROS 2 lacks,
  so retireable *never* rather than later.

## 2. The instrument could not fail — and found a live red on its first run

`just check cpp`'s "freestanding syntax" probe is host `c++ -ffreestanding`
against the host's **complete** libstdc++: `__has_include(<memory>)` is true,
`<chrono>`'s `duration_cast` is present, `std::abort` always resolves. A useful
syntax check; not a freestanding check.

Added a `-nostdinc++` lane against the two shims real lanes build against —
ThreadX-RISCV64's `cxx-compat`, and `zephyr/cxx-compat` over Zephyr's own
three-header minimal libcpp. Host compiler, real shim: header **parse**, not
codegen, so no cross toolchain. It does not model the C library, and the comment
says so.

**First run found:** RFC-0089 W3.b's refusal path calls `::std::abort()`
(`nros.hpp:369`); the ThreadX shim's `cstdlib` re-exported
malloc/calloc/free/getenv/realloc and **not abort**; and
`examples/qemu-riscv64-threadx/cpp/talker` includes `<nros/nros.hpp>`. Every
ThreadX C++ image stopped compiling the moment `rclcpp::init` entered the
umbrella header. One missing `using ::abort;`.

## 3. A capability probe was changing a layout, and it shipped

`rclcpp::Node` held `std::vector<std::shared_ptr<detail::WallTimer>> timers_`
behind `NROS_CPP_HAS_STD_CHRONO`, reachable only through `NROS_CPP_STD` — a
macro nothing in the build system sets, **except
`examples/px4/.../nros_uorb_bridge/CMakeLists.txt:123`, on one module of a
larger image.**

So that module compiled a **3776-byte** node and every other TU compiled a
**3752-byte** one. They linked. Each wrote the object through its own layout.
Measured, not inferred — restoring both halves reproduces those numbers exactly.

**Fix (1):** `timers_` is gone; cells go in the unconditional
`owned_entities_`. Its only remaining job was keeping cells alive — the typed
vector existed so the deleted `pump()` could iterate it.

**Fix (2):** `check-cpp-capability-layout.sh` measures the invariant: 3 types ×
7 capability macros, reading `sizeof` out of the compiler's own diagnostic for
an incomplete `ShowSize<sizeof(T)>`. No link, no run, no library.

Two TUs of one image disagreeing about a capability is a **supported** state —
px4 sets the macro deliberately, `nros_rmw_cyclonedds.cmake` adds an include dir
for some targets only. And the probes are unreliable: three failure modes
measured in one day (`<type_traits>` present-but-hollow on Zephyr,
`NROS_CPP_STD` set by nothing that ships, `<memory>` present-then-`#error` under
`-ffreestanding`). So the rule is not "get the probe right" — it is "make being
wrong unable to change a layout".

## Things this got wrong first, kept for the next reader

A text scanner for "member inside an `#if`" was written and thrown away: it
could not tell a member from a local variable, and the file-level include guard
made every line look conditional — 200+ false positives, the same
guard-blindness that made an earlier scanner in this tree report **zero**.

`check-gate-selftests` then caught the replacement shipping without a negative
control. That matters more than usual here: forcing a macro already on in the
baseline is a no-op, so the gate could have looked green while detecting
nothing. It now asserts both directions — a conditional member must change
`sizeof`, a gated method must not.

## Verified

`just ci gate`: `check-fast` 202 OK. The two `check-build` reds are the ones
**#438** fixes on `main` and are not from this branch.


jerry73204:
Correcting a mistake of mine on this PR, and the correction is the interesting part.

**This is a stacked PR** — its base is `phase-417-ros2-api-adoption` (#329), not `main`. I rebased it onto `main` twice while sweeping conflicting PRs, which is wrong for a stacked branch: it pulled `main`'s commits (and #329's, re-applied under new SHAs) onto a branch whose base is #329. GitHub kept reporting `CONFLICTING` because it computes mergeability against the **base branch**, and each main-rebase pushed the branch further from it.

Rebuilt correctly: `origin/phase-417-ros2-api-adoption` plus this branch's own **20** commits, cherry-picked in order. Now `MERGEABLE`.

How I identified "its own": subjects present here but in neither #329 nor the last 400 commits of `main` — 21 of them. Twenty applied cleanly; the twenty-first (`273f7f21e`, *"two log sites W-B1 missed, and a rustup probe that could answer twice"*) came out empty, and I checked rather than assumed — all three of its files (`preflight.rs`, `bridge/src/cffi.rs`, `bridge/src/lib.rs`) are byte-identical to what the branch already has, so it was a genuine duplicate.

I also diffed the resulting tree against the pre-rebase tip. The only paths that disappeared are ones `main` has since deleted — `book/src/reference/{static-pool-inventory,configuration-surface}.md` and the typed entry templates — so nothing of this branch's work was dropped.

One thing worth knowing for the stack generally: **a stacked PR cannot be armed for auto-merge.** `gh pr merge --auto` on it returns `Protected branch rules not configured for this branch`, because the ruleset and the merge queue are on `main` and this PR targets #329's branch. It becomes queueable only once #329 merges and GitHub retargets it. #546 (`phase-431 W4` on top of W3) is in the same position.

jerry73204:
Rebased onto current main: **20 commits -> 18**, and the two that dropped are accounted for.

The conflict was add/add on `docs/roadmap/phase-428-api-porting-principle-sweep.md`: both this branch and main create that file, and main has since grown it to 904 lines against this branch's 225. Main is a strict superset — the single line unique to this branch was `NROS_RMW_RELIABILITY_UNKNOWN` exists and is unused, which main does not merely also contain but **corrects** ("False — it is written…"). So every conflict on that file was resolved by taking main's version.

The two dropped commits are `f9fbbd1a9` ("fold the RMW sweep findings and revise the work items") and `0622eb629` ("correct the zenoh QoS finding"). Both touch **only** that one doc, so once main's version won they had nothing left to apply; their substance is already on main. Nothing else was lost — the branch's 14 non-doc files are intact, and the doc is now byte-identical to main's.

Verified: no conflict markers anywhere in the tree, `just --list` clean.

PR #480, opened 2026-09-05
fix(#1075): the Zephyr stack probe is guarded at COMPILE time, not run time

jerry73204:
**Stacked on #477** (queued while this fix was being verified, so its branch was
locked). Carries #477's issue commit plus the fix.

Every Zephyr `native_sim` image failed at the final link with `undefined
reference to z_impl_k_thread_stack_space_get`. **Found by a tier-2 fixture
build; the compile tiers were green throughout, which is how it reached main.**

## The mistake

`411addfd2` documented a fallback:

> Needs CONFIG_INIT_STACKS to have painted the stack; without it the kernel has
> no watermark to read and **returns an error, which becomes the documented 0**.

That reasoning is about **run time**; the failure is at **link time**. There is
no function to return that error — `zephyr/kernel/thread.c:806` encloses
`z_impl_k_thread_stack_space_get` (`:862`) in:

```c
#if defined(CONFIG_INIT_STACKS) && defined(CONFIG_THREAD_STACK_INFO)
```

Without both symbols the implementation isn't compiled at all. The comment also
named only the first of the two, and **neither is set anywhere in this repo**.

## The fix

The body is compiled two ways: the real probe under both symbols, `return 0`
otherwise — the ABI's documented *"this port does not instrument it"*, the same
answer `nros_platform_heap_used_bytes` gives. An image without the Kconfigs
reports no headroom rather than failing to build.

**Both symbols, deliberately.** Guarding on `CONFIG_INIT_STACKS` alone
reproduces the bug wherever only that one is set — the exact mistake the original
comment made. The old comment is kept in the replacement, quoted, with why it was
wrong: someone will read this function again and the trap is worth leaving
visible.

## Verified by rebuilding the fixture that failed

Not by a compile-tier check — that passed before the fix and would pass after it.

```
just zephyr build-one cpp/service-client xrce

before:  undefined reference to `z_impl_k_thread_stack_space_get'
         ninja: build stopped: subcommand failed.

after:   [148/148] Running utility command for native_runner_executable
         Built: …/zephyr.elf  →  zephyr.exe, 11.8 MB, exit 0
         undefined references: 0
```

The two surviving `ld: error in …(.eh_frame); no .eh_frame_hdr table will be
created` lines are unrelated and predate the fatal error.

## Not done

Turning the two Kconfigs **on**. That would make the probe return a real number
instead of `0`, at the cost of stack painting at every thread start — a decision
about the shipped image, belonging with whoever needs the safety argument the
probe was written for. Recorded as option B in the issue.

### DIFF
--- a/docs/issues/archived/1075-zephyr-stack-headroom-probe-breaks-native-sim-link.md
+++ b/docs/issues/archived/1075-zephyr-stack-headroom-probe-breaks-native-sim-link.md
@@ -1,7 +1,7 @@
 ---
 id: 1075
 title: "The Zephyr stack-headroom probe calls a syscall that is not COMPILED unless two Kconfigs are set, so every native_sim image fails to link"
-status: open
+status: resolved
 type: bug
 area: platform, build
 severity: high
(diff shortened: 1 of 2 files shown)


jerry73204:
Closing: superseded. `b9fb602d9` (*"fix(#1075): the Zephyr stack probe calls a syscall Kconfig can compile OUT"*) landed the same fix on main independently, and archived the issue doc with it.

The two guards are functionally identical — both key on `CONFIG_INIT_STACKS && CONFIG_THREAD_STACK_INFO`, which is the pair `zephyr/kernel/thread.c` encloses `z_impl_k_thread_stack_space_get` in. Main's places the `#if` inside the function body; this PR wrapped two whole definitions. No behavioural difference, and main's is the one that shipped.

Verified in passing that the fix is real and load-bearing: a `lane=tier2` fixture build on main *before* this landed died at

```
.../generated/zephyr/syscalls/kernel.h:102: undefined reference to `z_impl_k_thread_stack_space_get'
```

and with the guard applied, `== zephyr == OK`.

PR #482, opened 2026-09-05
phase-420 W9 (zenoh) + #1069: one derivation per vendored tree, and two gates that check wiring rather than shape

jerry73204:
Builds on #473 (issue 1068, now merged) — its `xrce-sources.txt` is the sibling these sit beside.

Two vendored trees, the same defect class in each: a fact restated by hand in two lanes, with nothing checking they agreed.

## phase-420 W9 — zenoh-pico gets one source list

zenoh-pico is compiled twice, and each lane held its own copy of the selection: the same nine directory globs, written out twice. `packages/rmw/zenoh/zpico-sys/zenoh-sources.txt` is now the one list.

It names **directories**, unlike `xrce-sources.txt`, and the manifest header argues why: XRCE's list is a deliberate *subset* of a larger tree, so every line carries a decision; zenoh-pico's selection is a *rule* — the whole core, recursively — on a rebased patch line against a moving upstream. Expanding it to ~130 paths would restate the rule as data and lose it, so the next reader could not tell an omission from an exclusion. The safety a file list would buy is bought instead by asserting every `.c` in the tree is covered by a record, left to the per-platform axis, or on a documented not-compiled list — so a new file in a listed directory compiles, and a new *directory*, the case that needs a human, fails.

**Second defect, closed inline:** the old `file(GLOB_RECURSE ...)` had no `CONFIGURE_DEPENDS`. A `.c` appearing from an upstream rebase had no rebuild edge at all.

Verified against the pre-W9 selection rather than asserted — cargo: a compiler-wrapper capture of every `.c`, 134 sources, diff empty; cmake: a configure that stubs the `zephyr_library_*` calls and runs the real function against a transcription of the old globs, 133/133 noisotp and 134/134 isotp, both empty. No west build was run; that is the one thing not verified end to end.

## Issue 1069 — both lanes read the version out of the tree that has it

`MICROCDR_VERSION_STR` compiled as `2.4.1` under CMake and `2.0.2` under cargo, for a tree that is 2.0.2; the client said `2.4.1` in **both** lanes against a 3.0.1 checkout. Both templates take the same `@PROJECT_VERSION*@` tokens and the CMakeLists set them once for the client and never reset them. Both lanes now read `project(<name> VERSION …)` out of the vendored `CMakeLists.txt`, and the re-point is explicit rather than a property of ordering.

Two corrections to the issue's own text, both measured:

- Its proposed `git describe --tags` **cannot work** — a submodule is fetched by SHA with no tags, so it answers `No tags can describe` on a normal checkout.
- Its "zero readers, so the drift is cosmetic" is **false**. `uxr/client/config.h.in` ends with `#if UXR_CLIENT_VERSION_MAJOR >= 4` guarding three `#error`s — upstream's own tripwire for work due at 4.x. Compiling `MAJOR = 2` against a 3.0.1 tree disarmed it by a whole major version, in both lanes.

The `nros-sdk-index.toml` rows stay authored and are gated instead: `nros setup --source` runs precisely when the checkout is absent, so it cannot read the tree. `micro-xrce-dds-client` said `2.4.3-nros1`, which is not a client version at all — it is the agent tool's release label, copied across.

## Both gates were rewritten after review found the same hole in each

Each gate's first version checked **shape** — no literals, a block exists, both lanes answer every token — and each let a shape-preserving mutation through:

- XRCE: point micro-CDR's derivation at the *client's* tree. Every shape check passes; the configure emits `MICROCDR_VERSION_STR "3.0.1"`, which is 1069 verbatim.
- zenoh: move `dir core … utils` to `dir zephyr_system … utils`. Every count in the gate's own output is identical; nine `.c` vanish from every non-Zephyr build, sliding under the reader's "selected no sources" guard because 125 is not 0.

Both now check attachment, not shape. XRCE declares each tree's every alias once — dirname, upstream project id, template, CMake dir var and prefix, Rust binding, index row — and checks each hop, resolving `${var}` indirection to assert the four `PROJECT_VERSION*` re-points before each `configure_file` come from that template's own prefix. zenoh states the rule its own manifest header implies: **platform-conditional compilation is legitimate only for the platform trees** — a path under `src/system/<platform>/` must sit in a conditional group, the core must not. Nine mutations each, every one watched fail and restored. Both stay buildless.

Recorded rather than glossed: the cargo lane's crosswires are **not** silent — each generator takes one `&Path` used for both the template read and the derivation, so a crosswire fails to compile or panics on a missing template. Those checks state the invariant instead of relying on today's function signatures; they were not load-bearing.

## Left open, deliberately

`scripts/gen-support-status.py` hard-codes "client and agent are pinned to the same upstream release" — false at 3.0.1 vs 2.4.3, and always was; the index restating the agent's number is what hid it. Outside these files, worth its own issue.

## What remains in W9

The XRCE vendor build: `microxrcedds_client` + `microcdr` targets and a cargo `links` crate, so `nros-rmw-xrce` and `nros-rmw-xrce-cffi` consume one build instead of two. #473 and this PR remove the two duplicated *lists*; the duplicated *build* is the rest.

https://claude.ai/code/session_012pdCiT6U5eYGsEdWSDgXZV


jerry73204:
Ejection here is not this PR — `queue` is failing in the merge group for seven PRs (#333, #429, #438, #459, #478, #479, #482) with an identical `cannot update the lock file … --locked` error. Cause and fix in #490: `_setup-common`'s submodule guard tests for a file rather than for the pin, so the self-hosted runner keeps building against play_launch `8fda8d89` while the tree pins `4c214a63`. Not worth re-queuing anything until #490 lands.

jerry73204:
Superseded by #483, which now targets `main` and contains both of these commits rebased past it. The rebased versions are the ones to take: resolving against current `main` required adopting issue 1033's entity-cap minimums into `xrce-config.txt` (this wave moved those minimums out of `build.rs`, so a mechanical resolution would have silently restored the pre-1033 values in the new file). Closing here rather than re-queuing so the same commits are not in the queue twice.

PR #483, opened 2026-09-05
phase-420 W9 step 4 + #1077: one compile for XRCE, a measured refusal for zenoh, and the flake that made four gates lie

jerry73204:
**Phase 420 is complete with this PR.** Retargeted to `main` and now carries everything from #482 as well (rebased past it), so this is the whole of W9 plus issue 1077. Nine commits.

## W9 step 4 — the last step, and it resolved differently for the two vendored trees

Both answers are measurements rather than preferences.

**XRCE — one compile, in the lane that ships.** The vendored client and micro-CDR were compiled twice; step 1 had already measured that the second reaches no shipping image, so the CTests were validating objects nothing contains. They now link the objects images link. Direction was decided, not defaulted: making cargo shell out to cmake would put cmake plus a toolchain file in the path of every cross build, for six target families cc-rs reaches today with neither.

The archive is **located, never globbed** — the `<hash>` in `target/<profile>/build/…-<hash>/out/` is neither stable nor predictable, and taking a glob's first match is issue 0500's defect. The recipe reads `OUT_DIR` out of `cargo build --message-format=json`; the build script writes a pointer file naming the archive and header directory. There is deliberately **no fallback** to compiling here — a fallback would put the duplication back while every symptom said it was gone.

Two things worth reading: the archive **cannot** be split (`platform_aliases.c` defines `uxr_millis`/`uxr_nanos` which the vendored TUs call, while backend TUs call the vendored session API — mutually recursive at link, and rustc emits no `--start-group`), and **no diagnostic was lost**: the CMake `-Wall -Wextra -Wpedantic` never actually ran over the vendored sources, because the same target carried `-Wno-unused-parameter -Wno-pedantic`. Our own `src/*.c` keep their warnings, with both suppressions now removed and compiling clean.

**zenoh — there is no duplicate compile to remove.** Measured: `cargo build -p zpico-sys` compiles 131 vendored `.c`; the same build `--features zephyr` compiles **zero**, as does `nros-c` with the literal feature string `zephyr/CMakeLists.txt` passes. The cargo lane *cannot* compile these TUs on Zephyr — `system/platform/zephyr.h` includes a `version.h` only the west build generates (issue 0541). So a Zephyr image holds zenoh-pico from the west lane alone and every other platform from the cargo lane alone. Not the 0135-class split the scoping feared; no issue filed for it.

"One build consumed by both" would mean feeding a west artifact to cc-rs for **zero** platforms that build twice, in the one direction `version.h` makes impossible. Refused with the evidence rather than built. What the evidence *does* support: the split is load-bearing and nothing checked it — and a Zephyr C/C++ image links with `-Wl,--allow-multiple-definition`, so if both lanes ever did compile the tree, one lane's objects would silently win rather than failing the link. `check-zenoh-lane-ownership` holds its three statements together.

## A rebase that nearly unshipped a fix

Issue **1033** landed on main while this was in review, lowering three entity-cap minimums from 1 to 0 — the change that takes `xrce_session_state_t` from 76,624 to 54,928 bytes on the zephyr cpp listener, under its 66,048 heap, so the image boots. This wave moved those minimums out of `build.rs` into `xrce-config.txt`, so a mechanical conflict resolution would have restored the old `1`s in the new file with nothing to say so.

They are adopted, with 1033's reasoning moved beside them, and the revert is **measurable**: with the minimum back at 1, `NROS_XRCE_MAX_SUBSCRIBERS=0` panics `is below the minimum 1` — `stated()` refuses rather than clamping, which is issue 0827's rule.

## Every gate here was rewritten after review found a hole in it

Four gates, four holes, all the same shape: the author's mutations broke *structure*, while the real risk is a record **wired to the wrong thing**.

- XRCE versions: point micro-CDR's derivation at the client's tree — every shape check passes, the configure emits `MICROCDR_VERSION_STR "3.0.1"`.
- zenoh sources: move `dir core … utils` into a conditional group — every count identical, nine `.c` vanish from every non-Zephyr build.
- XRCE one-compile: hand-spell the archive name instead of reading the pointer file.
- zenoh lane-ownership: swap two groups' conditions — token set unchanged, so the sibling gate cannot see it; `system/zephyr/network.c` would compile only under ISO-TP.

Each now checks attachment, with the invariant stated in words. One agent's own sweep found a fifth (a "FATAL_ERROR within 400 chars" rule that was itself shape-valid and wired to nothing).

## Issue 1077 — the flake, root-caused

`grep -q` exits at the first match; bash's builtin `printf` flushes per line; under `pipefail` the writer's SIGPIPE becomes the pipeline's status, so an assertion reports a string absent from a variable that contains it. **My own filing had eliminated this hypothesis, and was wrong twice** — the probe used a single-line 2 MB haystack when the hazard scales with *lines after the match*, and it read `PIPESTATUS` after `rc=$?`, which resets the array.

Trials: primitive 60/3000 → **0/5000**; the gate at 12 concurrent copies 13/300 → **0/300**. `just check fast` is now **230/230 green**, twice.

## Bookkeeping

W9 checked, status line rewritten, and the doc moved to `docs/roadmap/archived/` with phase-421's link repointed in the same commit — it is the one live doc that names this file by path.

Three issues filed and deliberately left outside the phase: **1096** (a *third* compiler of zenoh-pico in a qemu readiness script, restating the source selection and the config header — its own comments say *"matches build.rs"* four times), **1097** (`just xrce check-rust-rmw` runs `cargo check` in a directory with no `Cargo.toml` since phase-140), **1078** (five XRCE pool knobs sizing ~86 % of `xrce_session_state_t`, absent from the static pool inventory).

https://claude.ai/code/session_012pdCiT6U5eYGsEdWSDgXZV


jerry73204:
Rebased onto current main — 18 commits, none dropped. Five conflicts, all resolved toward main, and in every case main's version was equal or better:

- `tests/cmake-support-library-tests.sh` (5 regions) and `tests/fixture-staleness-probe-tests.sh` (2) — both sides fixed the SAME `grep -q`/SIGPIPE bug the same way (capture-then-herestring), this branch citing #1077 and main citing #1173. Main's staleness-probe version is a strict SUPERSET: it also checks the build's exit status and stops discarding stderr, because without that a permission-denied cargo lock produced no `"fresh":false` and the check reported "this fixture no longer models the defect" — a sentence aimed at the fixture for a fault in the environment.
- `tests/cmake-message-bounds-tests.sh` (5 regions) — `"$_p_out"` vs `"${_p_out}"`. Identical semantics; took main's bracing.
- `tests/cmake-reconfigure-tests.sh` (2 regions) — this branch had bare `grep -q`, main has `nros_grep_q`. Main's is required: a bare `grep -q` cannot distinguish a forked grep's EAGAIN from "no match", which is why 76 sites were converted and `check-grep-q-error-conflation` keeps new ones out.

One conflict needed both sides kept rather than a pick: `just/native.just`, where main ADDED `test-ros2-advertised-state` (phase-433 W6) while this branch DELETED `test-ros2-shell` (issue 1097 — it ran `./tests/ros2-interop.sh`, a script phase-16 planned and never committed, so the recipe never worked on any host). Kept main's new recipe and applied this branch's removal.

Verified: `just --list` clean, no conflict markers anywhere, and `check grep-q-error-conflation`, `just-recipe-paths`, `no-tracked-file-find` all pass.

PR #484, opened 2026-09-05
phase-412: a boot self-report on silicon, and the contract that sizes the pools

jerry73204:
Twenty-two commits from the phase-412 campaign, all found or verified on an MR-CANHUBK344.

## The instrument

The island reached FirstSpin on no run and no channel could say why. The board has one UART and zenoh owns it, so every diagnostic written to explain an arena or entity failure went to a sink that does not exist (issue 1036). `boot_report` is a fixed RAM record -- `#[no_mangle]` statics read back by halting the core, address resolved from the ELF -- because the failures halt before any stream could carry them. `scripts/read-boot-report.py` decodes it and `scripts/check-boot-report-layout.py` keeps the Rust struct, the snapshot and the decoder in agreement.

## Three root causes it found

* An **undeclared eleventh subscription**. `mrm_handler` creates seven and its `ENTITIES` list declared six, so every pool derived from that list was short by one. The eleventh hit the buffer-index guard, returned `SubscriberCreationFailed`, and the `_ =>` arm collapsed it into `-100 Backend("rmw_ret error")`.
* **`CONFIG_MAX_PTHREAD_MUTEX_COUNT=32`.** zenoh-pico takes one pthread mutex per subscriber sync group. Ten subscribers boot at 32; eleven fail at 32 and reach FirstSpin at 64. The configure now fails with the number to write instead of the image failing at boot.
* **The rlsf free path credited nothing.** `used()` was cumulative-allocated: 395,132 bytes reported in a 94,720 byte arena, 417 percent. Charging the usable size on alloc and crediting it before deallocate makes the real peak 75,480, or 79.7 percent. Three tests, two of which fail without the fix.

## The wiring, so the count stops being hand-maintained

A launch file cannot say a node creates seven subscriptions, and `<remap>` only renames a topic the node already has. play_launch supplies a contract sidecar for exactly that, and the resolver folds it into the SystemModel, whose `structure.topics` is what `count_callbacks` counts. `nros ws entity-inventory` gains `--model`, and takes the larger of the model and the declaration per component and per kind -- neither source is complete, because the model has no timer entity and the declaration is what was wrong.

Measured on the island with the recovered subscription deliberately deleted from the declaration:

```
metadata alone      28 entities, 14 callback slots
metadata + model    29 entities, 15 callback slots
```

Two cmake fixes went with it. `nano_ros_add_executable` never parsed `BRINGUP`, so every caller reaching the entry through the ament verb was pushed back onto the `MODEL` phase-405 W4 retired -- and an unparsed keyword does not error, it joins `SOURCES`. And nothing passed `--model` to the inventory, which is a capability nobody could observe.

## Two reds that were on main

Both reproduced on a clean `origin/main` worktree before being fixed here: `nros_platform_api::task` was alloc-gated while the executor calls its stack query on every feature set, and `check-api-parity` had no ledger row for `Executor::set_min_stack_headroom_bytes`.

## Falsified along the way

`transient_local` QoS, `NROS_ZEPHYR_HEAP_SIZE`, `CONFIG_HEAP_MEM_POOL_SIZE`, `Z_BATCH_UNICAST_SIZE`, `Z_RX_CACHE_SIZE` (compiled out), and the payload-class allocator. Each was proposed, instrumented and killed by measurement.

## Still open

Issue 1036 stays open: no cell exercises the boot report on silicon, which is the same shape as the defect it was built to find. And `_z_sync_group_state_create` returns `-1` where the source predicts `-80`; the empirical fix is verified but that attribution is a loose end.

## Verification

`just ci gate` green (`check::cli-fresh`, `check::fast`, `check::build`, `check::api-parity`, `test-unit`, `test-lane-contracts`). `nros_rmw_cyclonedds_ros2_pubsub_e2e` failed twice under the 21-way parallel gate and passed solo and on re-run with none of its inputs touched; recorded as load flake, not fixed.

https://claude.ai/code/session_01EST8f2UDcrw2Ztm5eLTuMC

### DIFF
--- a/.config/doc-recipe-refs-baseline.txt
+++ b/.config/doc-recipe-refs-baseline.txt
@@ -64,13 +64,3 @@ docs/roadmap/phase-292-asi-reference-consumer-revisit.md	zephyr
 docs/roadmap/phase-375-board-tier-policy-and-onboarding-cost.md	board-new
 docs/roadmap/phase-395-dev-workflow-migration.md	ci-l2
 docs/roadmap/phase-395-dev-workflow-migration.md	ci-l4-tier1
-packages/cli/CONTRIBUTING.md	build-build-tools
-packages/cli/CONTRIBUTING.md	build-user-libs
-packages/cli/CONTRIBUTING.md	clean-build-tools
-packages/cli/CONTRIBUTING.md	clean-user-libs
-packages/cli/CONTRIBUTING.md	test-build-tools
-packages/cli/CONTRIBUTING.md	test-user-libs
-packages/cli/testing_workspaces/complex_workspace/README.md	build-interfaces
-packages/cli/testing_workspaces/complex_workspace/README.md	build-rust
-packages/cli/testing_workspaces/complex_workspace/README.md	list
-packages/cli/testing_workspaces/complex_workspace/README.md	run
--- a/cmake/NanoRosEntityInventory.cmake
+++ b/cmake/NanoRosEntityInventory.cmake
@@ -122,8 +122,16 @@
 #   nros_derive_entity_inventory_knobs(
 #       CLI <path to nros>                  # required
 #       [METADATA <nros-metadata.json>]     # default ${CMAKE_BINARY_DIR}/...
+#       [MODEL <system_model.yaml>]         # resolved SystemModel, if any
 #       [OUTPUT_FILE <path>]                # default the knobs file below
 #       [QUIET])
+#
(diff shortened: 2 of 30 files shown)


jerry73204:
Superseded by #510, which is this same work rebased onto current main (the branch here carries a zenoh-pico pin that main has since moved past, so `check-submodule-pins` reads it as a rewind). #510 also retires `ENTITIES` outright, which this PR only prepared for.

Not force-pushed on purpose -- a new branch keeps this history readable rather than replacing it.

PR #485, opened 2026-09-05
fix(#1057): the lane test read the DISPATCHER, not the recipe that runs

jerry73204:
Closes #1057.

`ci_lane::tests::recipes_run_the_scope_their_lane_declares` has been red on main. Reproduced on a clean tree before touching anything:

```
Tier2 declares RunScope::LaneCoords but `just ci matrix` never exports
NROS_TEST_COORDS — the run would resolve every coordinate while the
preflight accepts a `lane=tier2` build
```

## The declaration and the recipe agree — the test was reading the wrong body

phase-410 split each tier's DEPTH into a fixed inner recipe, so `matrix` is now a `case` over `just ci::_matrix-{run,build}` and the export moved into `_matrix-run`. The recipe's own comment says why the split is shaped that way — *"so `check-lane-contracts` can walk each and prove its affordability claim"*. The Python gate was updated to walk them; this Rust test was not.

Nothing was dropped in the split, which was the issue's other hypothesis: `_matrix-run` carries `NROS_TEST_COORDS` on both of its `just` invocations.

## Why it matters more than one red test

`just check fast` runs no unit tests, and `test-unit` excludes `nros-tests` — so nothing pre-merge sees this. It fails where it costs the most, and an ejection from the queue is how anyone finds out.

## The fix

Follows the dispatch rather than hard-coding tier 2's inner name, so the next tier that grows a depth is covered by construction.

Only **module-qualified private** targets (`just <mod>::_<name>`) are chased. A qualified public one (`just check`) is a different lane with its own contract; an unqualified private one (`just _lane-gate`) is a root recipe in a file this test has not loaded. Following either would widen what an assertion about *this* recipe can accidentally read.

A named delegate that does not resolve is a **failure, not a skip** (issue 0196's rule) — silently reading a shorter body is exactly how this assertion would go quiet the next time a recipe is renamed.

Body extraction moved into `recipe_body` so the dispatcher and its delegates are parsed by one function instead of two spellings of the same scan.

## Verification

Mutation-checked three ways:

| mutation | result |
| --- | --- |
| remove every `NROS_TEST_COORDS` export from `_matrix-run` | FAILS |
| rename `_matrix-run` without updating the dispatcher | FAILS |
| unmodified tree | passes |

`just check fast` 217/217, `just test-lane-contracts` 17/17, all 12 `ci_lane` tests pass.

Left deliberately, and written into the archived issue: the assertion is "the body mentions the variable", so removing one of the two exports still passes. That is its existing granularity, not something this change narrowed; tightening it to "every `just` invocation in the body carries it" is a separate change across four tiers.

https://claude.ai/code/session_01JaST26nBgH9mLezz4TWKmP

### DIFF
--- a/docs/issues/archived/1057-tier2-run-scope-declares-lanecoords-but-recipe-exports-none.md
+++ b/docs/issues/archived/1057-tier2-run-scope-declares-lanecoords-but-recipe-exports-none.md
@@ -2,7 +2,7 @@
 id: 1057
 title: "`ci_lane` unit test is RED on main: tier 2 declares `RunScope::LaneCoords`
   and `just ci matrix` no longer exports `NROS_TEST_COORDS`"
-status: open
+status: resolved
 type: bug
 area: ci, testing
 severity: high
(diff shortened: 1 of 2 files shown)


jerry73204:
Superseded — the same defect was fixed independently on main (`ci_lane::recipes_run_the_scope_their_lane_declares` now follows a dispatcher into its private delegates, with a synthetic-justfile test for the extraction itself, which this branch lacked).

One difference is worth keeping and I will send it separately: main's version does `if let Some(inner) = recipe_body(...)`, so a delegate that does not resolve is silently skipped and the assertion reads a shorter body. That is the recurrence path — rename `_matrix-run` without updating the dispatcher and the test goes quiet again. Issue 0196's rule says it should fail.

PR #489, opened 2026-09-05
fix(#1080): the stack probe allocates nothing, so it must not be alloc-gated

jerry73204:
Every Zephyr cortex-m fixture fails to **compile**. Found by a tier-2 fixture
build at `build-cortex-m-c-talker-zenoh`.

```
error[E0433]: cannot find `task` in `nros_platform_api`
  --> packages/core/nros-node/src/executor/spin.rs:2415:41
  note: found an item that was configured out
       #[cfg(feature = "alloc")]
```

## Cause

`cb2be0ca4` added `check_stack_headroom_rule`, calling
`nros_platform_api::task::stack_unused_bytes()` unconditionally. `pub mod task`
is `#[cfg(feature = "alloc")]` — correctly for most of it, since `PlatformTask`
allocates its own stack and join slot.

**But `stack_unused_bytes` allocates nothing** — one `extern "C"` call returning
a `usize`. The gate is about its *neighbours*.

## Why not guard the call site

The obvious patch is `#[cfg(feature = "alloc")]` on the rule. It builds, and it's
wrong: the rule is a **safety** feature, and the targets it would be silently
removed from — `no_std`, no allocator, small embedded — are exactly where a stack
overflow is how one component corrupts another's state and nobody is watching a
console.

## The fix, and the mistake in the first attempt

`stack_unused_bytes` moves to a non-gated `crate::stack`, **and the caller names
that path**.

The first commit only moved it and re-exported from `task` — which fixed nothing,
because `task` is still gated. The tier-2 rebuild returned the identical error at
the identical line. The re-export stays for out-of-tree callers that have `alloc`
anyway.

## The check that finally discriminated

Three did not, and each was reported as verification before being shown otherwise:

| attempt | why it proved nothing |
| --- | --- |
| `cargo check -p nros-platform-api --no-default-features` | passes either way — the defect is in the caller |
| `cargo check -p nros-node --no-default-features` | passes with the fix **stashed** too; workspace resolution gives the API crate its default features regardless |
| `just zephyr build-one c/talker zenoh mps2_an385` | cannot reach the coordinate — `build-one` names its dir from example+rmw, cortex-m rows carry a distinct `west_build_name` |

The size probe records what it actually builds
(`build/sizes-probe/…/nros-probe-key-inputs.txt`): target `thumbv7m-none-eabi`,
features `rmw-cffi,ros-humble`, no defaults.

```
cargo check -p nros --target thumbv7m-none-eabi \
      --no-default-features --features rmw-cffi,ros-humble

  with `task::`   → error[E0433]: cannot find `task` in `nros_platform_api`
  with `stack::`  → Finished
```

Seconds instead of a 20-minute fixture build, **both directions run**, and it
fails for the right reason.

## The family

Second regression from the same stack-instrumentation work, failing at a
different build phase each time:

| commit | | |
|---|---|---|
| `411addfd2` | the platform probe | **link** — issue 1075 (PR #480) |
| `cb2be0ca4` | the executor rule calling it | **compile** — this |

Both invisible to the compile tiers, which build the host configuration where
`alloc` is on and the Zephyr syscall is irrelevant.

### DIFF
--- a/docs/issues/1080-stack-unused-bytes-trapped-behind-alloc-gate.md
+++ b/docs/issues/1080-stack-unused-bytes-trapped-behind-alloc-gate.md
@@ -1,0 +1,102 @@
+---
+id: 1080
+title: "`stack_unused_bytes` allocates nothing but lives in an `alloc`-gated module, so the new stack-headroom rule breaks every `no_std`-without-`alloc` image"
+status: open
+type: bug
+area: platform, core
+severity: high
+found: 2026-09-05
+related: [1075, 0589, phase-359]
+---
+
+# A gate that belongs to the neighbours
+
+## Symptom
+
+Every Zephyr cortex-m fixture fails to compile. Found by a tier-2 fixture build
+on 2026-09-05, at `build-cortex-m-c-talker-zenoh`:
+
+```
+error[E0433]: cannot find `task` in `nros_platform_api`
+    --> packages/core/nros-node/src/executor/spin.rs:2415:41
+     |
+2415 |         let unused = nros_platform_api::task::stack_unused_bytes();
+     |                                         ^^^^ could not find `task` in `nros_platform_api`
+     |
+note: found an item that was configured out
+    --> packages/platform/nros-platform-api/src/lib.rs:65:9
+     |
+  64 | #[cfg(feature = "alloc")]
+     |       ----------------- the item is gated behind the `alloc` feature
+```
+
+It surfaces through the `nros-c` size probe, which runs a nested cargo build, so
+the first line a reader sees is *"size probe could not locate the `nros` rlib"* —
+several frames from the cause.
+
+## Cause
+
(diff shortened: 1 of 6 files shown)


jerry73204:
Closing as superseded — `main` already carries this fix.

It landed independently in `b9450327d` + `92849933d`: `stack_unused_bytes` moved out of the `alloc`-gated `task` module to a top-level `nros_platform_api::stack_unused_bytes()`, with the `spin.rs` call site updated to match. Confirmed present on `main` now.

For the record, this bug drew three independent fixes — this PR, main's two commits, and my own #497 (also now closed). The reason it reached main in the first place is that the lane which catches it, `check::build`, runs only on `schedule`/`workflow_dispatch` and no merge-gating event, so both halves were individually green on the required `CI` context. That is worth deciding about separately; `workspace-all` is ~62 s and `workspace-features` ~435 s measured, so putting them on the PR lane is a real cost tradeoff rather than an obvious win.

PR #494, opened 2026-09-05
fix(#0964): an unbounded type cannot size a buffer — reroute, RX/TX audit, and the arm flipped

jerry73204:
You asked me to do the reroute and audit the RX/TX split. Both are here, and the audit changes the decision you're weighing.

## The lever was connected to nothing

#0964 says the flip is *"the `unbounded` arm of `detail::buffer_bounds` … the one line it turns on"*. That line existed — and **nothing called it**. `buffer_bounds` had no consumer anywhere outside `size_bound.hpp`. Flipping it would have changed nothing.

The estimate actually lived in **15** direct `uint8_t buf[M::SERIALIZED_SIZE_MAX]` declarations: `client` (3), `action_client` (3), `polling_action_server` (3), `action_server` (2), `tick_ctx` (2), `service` (1), `polling_action_client` (1). (24 grep hits, but 8 are doc comments and one is a stub defining its own constant — I'd quoted 24 earlier; 15 is the real number.)

## Audit: all 15 are TX, and none is the #0896 hazard

Every one is `X::ffi_serialize(&value, buf, sizeof(buf), &len)` with the return checked. **No RX buffer is sized by the estimate** — the receive paths already go through `component.hpp`, which passes the strict `rx_size_bound` that poisons on unbounded.

That lowers the stakes of the decision:

- the silent-truncation failure #0896 exists to remove **does not apply here** — these buffers are written by us, bounds-checked, and an over-long message returns non-zero rather than corrupting anything;
- an under-sized estimate on these paths is a **loud runtime failure**. Flipping converts it to a **compile-time** failure — better in kind, but a usability change, not a safety fix.

**The safety argument for the flip can't be made from these sites.** It has to be made about the receive path, which is already strict.

## Not a semantic no-op — I predicted it would be and was wrong

So this is measured on real generated headers rather than assumed. `buffer_bounds` answers `TX_MAX_SERIALIZED_SIZE` for a **derived** type, so those buffers shrink to the bound:

| | bytes |
|---|---:|
| estimate, over the 19 types carrying both constants | 1,928 |
| derived | **202** |
| TX buffers shrink by | **1,726** |

Zero buffers grow. The other 21 generated types state no derived bound, keep the estimate, and are untouched — the `legacy`/`unbounded` arm doing its job.

`action_server.hpp` also gains the `size_bound.hpp` include it had been getting by luck.

## What this leaves you

The flip is now genuinely one line at one control point, and **stageable** where it wasn't: the arm can be turned on for RX before TX, and `has_derived_size_bound<M>` already exists for a per-site opt-in.

The product decision is unchanged in substance — 81 of 120 stock Humble types have no derived bound, so flipping makes code that compiled yesterday stop until each type gains a `cap` or each call site uses the `_sized` form.

`just check cpp`, `just check c`, `just check fast` (213/213) all pass.

### DIFF
--- a/docs/issues/0964-two-different-sizes-for-the-same-type.md
+++ b/docs/issues/0964-two-different-sizes-for-the-same-type.md
@@ -219,3 +219,76 @@ simplest type in the corpus, which is a fair summary of why it was replaced.
 Not measured: an image delta. That still needs the second instrument 0896 named
 — these are stack frames, and the number of live buffers depends on which of the
 13 sites an image instantiates.
+
+
+## The lever was connected to nothing — rerouted, and the RX/TX split audited (2026-09-05)
+
+The open half above says the flip is "the `unbounded` arm of
+`detail::buffer_bounds` … the one line it turns on". That line existed;
+**nothing called it.** `buffer_bounds` had no consumer anywhere outside
+`size_bound.hpp`, so flipping it would have changed nothing at all.
+
+### Where the estimate actually was
+
+| | sites |
+| --- | ---: |
+| `uint8_t buf[M::SERIALIZED_SIZE_MAX]`, bypassing both templates | **15** |
+| strict `rx_size_bound` / `tx_size_bound` (poisons on unbounded) | 2 |
+
+(24 grep hits, but 8 are doc comments and one is a stub type defining its own
+constant. 15 are real buffer declarations.)
+
+Spread over `client` (3), `action_client` (3), `polling_action_server` (3),
+`action_server` (2), `tick_ctx` (2), `service` (1), `polling_action_client` (1).
+
+### The audit answer: all 15 are TX, and none is the 0896 hazard
+
+Every one of the fifteen is
+`X::ffi_serialize(&value, buf, sizeof(buf), &len)` with the return checked.
+**There is no RX buffer sized by the estimate** — the receive paths already go
+through `component.hpp`, which passes `nros::rx_size_bound<M>::value`, the
+STRICT template that poisons on an unbounded type.
+
+That matters for the product decision, and it lowers the stakes:
+
+* the silent-truncation failure 0896 exists to remove **does not apply here**.
+  These buffers are written by us, bounds-checked against `sizeof(buf)`, and an
+  over-long message returns a non-zero code rather than corrupting anything;
+* so an under-sized estimate on these paths is a LOUD runtime failure, not
+  data loss. Flipping the unbounded arm would convert that into a compile-time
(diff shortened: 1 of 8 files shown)


jerry73204:
## Decision applied: the `unbounded` arm now poisons

Per your call — avoid silent truncation, over-long messages fail explicitly, and an **under-size estimate is a compile-time error because the calculation is incorrect and cannot accept the message size the user expects**.

That's the argument the issue couldn't make for itself, and it's the right one. `compute_serialized_size_max` isn't a conservative bound — flat 512 per nested message, default capacity per string — and it over-stated 38 of 39 bounded types and **under-stated one**. An under-state sizes a stack array that can't hold what the program legitimately sends, and nothing at run time repairs a number already baked into `uint8_t buf[N]`.

`detail::buffer_bounds<M, bound_shape::unbounded>` delegates to `strict_bounds`. The **`legacy` arm is deliberately not flipped** — such a type predates the bound marker and never had a derivation to disagree with, so existing consumers keep compiling.

## Cost in-tree: nothing

Measured, not assumed: `compile-check-fixtures` builds **36 rows across 5 builders, zero** `states no serialized-size bound` errors. `check-cpp`, `check-c`, `check fast` (213/213) pass.

**How that number was nearly wrong:** the first run reported zero poison hits with `cxx=0` — the C++ checks hadn't run at all, because the config-header step died on the unrelated zenoh-pico INET6 bug (#488). A zero from a check that didn't execute is not a zero. Re-run with that fix present, it's a real 36-row zero.

## Held by

- `unbounded_buffer_probe.cpp` — expected-failure TU, both directions, must not compile. A `static_assert` can't express "this must not compile".
- Wired into `check-cpp`, **asserted-present first**, because a missing file is also a non-zero `c++` and reads as a pass.
- **Mutation-tested**: reverting the arm to the estimate fails `check-cpp` naming exactly that.

The old pin test is updated rather than deleted — its `Unbounded` column is gone from `instantiate_recv_paths`, its two positive assertions replaced by pointers to the probe. `has_derived_size_bound<Unbounded>` stays asserted: that's the *fact*; the poison is the *policy* on it.

## Migration, and the cost is real

81 of 120 stock Humble types state no derived bound. A user reaching these paths with one must bound the type (`string<=64`, or a `cap` in `nros-codegen.toml` — `heap`/`view` do **not** bound), or call the `_sized` form at that site.

One note for the record: since the RX paths were already strict, what this changes is the 15 TX sites — where the old failure was **loud**, not silent. The decision stands on the correctness of the number, which is exactly how you framed it.

PR #497, opened 2026-09-05
fix(platform): the stack probe needs no allocator, so stop gating it behind one

jerry73204:
`main` does not build for the embedded target:

    packages/core/nros-node/src/executor/spin.rs:2415
    error[E0433]: cannot find `task` in `nros_platform_api`
    note: found an item that was configured out — gated behind the `alloc` feature

Two commits met: cb2be0ca4 added `check_stack_headroom_rule`, which calls
`nros_platform_api::task::stack_unused_bytes()`, and 411addfd2 put the probe
inside `pub mod task` — a module gated on `alloc` because it also holds the
allocate-spawn-join helper. `stack_unused_bytes` is a plain FFI query with no
arguments and no storage; it needs nothing from an allocator. It was only in
there because the spawn helper lives there too.

So the module is no longer alloc-gated, and the half that actually allocates
is gated instead: `PlatformTask` with its `impl` and `Drop`, the four spawn
symbols, `TaskAttr`, `PRIORITY_INHERIT`, and the `c_void` import. The extern
declaration for the probe stands on its own. `check-ffi-struct-mirrors` reads
the source rather than a build, so gating `TaskAttr` does not stop its layout
being checked against the header.

Gating the CALLER instead would have compiled just as well and been wrong:
that disables stack-headroom monitoring precisely on the no-alloc targets,
which are the ones where a stack overflow silently corrupts another
component's state and where `heap_used_bytes` has had a counterpart since
RFC-0034 D7.

Why it sat on main: the lane that catches it is `check::build`, which
`schedule`/`workflow_dispatch` run and no merge-gating event does (CLAUDE.md,
"PR cheap, batch thorough"). Both halves were individually green on the
required `CI` context. Not proposing a lane change here — `workspace-all` is
62 s and `workspace-features` 435 s measured, which is a real per-PR cost and
somebody else's call.

Verified: `check-workspace-all` and `check-workspace-features` both green
(each was red before this); clippy `-D warnings` clean for
`nros-platform-api` under `--no-default-features` and `--features alloc`.

### DIFF
--- a/packages/platform/nros-platform-api/src/lib.rs
+++ b/packages/platform/nros-platform-api/src/lib.rs
@@ -61,7 +61,9 @@
 // an ABI belongs with the ABI.
 #[cfg(feature = "alloc")]
 extern crate alloc;
-#[cfg(feature = "alloc")]
+// Not `alloc`-gated: `task::stack_unused_bytes` is a plain FFI query, and a
+// stack-headroom probe is most load-bearing exactly on the no-alloc targets.
+// Only the allocate-spawn-join half inside needs the feature.
 pub mod task;
 
 use core::ffi::{c_int, c_void};
--- a/packages/platform/nros-platform-api/src/task.rs
+++ b/packages/platform/nros-platform-api/src/task.rs
@@ -23,6 +23,7 @@
 // once here is the cheaper truth.
 #![allow(dead_code)]
 
+#[cfg(feature = "alloc")]
 use core::ffi::c_void;
 
 /// The ABI's task attributes, mirrored for the `extern` declaration below.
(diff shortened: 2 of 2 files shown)


jerry73204:
Superseded — closing as redundant.

`main` fixed this independently while this PR was open, in two commits: `b9450327d` and `92849933d`. Their fix is the better shape: rather than keeping `stack_unused_bytes` inside `task` and gating everything around it (what this PR does), they moved the probe out to a top-level `nros_platform_api::stack_unused_bytes()` and updated the `spin.rs` call site to match. The probe never belonged in the spawn helper's module in the first place.

One correction to my own earlier reading: I measured `check workspace-all` still red against main's version and briefly took that as main's fix being incomplete. That was my error — I had applied only main's `packages/platform/nros-platform-api/` into my tree without its matching `spin.rs` change, so the call site was still pointing at the old path. Main is genuinely fixed.

Nothing here is worth rebasing onto main's version.

PR #501, opened 2026-09-05
test: run the executor suite on alloc, not only on a host

jerry73204:
> Stacked on #491. That PR is the unblock; this is the resting place it should not have been.

## Why

#491 gated the executor tests on `std` to stop `cargo test -p nros-node` failing. Correct as an unblock, wrong as an end state: it left **275 tests reachable only from a host**, on a crate whose entire point is running without one.

Reviewer's objection, and it was right: nano-ros core is designed no_std, so why is the fix to add `std`?

## Almost none of them needed std — the harness did

| what | fix |
|---|---|
| 98 × `std::sync::atomic` | `portable_atomic`, which the crate already depends on and the library already uses for this exact reason |
| `test_clock_us` read `std::time::Instant` | dual-mode: `std` keeps the real clock (existing behaviour untouched); alloc-only gets a deterministic counter |
| `std::sync::Arc`, `std::vec` | `alloc::sync::Arc`, `alloc::vec` — 34 Arc, 15 Vec, 10 `vec!` |

98 tests reach the clock through `executor_with_clock` wanting a *monotonic source*, not real elapsed time. A counter is monotonic, needs no host, and is reproducible — strictly better for those tests than wall time that varies run to run.

## The gate becomes `alloc`, which is what the API actually needs

`from_session`, `spin_blocking` and `halt_flag` live in an `impl` block gated `#[cfg(feature = "alloc")]`. A pure no_std build has a smaller `Executor` and could never have run those tests. `alloc` is the honest requirement; `std` was over-broad.

## 36 tests stay host-only, each with its reason

Marked individually rather than swept under a file-level gate:

- **19** sleep, read wall time, or print
- **10** use `std::sync::Mutex`, which has no no_std equivalent
- **7** need time to advance *independently of how often it is read* — they wait out `RATE_CHECK_INTERVAL_US` (5 s) or a timer period, and a per-read counter cannot close those windows

That last group is worth naming: they did not fail because the code is wrong, but because a read-driven clock cannot express "5 seconds passed while nobody looked".

## Result

```
default (no_std)   172 passed    unchanged
alloc              275 passed    was 0 — the module did not compile
std                321 + 5       unchanged, no regression
```

The alloc column is the point. That configuration is the shape of a target image, and until now none of the executor suite ran against it.

`cargo fmt` clean.

### DIFF
--- a/packages/core/nros-node/src/executor/mod.rs
+++ b/packages/core/nros-node/src/executor/mod.rs
@@ -104,7 +104,7 @@ pub mod action;
 // wall clock. Gating them is not hiding coverage, it is declaring what they
 // already required. `cargo test -p nros-node --features std` is unchanged and
 // remains the way to run them.
-#[cfg(all(test, feature = "std", not(feature = "rmw-cffi")))]
+#[cfg(all(test, feature = "alloc", not(feature = "rmw-cffi")))]
 mod tests;
 
 // Flat re-exports so users write `executor::Executor` etc.
--- a/packages/core/nros-node/src/executor/tests.rs
+++ b/packages/core/nros-node/src/executor/tests.rs
@@ -56,10 +56,31 @@ impl Drop for SimTimeGuard {
     }
 }
 
+#[cfg(feature = "std")]
 fn test_clock_us() -> u64 {
     use std::{sync::OnceLock, time::Instant};
     static EPOCH: OnceLock<Instant> = OnceLock::new();
     EPOCH.get_or_init(Instant::now).elapsed().as_micros() as u64
+}
+
+/// Deterministic stand-in for the host clock on an `alloc`-only build.
+///
+/// The real one above reads `std::time::Instant`, which is what kept this
+/// whole module host-only: 98 tests reach it through `executor_with_clock`
+/// without needing real elapsed time, only a monotonic source. A counter is
+/// monotonic, needs no host, and is reproducible -- which is strictly better
+/// for those tests than wall time that varies run to run.
+///
+/// Tests that genuinely need real time to pass (they sleep, then assert that
+/// time moved) are marked `#[cfg(feature = "std")]` individually and do not
+/// run here.
+#[cfg(not(feature = "std"))]
+fn test_clock_us() -> u64 {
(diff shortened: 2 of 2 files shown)


jerry73204:
Rebased onto current `main`. One conflict, in `packages/core/nros-node/src/executor/tests.rs`, and it was **mechanical** — two independent additions at one shared anchor, with an empty merge base between them:

- `main` (`7195c9e00`, issue 1104) inserted a 43-line `SimTimeGuard` block there, serialising the tests that touch `time_source`'s process-global state;
- this branch inserted a one-line `#[cfg(feature = "std")]` there.

Resolved as both, main's block first. No semantic disagreement — every other hunk on this branch (`std::sync::Arc` → `alloc::sync::Arc`, `portable_atomic`, the per-test `#[cfg(feature = "std")]` attributes) auto-merged. Verified with `cargo test -p nros-node --lib --features std,sim-time,param-services --no-run`, which is the lane `check.just:714` runs: compiles clean.

**One latent thing for you to decide, which I deliberately did not change.**

`SimTimeGuard` came from `main` gated on `#[cfg(feature = "sim-time")]` alone, and it uses `std::sync::Mutex` / `MutexGuard`. `sim-time = ["dep:nros-rosgraph-msgs"]` does not imply `std`, and this branch widens the module gate in `mod.rs` from `feature = "std"` to `feature = "alloc"`. So after this merge the file no longer holds the property the branch is about — *the test module compiles on alloc* — even though nothing breaks today, because the only lane that builds it passes `std,sim-time,param-services` and `check-no-std` uses a bare `cargo check` that never builds `cfg(test)`.

Everywhere else in this file the convention is explicit:

```rust
#[cfg(feature = "std")] // std::sync::Mutex has no no_std equivalent
```

If you want the alloc property to actually hold, `SimTimeGuard`'s three `#[cfg(feature = "sim-time")]` attributes want `feature = "std"` added, and then `ros_time_timer_follows_the_simulated_clock` (which today carries only `#[test]`, on `main` as well as here) needs the same gate. That changes what compiles where, so it is your call rather than something to fold into a rebase.

PR #506, opened 2026-09-05
fix(#1093): bump zenoh-pico — v4-only lwIP has no INET6_ADDRSTRLEN

jerry73204:
Every FreeRTOS fixture failed inside the vendored zenoh-pico:

```
platform.c:91:13: error: 'INET6_ADDRSTRLEN' undeclared (first use in this
function); did you mean 'INET_ADDRSTRLEN'?
```

## Cause

`platform.c:74` compiles both endpoint helpers for one arm covering
WINDOWS/LINUX/MACOS/BSD **and** `ZENOH_FREERTOS_LWIP`. `INET_ADDRSTRLEN` is fine
— lwIP always defines it — but **`INET6_ADDRSTRLEN` exists only under
`LWIP_IPV6`**, and nano-ros ships `LWIP_IPV6 0`
(`nros-board-freertos/config/lwipopts.h:45`).

Upstream's arm treats *"lwIP"* as implying IPv6. Ours is a v4-only build, which
lwIP supports.

The helper arrived with upstream `04c6b1c7` (*"Connectivity events and API"*,
#1159) on the **1.8.0 bump — phase-415**, 2026-09-04.

## Why it took a day to surface

FreeRTOS fixtures need a fixture build, and Zephyr builds ahead of them in the
same lane. Zephyr had been failing for three unrelated reasons — **1075** (link),
**1080** (compile), **1082** (a stale Kconfig value carried forward in a reused
build dir) — so the lane never reached FreeRTOS.

**Four failures stacked in one lane, each hiding the next.**

## The fork patch

`66556c99` on the `nano-ros` branch, beside the existing lwIP include:

```c
#ifndef INET6_ADDRSTRLEN
#define INET6_ADDRSTRLEN 46
#endif
```

**Chosen over excising the helper from the lwIP arm.** On a v4-only stack
`inet_ntop(AF_INET6, …)` returns NULL, which the function *already* treats as
`_Z_ERR_GENERIC` — so the fallback keeps that honest runtime answer, needs no
call-site surgery, and widens the fork delta by three lines instead of touching
every caller.

## Pin move

`49b4e6356 → 66556c991`, **forward-only, verified by `merge-base --is-ancestor`
against main's pin** rather than by eye — two hex strings no reviewer can order
by sight (the 2026-08-15 rewind is the precedent).

Branch pushed **first**, then the pointer bumped, per the vendored-fork order.
History is linear: the patch was a fast-forward on the branch tip `dd071b8d`, no
rebase needed.

The other commit in the range (`dd071b8d`, *"name the capability the randomness
arms need"*) was already on the branch and published; this bump carries it too.

### DIFF
--- a/docs/issues/archived/1093-zenoh-pico-1-8-0-ipv6-helper-breaks-lwip-build.md
+++ b/docs/issues/archived/1093-zenoh-pico-1-8-0-ipv6-helper-breaks-lwip-build.md
@@ -1,0 +1,82 @@
+---
+id: 1093
+title: "zenoh-pico 1.8.0 compiles an IPv6 endpoint helper for the lwIP arm, so every FreeRTOS fixture fails on an undeclared `INET6_ADDRSTRLEN`"
+status: resolved
+type: bug
+area: build, rmw
+severity: high
+found: 2026-09-05
+related: [1075, 1080, 1082, phase-415, 0507]
+---
+
+# Upstream assumed every lwIP build has IPv6
+
+## Symptom
+
+Every FreeRTOS fixture fails in the vendored zenoh-pico:
+
+```
+zenoh-pico/src/system/common/platform.c:91:13: error: 'INET6_ADDRSTRLEN'
+undeclared (first use in this function); did you mean 'INET_ADDRSTRLEN'?
+error: failed to run custom build command for `zpico-sys v0.5.0`
+```
+
+## Cause
+
+`platform.c:74` compiles both endpoint helpers for one arm:
+
+```c
+#if defined(ZENOH_WINDOWS) || defined(ZENOH_LINUX) || defined(ZENOH_MACOS) || \
+    defined(ZENOH_BSD) || defined(ZENOH_FREERTOS_LWIP)
+```
+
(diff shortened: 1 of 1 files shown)


jerry73204:
Closing: this PR ships nothing.

The FreeRTOS build failure it was filed for (`INET6_ADDRSTRLEN` undeclared on v4-only lwIP) was **already fixed on main** by `49b4e6356`, pinned in `c866bfad8`. The sweep that hit it ran on `tier2-run`, which pinned `c58531578` — an *ancestor* of that fix. `check-submodule-pins` reported the rewind by name; I read the report as a conflict with the new patch rather than as the diagnosis.

The fork commit this PR pinned (`66556c99`) was a strict subset of `49b4e6356`'s guard — same `#ifndef INET6_ADDRSTRLEN / #define … 46`, but nested inside the `ZENOH_FREERTOS_LWIP` block, where the existing one is unconditional and therefore also covers every other platform that gets neither `<arpa/inet.h>` nor an IPv6-enabled lwIP. It is reverted on the fork's `nano-ros` branch (`97557b00`), whose tree is now byte-identical to `dd071b8d` — which is what main pins today.

No issue filed: `check-submodule-pins` did its job.

PR #510, opened 2026-09-05
phase-412: retire ENTITIES; a contract beside the launch file sizes the image

jerry73204:
Supersedes #484, which is the same work on a branch that has since been rebased twice onto main. Twenty-four commits, all found or verified on an MR-CANHUBK344.

## The instrument

The island reached FirstSpin on no run and no channel could say why. The board has one usable UART and zenoh owns it, so every diagnostic written to explain an arena or entity failure went to a sink that does not exist (issue 1036). `boot_report` is a fixed 80-byte RAM record, read by halting the core, because the failures halt before any stream could carry them.

## Three root causes it found

* An **undeclared eleventh subscription**. `mrm_handler` creates seven and its `ENTITIES` list declared six. The eleventh hit the buffer-index guard, returned `SubscriberCreationFailed`, and the funnel `_ =>` arm collapsed it into `-100 Backend("rmw_ret error")`.
* **`CONFIG_MAX_PTHREAD_MUTEX_COUNT=32`.** zenoh-pico takes one pthread mutex per subscriber sync group. Ten subscribers boot at 32; eleven fail at 32 and reach FirstSpin at 64. The configure now fails with the number to write.
* **The rlsf free path credited nothing.** `used()` was cumulative-allocated: 395,132 bytes in a 94,720 byte arena, 417 percent. Real peak after the fix, 75,480.

## ENTITIES is retired

`nano_ros_node_register(ENTITIES ...)` is gone. The keyword stays parsed and is a FATAL_ERROR naming its replacement, because an unparsed cmake keyword does not error, it joins SOURCES.

The reasoning that put it there still holds: components wire themselves in constructors at runtime, so emitted evidence can verify a count but never supply one. What was wrong was WHO states it and WHERE. A list in a component CMakeLists is hand-maintained beside the code with nothing comparing the two, and on the island it drifted by exactly one subscription.

The replacement is a contract sidecar beside the launch file, folded into the SystemModel by the resolver. `EntityInventory::from_model` now reads every kind from it:

| kind | model location |
| --- | --- |
| publisher, subscription | `structure.topics[*].{publishers,subscribers}` |
| service server / client | `structure.services[*].{server,client}` |
| action server / client | `structure.actions[*].{server,client}` |
| timer | `contracts.node_paths[*]` with an EMPTY `input` |

The timer row is what made the retirement possible. "The model has no timer entity" was written into three places in this repo and was the stated reason ENTITIES had to survive. It is false: `PathContract::input` is documented as "empty = periodic (timer-driven)", and a contract `trigger: { timer: { rate_hz: N } }` resolves to exactly that.

## Measured

Island, all four declarations deleted: 29 entities, 15 callback slots, MAX_SUBSCRIBERS 11, RMW_SUBSCRIBER_SLOTS 11 -- identical to what the declarations produced, and 11 is the count the hardware measured. cpp workspace examples likewise identical: talker+listener 3 entities / 2 slots, action_server 2 entities / 2 slots.

On silicon, with a serial router attached: stage 6 FirstSpin, arena 17704/40960 (43.2%), 17 allocations. The last known-good pre-change image reaches stage 3 on the same rig.

## Two cmake gaps closed on the way

`nano_ros_add_executable` never parsed `BRINGUP`, so every caller reaching the entry through the ament verb was pushed back onto the `MODEL` phase-405 W4 retired. And nothing passed `--model` to the inventory verb, so its ability to read a contract had no caller.

## Three reds that were on main

Each reproduced on a clean `origin/main` worktree before being fixed here:

* `nros_platform_api::task` was alloc-gated while the executor calls its stack query on every feature set (also fixed upstream by phase-425 while this branch was apart; that commit is dropped).
* `check-api-parity` had no ledger row for `Executor::set_min_stack_headroom_bytes`.
* `reconcile_ros_time_source` wrote the process-global sim-time gate from an executor that had never heard of `use_sim_time`, switching off a clock another executor installed. Deterministic: 3 runs of 3 red in the full `nros-node` binary, green alone.

## Known limits, both filed

* **Issue 1083** -- a service whose server lives outside the image cannot be declared. Topics have an `external:` mark, services do not, and `dangling-entity` is an ERROR rather than a warning for services. `service_client_pkg` is half of a deliberately-split pair, so it authors no contract and keeps its configured sizes.
* **Issue 1084** -- the declared-QoS-depth static_assert survives its producer. The renderer, the header pickup and the compile test all still work; nothing writes a table. No image in the tree ever declared a depth.

Issue 1036 also stays open: no cell exercises the boot report on silicon.

## Verification

`just ci gate` green. `nros_rmw_cyclonedds_ros2_pubsub_e2e` failed under the 21-way parallel gate on three separate runs, each time in a different sub-case, and passed solo every time; recorded as load flake, not fixed. `just format` fails on a clean `origin/main` checkout too (`examples/workspaces/features/src/rust_param_talker_pkg` is in neither `workspace.members` nor `workspace.exclude`); untouched here.

https://claude.ai/code/session_01EST8f2UDcrw2Ztm5eLTuMC


jerry73204:
Issue 1083 is fixed and folded in, so the "one image cannot move" caveat in the description no longer stands.

`service_client_pkg` was the tenth of ten declarations and the only one that could not become a contract: a service with a client and no server is a hard `dangling-entity` error, and unlike a topic there was no `external:` mark to excuse it. Fixed upstream rather than worked around here:

* **ros-launch-manifest `c66c534`, tagged `v0.1.23`** — `external: server | client | both` on `ServiceDecl` and `ActionDecl`. A separate `ExternalEndpointSide` rather than reusing `ExternalSide`: a topic's sides are `pub`/`sub` and a service's are `server`/`client`, so one enum would accept `external: pub` on a service, where it names nothing. That spelling is now a parse error listing the legal sides, with a test, because it is the mistake someone who just wrote a `topics:` block will make.
* **play_launch `db4af878`** — `ManifestIndex::endpoint_externals` and the cross-scope check. Its own map, not a merge into `externals`: a service and a topic can share an FQN.

The rule keys on WHICH side, so `external: client` does not excuse a missing server. Negative controls on both sides of the boundary, and two on the resolve path end-to-end (mark removed, mark set to `client` — both still refuse).

Every cpp workspace image now derives exactly what its retired `ENTITIES` line said:

```
system           3 entities, 2 cbs   pub 1, sub 1, timer 1
service_server   1 entity,   1 cb    service_server 1
service_client   2 entities, 2 cbs   service_client 1, timer 1
action_server    2 entities, 2 cbs   action_server 1, timer 1
action_client    2 entities, 2 cbs   action_client 1, timer 1
```

Ten of ten. No image in this tree or on the island carries a declaration any more. Issue 1083 is resolved and archived; `just ci gate` green.

PR #529, opened 2026-09-05
feat(cpp): expose the stack-headroom bound to C++ entries, and give it a caller

jerry73204:
> **Scope grew after review.** The setter as originally opened had **no caller**, so `stack-headroom-runtime` stayed off on every image even though the rule and the platform accessor both existed. A slot nothing calls is the "a slot exists, therefore it works" overstatement issue 0800 measured. Second commit adds the caller.

## 1 — the setter (original)

`nros_cpp_executor_set_min_stack_headroom` exposes `Executor::set_min_stack_headroom_bytes` to C++ entries.

## 2 — the caller

`check_stack_headroom` says why the bound must be a **declaration**:

> *"the executor never sees `stack_bytes` — it lives in the spawn attr and goes no further"*

and FreeRTOS reports a high-water mark without the size it was taken against, so a percentage isn't computable there even in principle.

The executor can't. The **board** can — it just created the thread. So the policy lives once, in Rust beside the setter (`derive_min_stack_headroom`, 1/8 of the stack), exported and called by both board entries, rather than the same fraction spelt twice in two C files that would drift. Keeping it in Rust also made the fraction unit-testable, which the C version wouldn't have been.

Called **before** the tier's `setup()`, so an explicit setter call inside setup overrides it — declared beats derived by ordering alone, no precedence logic. Same shape as `set_spin_nominal_us`.

## Three corrections found while wiring it

Each the same mistake: deriving from a number that isn't the real stack.

| | |
|---|---|
| **The boot tier isn't spawned with `stack_bytes`** | It runs on Zephyr's `main()` thread / FreeRTOS's `app_task`. Zephyr can answer (`CONFIG_MAIN_STACK_SIZE`, new shim accessor) and derives; FreeRTOS can't, so its boot tier stays underived rather than hardcoding a number `startup.c` owns. |
| **`ctx->stack_bytes` was never assigned on Zephyr** | The field existed; nothing set it. |
| **The Zephyr shim ignores `stack_bytes`** | Every tier thread gets the fixed `NROS_ZEPHYR_TIER_STACK_SIZE` pool slot; a declared size only earns a warning when it exceeds the slot. `nros_zephyr_tier_stack_size()` now exposes what the thread actually got. |

All three call sites **fail loud** when the bound doesn't take — a monitor that silently fails to arm is indistinguishable from a healthy system, which is the argument `freertos_announce_spawn_failure` already makes one function up.

## Verification — which needed its own correction

The first three FVP runs linked a **stale object**: the build reused a `.obj` six minutes older than the source, so none of them contained this change. `strings` on the ELF found zero copies of the diagnostic. Any conclusion from those runs was worthless.

On a clean build, with the divisor temporarily set to 1 so the bound is the whole stack:

```
nros: DEBUG headroom derive main_stack=524288 rc=0
contract violation: stack-headroom-runtime stack measured=428320 declared=524288
```

`rc=0` — handle valid, bound set. And the rule **fires**. At the shipped 1/8 it's silent (65536 bound vs 428320 unused, ~6.5× headroom) — and that silence now means something, which a quiet run alone could never have established.

`cargo test -p nros-cpp --features std,rmw-cffi`: 19 passed, 0 failed (4 new, RED first). `cargo fmt --check` clean.

### DIFF
--- a/docs/issues/1232-zephyr-tier-stack-bytes-is-ignored.md
+++ b/docs/issues/1232-zephyr-tier-stack-bytes-is-ignored.md
@@ -1,0 +1,91 @@
+---
+id: 1232
+title: "A tier's declared `stack_bytes` does nothing on Zephyr — every tier thread gets the fixed pool slot"
+status: open
+type: bug
+area: zephyr
+related: [phase-436]
+---
+
+## Problem
+
+`stack_bytes` is a contract field. It rides the whole way from `system.toml`
+through the resolver, `NativeTierSpecC` and the generated entry into
+`nros_zephyr_tier_task_create` — and is then **not used to size the stack**.
+
+```c
+k_tid_t tid = k_thread_create(&nros_tier_threads[idx], nros_tier_stacks[idx],
+                              NROS_ZEPHYR_TIER_STACK_SIZE, nros_zephyr_tier_trampoline,
+                              (void*)entry, arg, NULL, (int)priority, 0, start_delay);
+```
+(`zephyr/nros_platform_zephyr_shims.c:696`)
+
+The stacks are a compile-time pool:
+
+```c
+#ifndef NROS_ZEPHYR_TIER_STACK_SIZE
+#define NROS_ZEPHYR_TIER_STACK_SIZE 16384
+K_THREAD_STACK_ARRAY_DEFINE(nros_tier_stacks, NROS_ZEPHYR_MAX_TIERS, NROS_ZEPHYR_TIER_STACK_SIZE);
+```
+(`shims.c:639`)
+
+So **every** tier thread gets `NROS_ZEPHYR_TIER_STACK_SIZE`, whatever the tier
+declared. `stack_bytes` is read once, only to decide whether to print a
+warning:
+
+```c
+if (stack_bytes > (size_t) NROS_ZEPHYR_TIER_STACK_SIZE) {
+    printk("nros: tier stack request %u > fixed slot %u tier=`%s` — running with the "
+           "slot; raise NROS_ZEPHYR_TIER_STACK_SIZE\n", ...);
+}
+```
+
+## Why this is worse than a missing feature
+
+A tier that declares **less** than the slot gets silently more, which is
+merely wasteful. A tier that declares **more** gets a printk and runs
+undersized — and that warning is the only signal, on a lane where a stack
+overflow is the classic spatial-freedom-from-interference failure ISO 26262
+asks about.
+
+Worse, the number is *believed* elsewhere. Anything deriving from
+`stack_bytes` on Zephyr computes against a stack that does not exist. Phase-436
+hit this directly: the first version of the stack-headroom derive used
+`stack_bytes` and had to be changed to ask `nros_zephyr_tier_stack_size()` for
+what the thread actually got. That workaround is in place, but it works
+*around* the contract rather than fixing it.
+
+Two neighbouring defects found in the same pass, both fixed there:
+
+* `ctx->stack_bytes` in `zephyr_run_tiers.c` was **never assigned** — the
+  field existed and nothing set it.
+* The boot tier's `stack_bytes` describes no thread at all: it runs on the
+  Zephyr `main()` thread, sized by `CONFIG_MAIN_STACK_SIZE`.
+
+## Contrast: FreeRTOS honours it
+
+```c
+uint32_t stack_words = (t->stack_bytes > 0u) ? (uint32_t)(t->stack_bytes / 4u) : (262144u / 4u);
+```
+(`freertos_run_tiers.c`) — declared size honoured, with a documented default.
+So the same contract field means two different things depending on the board,
+and only one of them is what it says.
+
+## Fix direction
+
+Three options, roughly increasing cost:
+
+1. **Say so.** Document `stack_bytes` as advisory-on-Zephyr in the contract and
+   in `NativeTierSpecC`, and make the oversize case louder than a printk. Cheap
+   and honest, but leaves the field lying by default.
+2. **Refuse.** Fail the build (or the resolve) when a Zephyr tier declares a
+   `stack_bytes` the pool cannot provide, rather than warning at runtime and
+   continuing undersized. Matches the repo's "refuse rather than degrade"
+   discipline (issue 0709).
+3. **Honour it.** Size the pool slots from the resolved tiers at build time —
+   the toolchain holds every tier's `stack_bytes` before the image is built, so
+   `NROS_ZEPHYR_TIER_STACK_SIZE` could be generated as the max, or per-tier
+   stacks emitted. This is the only option under which the contract is true.
+
+(2) is the minimum that stops the field misleading. (3) is what the field
+already promises.
--- a/packages/api/nros-cpp/include/nros/nros_cpp_ffi.h
+++ b/packages/api/nros-cpp/include/nros/nros_cpp_ffi.h
@@ -1400,7 +1400,24 @@ nros_cpp_ret_t nros_cpp_bind_group_sched(void *handle,
  * `node_name`, `from`, `to` must be valid null-terminated UTF-8 strings.
  * `node_namespace` may be NULL (defaults to `"/"`), otherwise must be a valid
  * null-terminated UTF-8 string.
+ * Declare the minimum stack headroom this executor's thread must keep, in
+ * bytes. `0` (the default) leaves the `stack-headroom-runtime` rule off.
+ *
+ * Exists because the bound cannot be derived from anything the executor can
+ * see. It never receives `stack_bytes` -- that lives in the spawn attr and
+ * goes no further -- and no portable query returns a task's total stack, so
+ * neither an absolute floor nor a percentage can be inferred. The entry that
+ * spawned the thread is the only party that knows what it handed over, and
+ * on a C++ image that entry is on this side of the ABI.
+ *
+ * Call from a tier's `setup(executor)`, beside the other declarations the
+ * generated entry already makes there.
+ *
+ * # Safety
+ * `handle` must be a live executor handle from this ABI, or NULL.
  */
+nros_cpp_ret_t nros_cpp_executor_set_min_stack_headroom(void *handle, size_t bytes);
+
 nros_cpp_ret_t nros_cpp_declare_remap(void *handle,
                                       const char *node_name,
                                       const char *node_namespace,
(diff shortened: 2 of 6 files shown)


jerry73204:
Rebased onto current `main` — this branch was **429 commits behind**, and that is the whole defect.

GitHub reported it `MERGEABLE / CLEAN` because there is no TEXTUAL conflict, so nothing surfaced it. But the merge queue builds the PR merged into `main`, and a branch this old merges into a tree its C and Rust no longer agree with. Its two commits are fine; they were written against a `main` from 2026-09-05.

That is why the failures looked like defects in this PR and were not reproducible against it solo:

    'nros_freertos_tier_ctx_t' has no member named 'stack_bytes'    (queue.yml L3)
    cannot find value `STACK_HEADROOM_DIVISOR`                     (gate, 4 unrelated PRs)
    cannot find function `cpp_ctx_checked`

After the rebase both symbols resolve — `STACK_HEADROOM_DIVISOR` is at `packages/api/nros-cpp/src/lib.rs:3201` — and the shipped shape the queue actually compiles is clean:

    cargo check -p nros-c -p nros-cpp --no-default-features \
        --features "std,rmw-cffi,platform-posix,ros-humble"      exit 0
    cargo +nightly fmt --check --all                              clean

Both commits survived the rebase (2 before, 2 after), no conflicts.

Worth flagging for whoever owns this: a lane triage found this PR failing the required `gate` check for **four other queued PRs** (#719, #648, #206, #733) with the same errors, because they batch together in the merge group. `just queue-triage` correctly classified those as "not your PR". If that is still the case after this rebase, the batch should clear.

jerry73204:
**Correction to my previous comment — the rebase did NOT land.** The push was rejected:

    remote: A pull request for this branch has been added to a merge queue.
            Branches that are queued for merging cannot be updated.
    ! [remote rejected] (protected branch hook declined)

So the remote branch is unchanged and still 429 commits behind. Everything I reported about the rebase was measured in a local worktree only: the clean 2-of-2 replay, `cargo check -p nros-c -p nros-cpp --no-default-features --features "std,rmw-cffi,platform-posix,ros-humble"` exiting 0, and `cargo +nightly fmt --check --all` clean. Those results stand as evidence that a rebase FIXES the reported errors; they are not the state of this branch right now.

Landing it needs the PR dequeued first, which is the author's call — I have not dequeued it.

One thing this changes about the diagnosis, and I would rather flag it than leave it implied: the queue merge method here is REBASE, so the merge-group tree is already this branch rebased onto main. That the errors appeared THERE while a local rebase compiles clean means staleness alone may not be the whole story — the merge group also batches several PRs together, so the failure may need that combination. Worth reproducing against the actual batch before concluding.

jerry73204:
Added **issue 1232** — the finding behind this PR's third correction, filed rather than left as a code comment.

`stack_bytes` travels the whole way from `system.toml` through the resolver, `NativeTierSpecC` and the generated entry into `nros_zephyr_tier_task_create` — and is then **not used to size the stack**. The pool is `K_THREAD_STACK_ARRAY_DEFINE` at a compile-time `NROS_ZEPHYR_TIER_STACK_SIZE`, so every tier thread gets the slot whatever it declared; the declared number is read once, only to decide whether to print a warning.

A tier declaring *less* silently gets more. One declaring *more* gets a printk and **runs undersized** — and that warning is the only signal, on the lane where stack overflow is the spatial-freedom-from-interference failure ISO 26262 asks about.

FreeRTOS honours the same field, so one contract field means two different things by board and only one of them is what it says.

This PR's derive asks `nros_zephyr_tier_stack_size()` for what the thread actually got, which works *around* the contract rather than fixing it. Both that call site and the shim now cross-reference 1232, so the workaround can follow the field once the field means something.

jerry73204:
Heads-up from the phase-426 landing session: this pull request was ejected from the merge queue, and every entry queued behind it inherited the failure — the batch branches `pr-766-…`, `pr-722-…`, `pr-763-…` and `pr-738-…` all failed with the identical error, because a merge-queue batch entry contains every entry ahead of it. `just queue-triage` reads that as "the same check failed for several unrelated pull requests, so it is not yours"; here it was one poisoned batch head rather than shared infrastructure, so the heuristic pointed away from the cause.

The failure is `just test-unit`, in the merge group only:

```
error[E0425]: cannot find value `STACK_HEADROOM_DIVISOR` in this scope
    --> packages/api/nros-cpp/src/lib.rs:3222:19
note: found an item that was configured out
    --> packages/api/nros-cpp/src/lib.rs:3201:7
3181 | #[cfg(feature = "rmw-cffi")]
3201 | const STACK_HEADROOM_DIVISOR: usize = 8;

error[E0425]: cannot find function `cpp_ctx_checked` in this scope
    --> packages/api/nros-cpp/src/lib.rs:3252:31
    --> packages/api/nros-cpp/src/lib.rs:3270:31
```

On this branch's own head the same shape sits at `packages/api/nros-cpp/src/lib.rs:2845`. A `#[cfg(feature = "rmw-cffi")]` attribute is followed by the doc comment for a different item, so the gate lands on `const STACK_HEADROOM_DIVISOR` while `derive_min_stack_headroom` — which reads it — is left ungated, along with two later `cpp_ctx_checked` call sites. Without the feature the definitions are configured out and the users are not, so `cargo test --no-run --workspace` (no features) cannot compile `nros-cpp`.

Why the pull-request checks stayed green: `test-unit` is not in the `pull_request` required set. CLAUDE.md's "PR cheap, batch thorough" split runs it on `merge_group`, so a `--workspace`-with-no-features compile break is invisible until the queue builds the batch — which is exactly the feedback path that entry describes.

Reproduce locally with `just ci gate`, or directly:

```
cargo test --no-run --workspace --exclude nros-tests --exclude nros-rmw-xrce-cffi \
  --exclude nros-rmw-xrce-cffi-staticlib --exclude nros-build-paths \
  --profile nros-relwithdebinfo --locked
```

Nothing here is a judgement on the change itself — the gate is one misplaced attribute. Re-queue once `cargo test --no-run --workspace` compiles, since re-queuing before then burns a batch slot and ejects everything behind it again.

jerry73204:
Re-queued and ejected a second time, on the same cause. The head is unchanged at the line that fails: `#[cfg(feature = "rmw-cffi")]` at `packages/api/nros-cpp/src/lib.rs:2825` is followed by the doc comment for a different item, so the attribute lands on `const STACK_HEADROOM_DIVISOR` (2845) while `derive_min_stack_headroom` (2865), which reads it, is left ungated. Without the feature the definition is configured out and the use is not.

That reads like an orphaned attribute rather than a deliberate gate — the doc block under it ("Declare the minimum stack headroom this executor's thread must keep, in bytes") describes an `extern "C"` setter, not a `usize` constant, so the two look as though they came apart during an edit.

It is invisible to the pull-request checks by design: `test-unit` runs on `merge_group`, not on `pull_request`, so a `--workspace`-with-no-features compile break only shows once the queue builds the batch. `just ci gate` runs it locally, or directly:

```
cargo test --no-run --workspace --exclude nros-tests --exclude nros-rmw-xrce-cffi \
  --exclude nros-rmw-xrce-cffi-staticlib --exclude nros-build-paths \
  --profile nros-relwithdebinfo --locked
```

Each re-queue costs a batch and ejects the entries behind it — this round took #641, #734, #780 and #781 with it before the queue re-formed without this entry. Worth getting that one command green before arming auto-merge again.

jerry73204:
**Pushed a fix for the merge-queue blocker on this branch, and rebased onto current `main`.** Details, because the diagnosis is a little different from what was reported earlier on this PR.

## The failure

`nros-cpp` does not compile without `rmw-cffi`, which takes `test-unit` with it, which takes the whole merge-queue **batch** with it:

```
cargo check -p nros-cpp --no-default-features --features "std,platform-posix,ros-humble"

error[E0425]: cannot find value `STACK_HEADROOM_DIVISOR` in this scope
error[E0425]: cannot find function `cpp_ctx_checked` in this scope   (x2)
```

**Three errors, not one.** The earlier comments on this PR named only the first.

## Cause: one misplacement, four consequences

`#[cfg(feature = "rmw-cffi")]` belonged to `nros_cpp_declare_remap`, sitting above that function's `# Safety` doc. This branch inserted the stack-headroom work **between the attribute and its function**, so:

1. the attribute landed on `const STACK_HEADROOM_DIVISOR`, which `derive_min_stack_headroom` reads while itself ungated;
2. `nros_cpp_executor_set_min_stack_headroom` and `nros_cpp_executor_derive_min_stack_headroom` call `cpp_ctx_checked`, which *is* `rmw-cffi`-only, while themselves ungated;
3. **`nros_cpp_declare_remap` silently lost its gate** — a fourth defect none of the three errors names, because in the failing configuration the missing `cpp_ctx_checked` reports first;
4. `set_min_stack_headroom`'s doc comment was stranded above the const and merged into the const's doc, leaving the function with a bare `# Safety`.

Fixed by putting each attribute on the item it is about, and returning the stranded doc to the function it documents.

`STACK_HEADROOM_DIVISOR` and `derive_min_stack_headroom` stay **ungated** deliberately: pure arithmetic, no context, and the four `stack_headroom_derivation_tests` cases exercise them in a build with no `rmw-cffi`. Gating them would trade a compile error for a silently skipped test.

## Verified both directions

A fix that only satisfies the failing configuration can break the working one:

```
--features "std,platform-posix,ros-humble"           Finished
--features "std,rmw-cffi,platform-posix,ros-humble"  Finished
cargo test -p nros-cpp --lib (no rmw-cffi)           19 passed; 0 failed
```

`just regen-c-headers` was required and is folded in. Its diff is **13 lines, doc-comment only, no symbol change** — worth stating, since regating an exported function is exactly the kind of edit that could have moved the ABI, and it did not.

## Also rebased onto `main`

The branch was four days stale and its own pre-push gates were red for inherited reasons — `c-fmt` (`examples/zephyr/c/action-server/src/FibonacciServer.c`), `cpp-fmt` (`tick_ctx.hpp`), `submodule-pins` (a pin rewind), `cbindgen-headers`. All four clear on current `main`; none were in files this branch touches.

Three gates still fail in a fresh worktree and none is from this branch: `capability-conditionals` and `xrce-vendored-versions` need submodules/vendored trees a fresh worktree has not provisioned, and `doc-commit-citations` is `AGENTS.md:535`, red on `main` and fixed in #790.

## Why this sat for days

`test-unit` runs on `merge_group`, not on `pull_request` — CLAUDE.md's "PR cheap, batch thorough" — so a green `CI` on this pull request never compiled the failing configuration. It surfaces only in the batch, where a failure reads as N unrelated pull requests failing. It took two batches, ejecting #766/#722/#763/#738 and then #641/#734/#780/#781, before the head was identified.

Worth a follow-up: `just queue-triage` reaches "NOT YOUR PULL REQUEST — shared infra" here. That is the right verdict for the ejected entries and the wrong reason, and it points away from the one fixable cause.

PR #536, opened 2026-09-05
fix(#0965, phase-392 W5.g): measured on configured Zephyr images, which existed all along

jerry73204:
Both of #0965's remaining items **and** W5.g's measurement, read off real CMake configures instead of the derivation verb. The images were already in the tree — `zephyr-workspace/build-cpp-*`, Zephyr configures through the real seam.

## 1. "Not configured or flashed" — it's configured, six times

Six report `NROS_ENTITY_INVENTORY_STATUS "derived"`:

| image | `MAX_CBS` | `ACTION_CLIENTS` | `MAX_SUBSCRIBERS` | `MAX_PUBLISHERS` |
|---|---:|---:|---:|---:|
| `cpp-action-client-xrce` | 1 | 1 | 1 | 0 |
| `cpp-action-server-xrce` | 2 | 1 | 0 | 2 |
| `cpp-listener-xrce` | 1 | 0 | 1 | 0 |
| `cpp-service-client-xrce` | 2 | 0 | 0 | 0 |
| `cpp-service-server-xrce` | 1 | 0 | 0 | 0 |
| `cpp-talker-xrce` | 1 | 0 | 0 | 1 |

Each **declares** — the listener's metadata reads `entities: ['sub:std_msgs/msg/String:/chatter']` — so this is the register call → reader → knobs on a real configure, not the verb talking to itself.

The rows also show the per-kind rules landing where they should: the action client budgets a **heavy** slot and opens a feedback subscription it never declared; the action server takes two publishers for feedback and status; the service client claims two callback slots. None of that is stated in a `.msg`.

## 2. "Received-type set resolved but unspent" — spent, and it buys nothing *here*

`cpp-listener-xrce` resolves the set to `std_msgs/msg/String`. The answer is that it changes nothing, for a recorded reason rather than the issue's guess ("may well be nothing"):

**`std_msgs/msg/String` is itself `unbounded` in this image** — the closure refuses with 15 of 32 types open, and `String` is one of them. There's no bound to narrow *to*, and #0963's join refuses exactly there by design.

So this image is a live confirmation of #0963's **guard**, not of its saving. An image that caps `String` would show the saving; this one shows the refusal is right.

## W5.g

Same six rows are the wave's saving on a real cmake workspace image, against crate defaults of `MAX_CBS` 4 and session pools at 8.

**The queryable table specifically is absent from all six, and that's the finding rather than a gap.** The inventory declines to emit it because it can't see whether an image enables the parameter or lifecycle service families (6 and 5 queryables). #1061 records the same refusal on the cargo side.

The earlier `lane=native` attempt stays in the phase doc — that recipe is still wrong, and the next person would otherwise re-run it.

`just check fast` 224/224.

### DIFF
--- a/docs/issues/archived/0965-no-entity-inventory-so-three-consumers-cannot-be-derived.md
+++ b/docs/issues/archived/0965-no-entity-inventory-so-three-consumers-cannot-be-derived.md
@@ -1,7 +1,7 @@
 ---
 id: 965
 title: "Nothing states which entities an image creates, so the arena, MAX_CBS and the payload classes stay hand-set"
-status: open
+status: resolved
 area: codegen, memory, build
 severity: high
 related: [0900, 0939, 0940, 0941, phase-403, phase-403-W4, phase-403-W5]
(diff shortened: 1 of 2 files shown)


jerry73204:
I rebased the other seven conflicting PRs but stopped on this one, because the conflict is **not** mechanical and resolving it carelessly would lose something in a way review cannot see.

**What conflicts.** `docs/issues/archived/0965-…md`. At the text level both sides are pure appends to the end of the same document — neither rewrote a word of the shared body (`diff base→branch` removes exactly one line, `status: open`; `diff base→main` removes nothing). But the two appends contradict each other:

- `main` appended *"## Update 2026-09-05 (phase-412) — the SOURCE moved, the answer did not"*: `ENTITIES` is retired, the statement is now a contract sidecar beside the launch file, `<stem>.contract.yaml`.
- this branch appended *"## Both remaining items MEASURED on configured images (2026-09-06)"* — measured through the `ENTITIES` seam.

I checked the six images the branch measures — `examples/workspaces/cpp/src/{talker,listener,action_client,action_server,service_client,service_server}_pkg/CMakeLists.txt` — and on `main` all six now read `# phase-412 -- \`ENTITIES\` is retired.` So the table was taken through a seam that no longer exists. The section's opening sentence (*"The section above lists two things this issue 'still does not do'"*) also stops being true after a rebase, since "the section above" becomes main's phase-412 update — which already reports the same class of measurement and closes part of that list from the other direction (issue 1083, `service_client_pkg`).

**Why I did not just union the two appends.** This commit also `git mv`s the file into `archived/`. Taking "theirs" wholesale would drop main's phase-412 section, and because the file moves at the same time, a reviewer sees a delete plus an add rather than a diff — so the loss would be invisible in review. That is exactly how `c7f2b09a0` happened (*"docs(0937): the archived file kept the FIRST diagnosis — restore the correction"*, +81/−72 restoring measured text that an archiving commit had shipped stale). This is the second time the archive boundary has come close to eating content in a short window.

**What I think it needs**, but it is your call because it is an authoring decision, not a merge: keep both appends — main's phase-412 update first, then the measurement section either re-taken through the contract sidecar or explicitly labelled as taken through the pre-412 `ENTITIES` seam. Keep `status: resolved` and the archive move only if phase-412 has not reopened the question.

For what it is worth, the file is not a structural conflict surface: `0965-…md` is touched by 2 of the last 300 commits on main, and `docs/issues/` as a directory spreads 123 commits across 118 distinct files. This one is a genuine disagreement about content, not a hot-file collision.

PR #546, opened 2026-09-05
phase-431 W4–W6: the download front door, the release workflow, and docs by audience

jerry73204:
Stacked on #545 (W3) → #543 (W2) → #537 (W1).

## The shape changed, and the reason is structural

W4 was written as **"`bootstrap.sh` downloads by default, `--from-source` opts out"**. That is not implementable:

- `bootstrap.sh` lives **in** a checkout, so its download would put a released binary on a contributor's PATH beside that checkout's sources;
- a released binary run against a nano-ros checkout is refused by #537 and reported as a `[FAIL]` by #543 — both deliberately;
- and it cannot *be* the checkout's binary either: `nros source-stamp` compares against the sources it was built from, so a release from another commit reads stale the moment it is copied into `packages/cli/target/`.

That is RFC-0090's thesis rather than an inconvenience — the release is for people building **their** workspace, and inside nano-ros the tree's own build is the only correct binary. So the two front doors split by **audience**:

| | who | what |
|---|---|---|
| `scripts/bootstrap.sh` | contributors, in a checkout | builds from source (unchanged) |
| `scripts/install.sh` | users, anywhere | downloads a release into the store |

```
curl -fsSL https://raw.githubusercontent.com/NEWSLabNTU/nano-ros/main/scripts/install.sh | sh
```

POSIX `sh`, because it runs from a curl pipe on a machine with nothing provisioned — and with no `nros`, which is the bootstrap paradox it exists to break.

## One answer to "which version does `nros` mean"

It never touches a checkout: it installs into `<store>/nros/<version>/` and then calls the binary it just placed with `nros sdk-front nros --front bin/nros`. So the selection stays #545's `front_newest`, reading the store, and the shell does not get a second opinion — which matters because the shell is the one place that cannot watch the store accumulate. `--front` exists for this caller alone: a user outside a checkout has no `nros-sdk-index.toml` until the install finishes.

## The checksum is mandatory, not best-effort

An asset served over a hijacked CDN and an asset that arrived intact look identical to `tar`, and the person running this cannot inspect what was fetched. **A missing `.sha256` is a refusal, not a fallback.** `zstd`'s absence is probed *before* the download — the same rule `sdk_store::execute` follows (issue 0385) — otherwise it fails deep inside `tar` with `zstd: Cannot exec` after the bytes are already down.

`NROS_INSTALL_URL` relaxes the `--proto =https` restriction, and only for a URL the operator typed: refusing their mirror just pushes them to `curl | tar` with no checksum at all. The checksum stays required on that path too.

## The store version comes from the asset

`share/nros/VERSION`, not `nros --version` — they differ by the `-nrosN` repackaging counter (`0.5.0` vs `0.5.0-nros1`), and the wrong one means a later `nros setup --tool nros` installs a **second copy of the same binary** under the other name. W5's release job writes that file from the index; a pre-W5 asset falls back to `--version` and says so.

## Observed

`tests/nros-installer-tests.sh` (`just check nros-installer`, fast lane) serves a tarball built from the checkout's own `nros` over loopback — no release needs to exist, no network is reached. The refusals are the interesting half, since an asset that installs anyway leaves a binary emitting code into someone's workspace:

```
happy path        -> installed at <store>/nros/9.9.9-nros7, fronted, runs
version pinning   -> the prefix is the ASSET's version, and the only one
bad checksum      -> refuses, and installs NOTHING
absent .sha256    -> refuses (unverified is not a fallback)
no asset at all   -> refuses, naming the source build as the way forward
```

`just check fast`: 233/233.

The W2 gate moved with the decision: it flagged `bootstrap.sh` without `--from-source` as the release path, which is now false. It flags `scripts/install.sh` instead — bootstrap building from source is exactly what the gate wants. W5's newly-discovered requirements (the `VERSION` file, the `.sha256`, the prefix-rooted shape) are recorded in the phase doc.


jerry73204:
Rebased onto main and **retitled**: #547 (W5) and #549 (W6) merged into this branch while it was stacked, so this one PR now carries **W4, W5 and W6** — the whole remainder of phase-431. W1–W3 are on main (#537, #543, #545).

One conflict, in `just/check.just`: main replaced the long `fast-serial:` dependency list with `scripts/build/run-gates-parallel.sh --serial`. Took main's form; the two gates this branch adds (`ci-cli-from-source`, `nros-installer`) are still on the `fast` list, which is a separate recipe.

`just check fast`: 236/236 on the rebased tree.

jerry73204:
Rebased onto `main`, unblocking the merge queue — and my first attempt at it was wrong in a way worth recording.

## Why this PR mattered to everyone else

`check-prose-issue-refs` was red in the **merge group**, not on this pull request, and the queue builds each entry on top of the ones before it — so the batch was ejected and unrelated PRs went with it.

## What I got wrong first

I built the rebase from a **stale local branch ref** that lagged `origin` by two commits, so I never saw `fix(phase-431 W4): the installer test had no caller…` or `docs: drop the issue-1110 file this branch carried`. `--force-with-lease` did not stop the push: the lease compares against the remote you have FETCHED, and my clone had never fetched those two.

That second commit is not bookkeeping — it changes the right answer. Without it I read the tree as "this branch FILES issue 1110", and removed the three baseline rows holding its citations. With it, the branch deliberately **deletes** that file, because the rustdoc lane is green now and an open issue against a green lane is a false claim. So the rows are still needed and my edit was backwards.

Both commits are restored (5 → 6 with the fix below); every commit subject on the pre-push tip is present, checked one by one.

## The actual fix

Deleting the file leaves the phase doc citing an issue with no file — a **fourth** dangling citation, created by the deletion rather than inherited:

```
docs/roadmap/phase-431-ship-the-nros-binary.md cites issue 1110,
and docs/issues/1110-*.md does not exist
```

Added to `.config/prose-issue-ref-baseline.txt` by hand, joining the existing RESERVED-NEVER-WRITTEN group with 0417 and 1080 — with a note saying it arrives from the opposite direction: those two were never filed, while 1110's file existed here and was removed on purpose. `refs/issue-ids/1110` is claimed either way, so the id is not free and the citations are not wrong.

Not `--write-baseline`: that emits rows only and deletes the classification header, which is the part of that file carrying its meaning.

## The one real conflict

`just/check/conventions.just` — `main` appended `pipefail-sigpipe-assertions` and `just-recipe-paths`; W4 appended `nros-installer`; and W6 **moves** `nros-installer` to `workflows.just`. Kept main's two, let W6's move proceed. Verified `nros-installer` now appears in `workflows.just` and not here.

## Verified

`prose-issue-refs` OK (10 known in-flight), `issue-ids` OK, `doc-refs` OK, `doc-recipe-refs` OK, `gate-lists` OK, `just-recipe-paths` OK, `leaf-lockfiles` OK.

jerry73204:
Rebased onto `main` (`632bdbeba`). **6 commits -> 6, nothing dropped** — the subject set before and after is identical.

**One conflict: `.config/prose-issue-ref-baseline.txt`.**

`main` moved this file in `0a02601b5`, which used `--write-baseline`. That sorted the file *including its comment lines*, interleaving the CLASSIFIED header into the row list — so on `main` today the header reads

```
docs/roadmap/phase-432-...:1080
#         issue 9999, which DOES NOT EXIST". Filing 9999 would destroy the
just/check/docs.just:1110
```

This branch's `fix(#1110)` commit edits that header, so the two could not merge.

**Resolution, by hand (not `--write-baseline`):** I restored the header's original order and applied the 1110 edit to it, then re-added `main`'s new row. Measured, not assumed — the scramble lost no content, so restoring the order is lossless:

- `comm` over the *line sets* of the pre-scramble blob (`5cb78b9b2`) and `main`'s version: nothing only-in-pre; the sole only-in-main line is the row `docs/issues/1231-cyclone-assert-publisher-liveliness-null.md:1164`, which this resolution keeps.
- `comm` over row sets, mine vs `main`: the only difference is this branch's own added row `docs/roadmap/phase-431-ship-the-nros-binary.md:1110`.
- The only comment line of `main`'s that is gone is `#         committed — issue 0999's shape ("the id was reserved, the fix`, which is exactly the line `fix(#1110)` rewrites to `#   1110  committed — …`.
- Order is not load-bearing for the gate: `load_baseline()` drops every `#` line and builds a set.

The four 1110 citations **stay** in the baseline, as intended — `959faaf7f`/`f19c9194f` deletes `docs/issues/1110-…md` on purpose because the rustdoc lane is green, and the citations record the 0319/0896 class either way.

**Two additions to the header while I was in it,** both because it promises every row is classified and `main`'s new row was not:

1. A note for `1164` — IN FLIGHT, its file is `docs/issues/1164-cyclone-status-events-unreachable.md` on PR #641 (`work/phase-433-w6-events-cell`), still open. Verified with `gh pr view 641 --json files`; `0a02601b5`'s own message says the same, it just had nowhere to say it after the sort.
2. The `Regenerate:` line now says that flag emits rows only and destroys the classification. That is the mistake this conflict is made of.

The 1164 filename is line-wrapped, matching how 1099's already is: `check-doc-refs` resolves any whole path it can see and failed on the unwrapped form (`.config/prose-issue-ref-baseline.txt:18 -> docs/issues/1164-…md`), which is correct — the file is deliberately not here yet. The wrap is now commented so the next reader does not "fix" it.

**Gates:** `just check fast` — 283 of 284 green; the one red is `xrce-vendored-versions`, which fails identically on a plain `origin/main` worktree ("no vendored tree checked out"), so it is environmental. Also green: `doc-refs`, `issue-ids`, `prose-issue-refs` (11 known in-flight), `doc-recipe-refs`, `check-doc-commit-citations.py`, `cargo +nightly fmt --check --all`, `cargo check -p nros-cli-core --all-targets`.

jerry73204:
Rebased onto `main` (`7ff68c777`). 6 commits before, 6 after; no-loss check (`comm -23` of subjects against pre-push sha `540c93e9d`) empty — nothing dropped.

**One conflict: `.config/prose-issue-ref-baseline.txt`** (the last commit, `fix(#1110)`).

`main` has since REPAIRED that file's header (`fix(#1241)`): the CLASSIFIED block is no longer scrambled, and the new `check_baseline_shape` refuses a baseline whose comments follow the first data row. This branch's copy was the scrambled pre-repair one, so the branch side could not be kept as written — and `--write-baseline` would have re-scrambled it, since it emits rows only.

Resolved by hand on top of `main`'s repaired file, preserving this PR's intent exactly:

- **The four 1110 citations STAY.** Three were already in the baseline on `main` (`docs/issues/1116-…md`, `just/check/docs.just`, `scripts/build/rustdoc-set.sh`); this PR adds the fourth, `docs/roadmap/phase-431-ship-the-nros-binary.md:1110`, in sorted position between the `phase-428` and `phase-432` rows. That citation is created by this branch's own deliberate deletion of `docs/issues/1110-…md` — the rustdoc lane is green, and an open issue against a green lane is a false claim in the series whose job is to say what is broken.
- The classification note is rewritten to sit ABOVE all data rows in `main`'s repaired header, as a `1110 DELIBERATELY WITHDRAWN` entry beside the `0417`/`1080` RESERVED-NEVER-WRITTEN group — same end state, reached from the opposite direction (that file existed, on this branch only, and was removed on purpose; `refs/issue-ids/1110` is claimed either way).

Baseline delta against `main` is +1 row and +1 comment block, nothing else. Confirmed `docs/issues/1110-*.md` is absent from `main` and from `docs/issues/archived/`, so the citations genuinely dangle and genuinely belong in the ratchet.

## Gates

- `just check prose-issue-refs` — OK, **10 known in-flight**, no new dangling reference.
- `just check doc-refs` / `issue-ids` (1211) / `doc-recipe-refs` (self-test 22 cases + 47 known dead, no new) — OK.
- `python3 scripts/check-doc-commit-citations.py` — OK, 197 citations across 907 live files resolve.
- `cargo +nightly fmt --check --all` clean at the repo root AND in the `packages/cli` sub-workspace (this PR touches `nros-cli-core`); `cargo check -p nros-cli-core --all-targets` clean.
- `just check fast` (script + `just/check/tools.just` + workflow changes earn it) — **2 of 285 failed, both environmental in a fresh worktree**: `xrce-vendored-versions` (micro-XRCE trees not checked out) and `capability-conditionals` (zenoh-pico submodule not populated). Both reproduced with identical messages on a plain `origin/main` worktree; after `git submodule update --init packages/rmw/zenoh/zpico-sys/zenoh-pico`, `capability-conditionals` is **OK (5 platform manifests, 5 capability-gated rows, IP_STACK = 'ip_stack')**.

PR #550, opened 2026-09-05
fix(rmw-doc): a visitor parameter hid the fields it yields, so the row read as a loss

jerry73204:
Stacked on #541 (queued); GitHub retargets this to `main` when that lands.

Second reader in two turns concluded we had dropped something, and both times the page was under-reporting our side rather than the tree being wrong. This one: *"ours does not return the namespace, does it"*.

**It does.** `rmw_node_visit_fn` is `(ctx, node_name, node_namespace, enclave)`, and both producers pass a namespace — zenoh demangles it off the liveliness token (`shim/session.rs:1236`), cyclone forwards its adapter's `ns` (`vtable.cpp:137`).

What the page showed was the **slot** signature, where upstream enumerates `node_names` and `node_namespaces` as separate out-parameters and ours collapses to `rmw_node_visitor_t visitor`. Two words against four — the comparison the page exists to support led straight to the wrong conclusion.

The payload now prints beside the parameter:

```c
rmw_node_visitor_t visitor  /* visit(ctx, node_name, node_namespace, enclave) */
```

**Parsed from the `*_visit_fn` typedefs in `rmw_vtable.h`, not restated.** The visitor argument list is the thing most likely to gain a field — issue 0785 added `enclave` to this exact one — and a hand-copied list would have been wrong the first time that happened.

Covers all three visitors (node, names-and-types, topic-endpoint-info), ten rows in total.

## Verification

`just check rmw-api-comparison` OK (88 symbols) · `rmw-api-parity` · `rmw-abi-shape` · `just check fast` 231/231.

Tally unchanged — 13 same, 34 re-shaped, 28 re-mapped, 13 not-supported. This changes what a cell **says**, not what any comparison computes.

https://claude.ai/code/session_01JaST26nBgH9mLezz4TWKmP

### DIFF
--- a/book/src/reference/rmw-api-comparison.md
+++ b/book/src/reference/rmw-api-comparison.md
@@ -523,7 +523,7 @@ text-transform:uppercase;margin:0 0 .45rem;opacity:.95}
   <span class='ty add'>const rmw_session_t *session</span><span class=pu>,</span>
   <span class='ty'>const char *node_name</span><span class=pu>,</span>
   <span class='ty'>const char *node_namespace</span><span class=pu>,</span>
-  <span class='ty add'>rmw_names_and_types_visitor_t visitor</span>
+  <span class='ty add'>rmw_names_and_types_visitor_t visitor  /* visit(ctx, name, types, types_count) */</span>
 <span class=pu>)</span></pre></td>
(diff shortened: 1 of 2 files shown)


jerry73204:
Rebased onto current main: **4 commits -> 1**. The `check-submodule-pins` failure was not a pin problem at all.

Two layers, both diagnosed before anything was touched:

**The reported error was a one-time race.** The step failed with `fatal: couldn't find remote ref refs/heads/work/rmw-comparison`. `GITHUB_BASE_REF` is baked into a job's env at trigger time; this PR was opened against `work/rmw-comparison`, that branch merged and was deleted mid-run (`automatic_base_change_succeeded` two minutes into the same run), and GitHub retargeted the PR to `main` while the job kept the stale ref. No workflow change is warranted — a fresh trigger reads `main`.

**But a re-run would still have failed, for a real-looking reason that was also not a defect.** `submodule-pins` reported `REWIND` on `packages/cli/third-party/play_launch`. None of this branch's commits touch that path; the branch was simply **281 commits behind main**, and `main` had independently advanced the pin. The gate compares the branch tip against current `main`, so a stale-but-correct branch reads as a rewind. Rebasing adopts main's already-merged advance — it is not a pin decision, and `just check submodule-pins` now reports `OK (0 pin(s) moved)`.

Three of the four commits dropped in the rebase because they are already on main under new hashes — verified by subject: `1e215bb69`, `10a6c8a7e`, `a83104079`. The surviving commit is the PR's actual subject, `fix(rmw-doc): a visitor parameter hid the fields it yields`.

PR #552, opened 2026-09-06
fix(phase-413 W2): the other two lanes check their labels AFTER provisioning too

jerry73204:
The third and second sites of an ordering defect whose first site was fixed in `#279`. Both commits were written on 2026-09-04, both sat on branches whose PR merged without them, and both are still broken on `main` today.

## The defect

`runner-doctor.sh` checks `$root/build/qemu/bin/qemu-system-arm` **first** — *"the project-local patched build is the PRIMARY path (phase-143) … a host with it needs nothing from the system"* — and only falls back to the system qemu when that is absent.

That path is inside the **checkout**, and `just setup tier2` is what creates it: the tier-2 preset includes the `qemu` module, whose `setup` ends in `just qemu setup-qemu`. So asserting the runner's labels *before* provisioning asks the host for something the next step supplies. On `run-matrix` that killed every run at step 2 against a system qemu 6.2, having provisioned and tested nothing.

## The three sites

```
                 before          after
run-matrix.yml   fixed in #279   unchanged
build-wide.yml   doctor  84      doctor 107   (setup  85)
queue.yml        doctor  70      doctor 117   (setup 100)
```

A genuinely mislabeled runner still fails — one step later, and `setup` would have failed on it first.

## Exposure

`build-wide.yml` is `workflow_dispatch`-only, so an operator dispatching it is the only one exposed. `queue.yml` runs on **`merge_group`**, but is currently **latent** rather than firing: its label list is `nros-sdk-zephyr,nros-big` with no `nros-qemu`, so on the present runner the doctor step passes (verified on run 33991966002 — the doctor step is green and the job fails later, at `just ci matrix build`). It is one label change from firing.

The reason to land it anyway is the one the original commit gives: the three copies of this job should not disagree about when the check runs, which is the class of drift phase-413 W1 exists to remove.

## Not included

There is **no gate for this ordering** — I looked; no `check-*` script asserts that a `runner-doctor.sh` step comes after the first `just setup`. That is what let two of three sites regress silently after `#279`. A gate would be cheap (compare step indices per workflow) and is the issue-0196 rule applied here, but it is a separate decision from landing the fix, so it is not in this PR.

Both commits are cherry-picks, `-x` trailers retained.

### DIFF
--- a/.github/workflows/build-wide.yml
+++ b/.github/workflows/build-wide.yml
@@ -80,14 +80,32 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v4
-      - name: Verify this runner's labels are true
-        run: ./scripts/ci/runner-doctor.sh nros-sdk-zephyr,nros-big
       # phase-411: a CI job is `just setup <scope>` then one command a developer
       # can type. Nothing here is CI-only orchestration.
       - name: just setup tier2
         run: |
           source ./activate.sh
           just setup tier2
+
+      # AFTER `setup` — phase-413 W2, same reason as `run-matrix.yml` (#279).
+      #
+      # `runner-doctor.sh` checks `$root/build/qemu/bin/qemu-system-arm` first
+      # ("the project-local patched build is the PRIMARY path (phase-143) … a
+      # host with it needs nothing from the system") and falls back to the
+      # system qemu only when that is absent. That path is inside the CHECKOUT
+      # and `just setup tier2` creates it — the tier-2 preset includes the
+      # `qemu` module, whose `setup` ends in `just qemu setup-qemu`.
+      #
+      # Asserting before provisioning asks the host for something the next step
+      # supplies. On `run-matrix` that killed every run at step 2 against a
+      # system qemu 6.2, having provisioned nothing. This lane is dispatch-only
+      # now, so the exposure is smaller — but it is the same bug, and an
+      # operator who dispatches it deserves the same answer as the cron lanes.
+      #
+      # A genuinely mislabeled runner still fails, one step later, and `setup`
+      # would have failed on it first.
+      - name: Verify this runner's labels are true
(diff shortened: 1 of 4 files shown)


jerry73204:
Added the gate this PR's body said it was leaving out — `check-workflow-doctor-after-setup`, now the third commit.

The argument for putting it here rather than in a follow-up is the PR's own finding: this defect had three sites, `#279` fixed one, and the other two survived months because nothing was watching. Landing the sweep without the gate would repeat exactly that.

**What it checks.** Within one job, if a step invokes `runner-doctor` and another invokes `just setup`, the doctor step must come after the *first* setup step. Only within a job — `build/qemu/bin/qemu-system-arm` is created in the job's own checkout and nothing carries it between jobs. A doctor with no `just setup` is fine: nothing to provision, nothing to assert too early.

Four jobs qualify today — `build-wide`, `queue`, `run-matrix`, and nightly's `matrix-nightly` — and all four pass with the fixes in this branch.

**It refuses to pass over nothing.** If the scan finds zero jobs carrying both kinds of step it exits non-zero instead of reporting OK. Not hypothetical: `check-lane-contracts` was found this week printing `OK — 0 test target(s) across 3 affordability tier(s)`, which is this failure one level up.

**Mutation tested.** Reverting the two workflow files to their pre-fix state on `main` makes it red, naming both jobs and both step indices:

```
check-workflow-doctor-after-setup: 2 job(s) assert their labels BEFORE provisioning:

  build-wide.yml  [l3]
      step 1 "Verify this runner's labels are true" runs the doctor,
      step 2 'just setup tier2' provisions.
  queue.yml  [l3]
      ...
```

Restoring them makes it green. It joins the fast lane automatically (the list is derived, issue 1072); `gate-lists`, `gate-selftests` and `default-gates-run-somewhere` all pass.

PR #557, opened 2026-09-06
ci(#1059, #1066): land the node-std-tests merge gate that was written and never merged

jerry73204:
This step was authored 2026-09-05, verified locally, and described in issue
#1059, in phase-428 and in PR #438's body as **landed**. It was not.

```
$ grep -c "node-std-tests" .github/workflows/gate.yml   # origin/main
0
$ git show 2fd56e28d504:.github/workflows/gate.yml | grep -c node-std-tests
0
```

Exactly one commit in the repo ever contained it, on a branch that is still open.

So the fourth red to reach `main` in that lane did **not** get past a gate — it
got past the absence of one, for the ordinary reason: `node-std-tests` runs only
in `check-build`, which is `schedule`/`workflow_dispatch` only.

## How it was diagnosed, because the wrong turn is the useful part

`5200437c5` **is** a descendant of #438's merge commit, so the tree the queue
tested would have carried the step had the step existed. That looked damning,
and the reach was for two elaborate explanations — an in-flight queue batch, an
environment-dependent test — before checking whether the code was there at all.
One `git merge-base --is-ancestor` and one `grep -c` settled it.

**A claim believed because it was AUTHORED rather than OBSERVED** is the exact
failure this campaign documents in other people's work.

## Also carried, from the same commit

phase-428 W2's prerequisites: `clang` in the CI image and the pinned nightly for
the rustdoc extractor. Both are needed before `check-api-parity` can run in a
container at all (#1066) — it is currently run by **no workflow on any event**.

## Verified

`just check lane-contracts` — 3 merge-gating lanes, none resolving an artifact
its job does not build. YAML parses, 7 jobs.

Measured cost, both cache states: **20.6 s cold, 2.4 s warm.** No fixture, no
SDK, no cross toolchain, no QEMU, no ROS.

### DIFF
--- a/.github/workflows/gate.yml
+++ b/.github/workflows/gate.yml
@@ -759,6 +759,64 @@ jobs:
           source ./activate.sh
           just check cli-tests
 
+      # issue 1059 — `node-std-tests` on BOTH merge-gating events, for the same
+      # reason `cli-tests` is above it, and with the same evidence behind it.
+      #
+      # It lived only in `check-build`, which is `schedule` / `workflow_dispatch`
+      # only, so nothing on the path to `main` ran it. `main` then carried a
+      # compile error for a day: phase-421 W1-W3 reverted an associated type
+      # across 142 impls and left NINE behind, and
+      #
+      #     error[E0437]: type `Format` is not a member of trait
+      #                   `nros_serdes::schema::Message`
+      #
+      # is not a subtle red. It hid because all nine survivors are
+      # `#[cfg(test)]`: `cargo check -p nros-node --lib --features std` PASSES,
+      # because test cfg is not compiled. Only a test BUILD reaches them, and no
+      # merge-gating lane did one. Third red to reach `main` that way in a single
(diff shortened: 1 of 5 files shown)


jerry73204:
Rebased onto current main. Dropped `2598a7c1e` ("the backing static's positive control gets a process of its own") — main landed a different fix for the same problem in the meantime: it marks `the_static_is_handed_out_once_and_only_once` `#[ignore]` and runs it via `-- --ignored --exact` in `node-std-tests`, asserting `1 passed` so the row cannot pass vacuously. That supersedes this branch's separate-test-binary approach, so `tests/executor_backing_static.rs` is not reintroduced.

The other two commits carry work that is still unlanded on main and are kept unchanged: the `sim-time` cfg gate (issue 1059) and the gate.yml/Dockerfile/extract_rust.py changes (issues 1059, 1066).

Verified: `just check node-std-tests` green, `gate.yml` parses, `just --list` clean.

PR #577, opened 2026-09-06
fix(#1015): a zero that sizes a C array is refused by every producer, at configure

jerry73204:
Closes #1015.

## The gap

`a38382164` floored **three** of the ten fixed array extents in `zpico.c` and closed nothing — which is issue 1015's own "Not covered" note: **a floor in one producer does not bind another.** `ZPICO_MAX_LIVELINESS` and `ZPICO_MAX_PENDING_GETS` are Kconfig-settable and size arrays in the very same struct, one `.conf` line away from the identical silent board.

For context on why "silent" is the word: the reference island derived `ZPICO_MAX_QUERYABLES = 0`, `queryable_entry_t queryables[0]` is a GNU extension that compiles silently and is not the last member of its struct, and the board transmitted **0 bytes in 15 seconds** — no panic, no log line, no fault, core in WFI. Every gate was green and correct. `check-knob-delivery` confirms the value ARRIVED; it did. `check-knob-fixpoint` converges; 0 is stable. Every check in the tree asks whether the number the image was BUILT with is the number that was DERIVED. **None asked whether the number is USABLE.**

## One rule, at every point that can enforce it

| where | what | binds |
| --- | --- | --- |
| derivation | `floor1` in `entity_inventory.rs` (already there) | the derived pools |
| **cmake** | `nros_assert_c_array_extent()` — `cmake/NanoRosCArrayExtents.cmake`, called from `nros_rmw_zenoh.cmake` for all 7 extents it `-D`s | environment, `.conf`, derivation, literal fallback — all four |
| **cargo** | `ShimConfig::check_c_array_extents()` | a build cmake never sees: a `fixtures.toml` `env` row, a shell export, a bare `cargo build` |
| **C source** | `#if <M> < 1` / `#error`, **3 → 10** extents | every producer, including one written tomorrow |

The cmake refusal is the one the issue asked for: it fires at **configure**, before a compiler exists, and names the macro, the array it sizes, the Kconfig option that overrides it, and where the declaration was supposed to come from. It sits at the point cmake commits a number to a compiler rather than at any of the four upstreams, so it binds all four and any fifth — phase-412 shipped six delivery-class defects in one wave because checks lived at individual upstreams.

It separates the three illegal states, which have different causes and want different advice:

- **EMPTY** — `-DX=` reaches the compiler and fails as `flexible array member not at end of struct`, on a struct nobody edited.
- **`-1`** — the DERIVE SENTINEL escaped its resolver; `size of array is negative`.
- **ZERO** — this issue.

`check_queryable_override` next to the cargo check does **not** already cover this: it asks whether the table fits the *declared* services and returns early when nothing was declared — which is exactly the state that derives zero.

## The gate the issue asked for

`check-c-array-extent-floors`, on the fast line. Keyed off the **array declarations in `zpico.c`**, not off a list, because a list is what the first fix was.

- **R1** every extent has a `#if <M> < 1` guard whose body actually raises `#error` (an empty body reads exactly like a guard).
- **R2** every extent the cmake lane `-D`s is asserted first — **R2b** guarding the *same expression* that is emitted (a guard on a different variable is exactly as green and exactly as useless: phase-412 W1's built-up `NROS_DERIVED_${_pool}` that resolved EMPTY), **R2c** *before* the emission.
- **R3** every extent `ShimConfig::defines()` emits is listed, **and nothing else is** — a floor asserted where the hazard does not exist is how a check stops meaning anything.

Nine self-test cases, one of them the scanner's own negative control: a `zpico.c` with no arrays is REPORTED, not passed. **Measured against the pre-fix tree: 15 problems**, including the 7 unguarded extents.

Behaviour rather than presence: `tests/cmake-c-array-extent-tests.sh` (`just check c-array-extent-floor-behaviour`, `cmake -P`, ~1 s, 26 assertions).

## Proof, both directions

**Still builds.** A pool of 1 — what the derivation's own floor produces, so the value a real single-service image arrives with — plus 4, 8, 65536, all configure. `zpico-sys` builds and links for real through the new cargo check (`just check rustdoc-links` compiles it). A tree-wide sweep finds **nothing** setting any of these knobs to 0 in any `.conf`, `.toml`, `Kconfig` or justfile, so no existing image moves.

**Fails loudly.** 0 in any of the ten extents is refused by every producer that supplies it. The C backstop was verified by extracting the ten guard blocks **verbatim** from `zpico.c` and compiling them: all ten at 1 compiles clean, and each one at 0 fails the compile naming itself.

**The tests catch a removed floor.** Mutating `floor1` to the identity fails `every_session_pool_floors_at_one` and `topics_become_per_node_subscriptions_and_publishers`.

## Tested

- `just check fast` — **232 ran, 0 failed**, 5 skipped for a missing built in-tree `nros`
- `cargo test -p nros-zpico-build` — 34/34
- `nros-cli-core` entity-inventory — 49/49
- `tests/cmake-c-array-extent-tests.sh` — 26 assertions

Tier 1 (`just ci`) and above were not run — this is a fresh worktree with no fixtures, and the change is confined to a configure-time assertion, a build-script assertion, C preprocessor guards, and two new gates.

Sweep for the next person: `python3 scripts/check-c-array-extent-floors.py`

## Not closed

The wider question issue 1015 raised — *what other derived value has an illegal special case at some boundary* — stays open. What is closed is the `0`-on-a-C-array case for every pool in the zenoh backend, with a gate that extends itself to a pool added later. The XRCE caps are the same shape one layer over and belong to #1033; nothing here touches that backend.


jerry73204:
I rebased the other seven conflicting PRs but **stopped on this one without touching it**, because resolving it would mean choosing between two gates that disagree — an authoring decision, not a merge.

**`main` already landed the gate this issue asked for.** `scripts/check-c-array-pool-floors.py`, with a `## Resolution` section in `docs/issues/archived/1015-…md` and a mutation table. This branch adds `scripts/check-c-array-extent-floors.py` for the same defect, and rewrites the same document with its own `## Fix` / `## Proof, both directions` / `## Still open, deliberately` sections.

The two are not the same rule:

- **branch:** *"a number that sizes a FIXED C ARRAY has a floor of one, and every producer of that number enforces it."*
- **main:** *"It refuses the general rule 'pools are positive' — 1033 measured that false — and requires a DECISION per knob instead: a `#if <KNOB> < 1` + `#error` beside the array, or a `ZERO_LEGAL` entry naming what measured the zero."*

So `main`'s version explicitly declines the blanket floor-of-one rule this branch enforces, citing issue 1033 as having measured it false. Landing both would give the tree two fast-lane gates for one rule, one of which forbids what the other permits.

The document conflict has the same shape and is the more dangerous half: both sides append to `docs/issues/archived/1015-…md`, identical through `## Options` and then divergent. A union would leave the archived record with **two contradictory resolutions**, which is exactly how `c7f2b09a0` (*"docs(0937): the archived file kept the FIRST diagnosis — restore the correction"*) happened.

What I think it needs, but it is your call:

1. Decide which rule is right — per-knob decision (main) or floor-of-one (this branch). Issue 1033 is the evidence that settles it, and `main`'s text says it already did.
2. If main's stands, this PR reduces to whatever `check-c-array-extent-floors.py` catches that `check-c-array-pool-floors.py` does not — possibly nothing.
3. The doc keeps one resolution section, not two.

The rest of the branch rebases cleanly. Its `entity_inventory.rs` conflict is a genuine add/add (both sides add issue-1015 tests, no duplicate fn names) and unions safely, and its two gate recipes migrate into `just/check/abi.just` under the new topic layout — I have both resolutions ready and will apply them the moment the rule question is settled.

jerry73204:
## Superseded by `49825c4d9` — recommend closing rather than rebasing

I was asked to rebase this onto current `origin/main`. I did not, because the conflicts are not textual: `49825c4d9` ("a pool's floor belongs to the pool, not to the shared derivation") landed a **different and better** fix for #1015, and this branch now contradicts it in a way that would reintroduce a measured regression.

### The contradiction, measured

This PR floors the pools inside `EntityInventory::derive()` and asserts that in two unit tests. Main deliberately **removed** that floor. Applying this PR's assertion verbatim to main's `derive()`:

```
assertion `left == right` failed: PR #577: a pool that backs a fixed C array
has a floor of ONE, even with nothing declared (issue 1015)
  left: 0
 right: 1
```

Making this PR green again means putting `.max(1)` back into the shared derivation — which is exactly what `49825c4d9` diagnosed as **silently defeating issue #1033**. The same derived numbers feed `NROS_XRCE_MAX_SUBSCRIBERS` / `NROS_XRCE_MAX_SERVICE_SERVERS`, where zero *is* the answer and is worth **33,296 and 4,384 bytes of heap per slot** (measured from the zephyr cpp listener's DWARF). The floor landed 2026-09-04 19:56; 1033's fix landed 2026-09-05 07:28. I did not know that when I wrote this.

### The other half: this PR's extra `#error` guards break main's gate

This PR extends the guards in `zpico.c` from 3 to 10 extents on the reasoning "fix the class". Main's `check-c-array-pool-floors` explicitly **refuses the general rule "pools are positive", which #1033 measured false**, and requires a per-knob decision. Adding one of this PR's guards to main:

```
[FAIL] ZPICO_MAX_LIVELINESS is guarded now: remove it from UNCLASSIFIED.
```

Five of the seven extents I guarded are in main's UNCLASSIFIED ratchet on purpose, tracked as **#1131**. Guarding them here is *making* the ruling that issue exists to make deliberately — and it is the same over-generalisation that cost #1033.

### Everything else is redundant

| this PR | main | verdict |
| --- | --- | --- |
| `check-c-array-extent-floors.py` — 10 arrays in `zpico.c`, guarded/not | `check-c-array-pool-floors.py` — **21** arrays tree-wide (zenoh, XRCE, platform, cpp, uorb), guarded / **zero-legal** / unclassified-ratcheted | superseded; mine cannot even express "zero is legal here" |
| `nros_assert_c_array_extent()` fatal at the `-D` site | `_nros_c_array_pool_floor()` floor at the resolver | main's floor is upstream, so my assertion is dead on the derived path; for an explicit `0` main deliberately prefers `zpico.c`'s `#error` naming the knob over a configure fatal |
| producer-layering rules R2b/R2c | `check_producers()` — already checks producers floor guarded knobs, that abstentions are written down, and that the shared derivation does **not** raise a count | mine are properties of a mechanism main doesn't use |
| archived issue doc | already archived with its own resolution text | duplicate (the `AA` conflict) |

Two overlapping gates with different classifications of the same arrays is the "two green tools disagreeing" antipattern this repo keeps paying for — the argument is in CLAUDE.md's `rmw-api-parity` note.

### Recommendation

**Close unmerged.** I've left the branch as-is rather than force-pushing a rebase, since every substantive piece is superseded, redundant, or actively contradictory, and a rebase would only make the contradiction harder to see. Nothing here is worth salvaging on top of `49825c4d9`; the remaining work in this area is #1131, which is main's ratchet to unwind knob by knob with a measurement each time.

The one thing worth carrying forward is a non-finding: this PR's premise — that the class fix is "floor everything that sizes a C array" — is wrong for this codebase, and #1033 is the measurement that says so.

jerry73204:
Closing unmerged — superseded, and rebasing this would push a regression.

`49825c4d9` ("fix(#1015): a pool's floor belongs to the pool, not to the shared derivation") landed on main from another session while this was in flight. It is a better fix for the same issue, and this branch now contradicts it. Measured rather than argued:

- This PR floors the pools inside `EntityInventory::derive()`. Main **deliberately removed** that floor. Applying this PR's assertion verbatim onto main's `derive()` fails `left: 0, right: 1` — and restoring `.max(1)` there is exactly what `49825c4d9` diagnoses as silently defeating issue #1033, where zero *is* the answer and is worth 33,296 B and 4,384 B per unused slot (from DWARF).
- The "fix the class" half extended `zpico.c`'s guards from 3 to 10 extents. Main's gate **explicitly refuses** the general rule "pools are positive", which #1033 measured false. Adding one of those guards to main's tree fails its gate: `ZPICO_MAX_LIVELINESS is guarded now: remove it from UNCLASSIFIED.` Five of the seven extra guards sit in main's UNCLASSIFIED ratchet on purpose, tracked as **#1131**.
- Scope: this gate covers 10 arrays in one file with a two-way rule and cannot express "zero is legal here"; main's covers 21 tree-wide (zenoh, XRCE, platform, cpp, uorb) with guarded / zero-legal / unclassified.

The durable finding is that the premise behind this PR — *the class fix is to floor everything that sizes a C array* — is wrong for this codebase, and #1033 is the measurement that says so.

Remaining work is #1131's ratchet, unwound knob by knob with a measurement each. Nothing here is salvageable on top of `49825c4d9`.

jerry73204:
Reshaped to the non-contradictory subset, as discussed. **The previous tip is `99df8ae7d`** if you want anything back.

## What this PR is now

`+47 / -6` across three files, instead of `+1470` across eleven:

- `zpico.c` — the `#if <KNOB> < 1` / `#error` guards, taken from this branch, for the extents `main`'s gate still carried as `UNCLASSIFIED`;
- `check-c-array-pool-floors.py` — those five knobs leave the ratchet, ceiling `15 -> 10`;
- the archived issue — one dated section recording it.

```
before   3 guarded, 3 zero-legal, 15 unclassified
after    8 guarded, 3 zero-legal, 10 unclassified
```

Two of the five (`ZPICO_MAX_LIVELINESS`, `ZPICO_MAX_PENDING_GETS`) are Kconfig-settable and size arrays in the **same struct** as `ZPICO_MAX_QUERYABLES` — so the board that went silent had four more ways to reach the same state. That was this branch's real finding, and it is kept.

`ZPICO_GRAPH_CACHE_SIZE` and `ZPICO_ZID_SIZE` are guarded too. The gate tracks neither (neither is build-settable), but both are extents in the same file and leaving two unruled beside eight ruled is a state the next reader has to re-derive.

## What was dropped, and why

`scripts/check-c-array-extent-floors.py`, `cmake/NanoRosCArrayExtents.cmake`, `tests/cmake-c-array-extent-tests.sh`, the `nros-zpico-build` and `zephyr/cmake` machinery, and the competing rewrite of the issue doc.

They implement *"a number that sizes a FIXED C ARRAY has a floor of one, and every producer of that number enforces it."* `main`'s gate declines exactly that rule, and cites the measurement:

> issue 1033 lowered the XRCE pool minima from 1 to 0 **on purpose** — the arrays are plain mid-struct members not flexible ones, `arr[0]` compiles on gnu99/c11/gnu11, every walk is `for (i = 0; i < MAX; ++i)` and does not run at 0, and a zero is worth **33,296 bytes of heap per subscriber slot** and **4,384 per service-server slot** from the zephyr cpp listener's DWARF — the difference between an image that fits its 65,536-byte heap and one that does not.

Your R1 is scoped to `zpico.c`, so it does not collide with those three knobs *today*. The stated rule does, and the next person applies the rule rather than its scope. Hence: same guards, per-knob decision, no second gate.

Also dropped: the `entity_inventory.rs` test. `main` already has an equivalent one whose comment reconciles both issues — *"the derivation publishes DEMAND, and the floor belongs to whichever consumer's storage cannot be empty"* — which is the settled position this PR no longer needs to restate.

## Verification

`zpico.c` cannot compile here (the zenoh-pico submodule is not checked out), so the guards were driven through `cc -E` directly. Each of the seven fires at 0 and is silent at 4, and all ten guards sit at preprocessor nesting depth **0**, so none is inside a conditional a build could switch off.

`check-c-array-pool-floors`, `doc-refs`, `issue-ids`, `ledger-doc-single-home`, `book-links` all pass; `nros-cli-core` compiles.

If you would rather keep the cmake-side enforcement (your R2/R3 — binding the Zephyr lane's `-D` at the point it supplies the extent), that is a separate and compatible PR: it enforces *whatever floor the knob was decided to have*, which is orthogonal to the floor-of-one rule. Say so and I will lift it out of `99df8ae7d`.

PR #602, opened 2026-09-06
fix(#1099): lifecycle transition ids take lifecycle_msgs' numbering — ours crossed four of eight

jerry73204:
Stacked on `phase-428-w10-qos-ssot`, where issue 1099 lives.

`LifecycleNode::trigger_transition(uint8_t)` took upstream's exact signature
over a different id space. Verified against `lifecycle_msgs/msg/Transition.msg`:

| id | ours | upstream |
| --- | --- | --- |
| 2 | Activate | **CLEANUP** |
| 3 | Deactivate | **ACTIVATE** |
| 4 | Cleanup | **DEACTIVATE** |
| 8 | ErrorRecovery | **DESTROY** |

So `trigger_transition(2)` on an Inactive node silently **activated** it where
ROS 2 cleans up, returning `ok`.

## Renumbered the core enum, not the C++ boundary

Translating only in `lifecycle.hpp` makes it worse: the C side carried the
identical latent defect (`NROS_LIFECYCLE_TRANSITION_CLEANUP == 4`), so a
boundary fix leaves `NROS_LIFECYCLE_TRANSITION_ACTIVATE` and
`trigger_transition(2)` meaning different things inside one mixed image, both
spellings looking correct. **Two independent audits reached this conclusion
without seeing each other.** The issue's own diagnosis (C++ header defect) was
too narrow.

ABI checked first: the discriminant crosses only as cbindgen `#define`s — no
`rmw_entity.h` entry, no bindgen output, no `_Static_assert`.

`ErrorRecovery` → 60 = `TRANSITION_ON_ERROR_SUCCESS`, the id the wire already
emitted (upstream's 8 is DESTROY, which we do not implement). `shutdown()` now
resolves from the current state — it hardcoded id 5, legal only from
Unconfigured. `autostart` used literals `1`/`2`; `2` is now CLEANUP, so
autostart-to-Active would have configured then immediately cleaned up.

New `const _` block pins the C literals to **upstream** values. Issue 0792's
block proves header-agrees-with-enum, which cannot catch both being wrong
together — which is what happened.

State ids deliberately **not** renumbered: `ErrorProcessing = 5` vs upstream
15, but 5 is unassigned upstream, so it can never silently mean a different
state. Recorded with a test.

## Verified

Mutation-tested against the pre-fix numbering: 2 nros-core + 2 nros-node tests
fail, and the new compile probe fails all six `static_assert`s. `just check c`,
`just check cpp`, `just check fast` (227 gates); nros-core 67/67, nros-node
365/365, nros-c 84/84, nros-cpp 17/17.

### DIFF
--- a/docs/reference/api-parity-ledger/lifecycle.json
+++ b/docs/reference/api-parity-ledger/lifecycle.json
@@ -438,7 +438,7 @@
   "cpp:Transition": {
     "verdict": "gap",
     "bucket": "theirs-only",
-    "why": "OUR LANGUAGES DISAGREE: Rust has a full `LifecycleTransition` enum (`from_u8`, `from_shorthand`, `source_state`), C has the `NROS_LIFECYCLE_TRANSITION_*` constants, and C++ has NO transition type or constants at all -- a caller passes a bare `uint8_t` to the protected `trigger()`. rclcpp's `Transition` carries `id`, `label`, `start_state`, `goal_state`. Nothing forbids an `enum class` here: we already ship `enum class LifecycleState` in the same header. Bucket-covered with its members. Phase 379 W3."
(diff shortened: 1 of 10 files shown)


jerry73204:
Rebased onto main and retargeted there: the branch never depended on `phase-428-w10-qos-ssot` (one commit over main), and that branch is itself being rebased over #329. The only conflict was `just/check.just`: main moved the `cpp` lane into `just/check/lanes.just`, so the issue-1099 compile-TU block now sits there, after the graph-query-entry-points TU.

Verified locally with #612's `nros_node_get_fully_qualified_name` fix cherry-picked on top (main's `nros-c` does not compile without it, E0428): `cargo build -p nros-c` green, `lifecycle_transition_ids.cpp` `-fsyntax-only` green, `cargo test -p nros-core --lib lifecycle` 15 passed, `cargo test -p nros-node --lib --features std lifecycle` 26 passed. Expect the compile-smoke check to stay red here until #612 merges.

PR #618, opened 2026-09-06
fix(#1134, #1133, phase-215): the Arm FVP was never licence-gated, and phase-215 is verified on the model

jerry73204:
Three things, one theme: **a belief nobody re-measured.** phase-215 now has **zero open boxes** — both remaining acceptance bullets verified on a real FVP.

## The Arm FVP was never licence-gated

`autoware-safety-island` had the evidence all along: its `.github/scripts/run-zephyr-fvp-ci.sh` downloads the model in CI from an Arm CDN permalink with a pinned SHA256 and **no authentication**. Measured here: HTTP 200, 68,450,956 bytes, digest reproduces byte-for-byte. Our index meanwhile said `version = "11.24"` while the pin anyone actually used was 11.31.

So `[gated.arm-fvp]` → `[tool.arm-fvp]`, and the "declared, never downloaded" policy that cost every user a manual install is gone. Verified end to end:

```
nros setup --tool arm-fvp  → prebuilt 11.31.28 → ~/.nros/sdk/arm-fvp/11.31.28
resolve-fvp-bin.sh         → …/arm-fvp/11.31.28/bin
FVP_BaseR_AEMv8R --version → Fast Models [11.31.28]
```

The board field `gated = [...]` becomes `tools = [...]` — a field named for a **policy** invited exactly this mistake. `gated` stays a serde alias; an out-of-tree descriptor is a user's file.

## phase-215 verified on the model

```
just zephyr build-fvp-board-import → zephyr.elf linked via nano_ros_use_board()
just zephyr run-fvp-board-import   → nros: smoke ok
```

The one-call bar is met by `fixtures/board_import_fvp/CMakeLists.txt`, six lines total.

**Two defects only running it could find**, both fixed:

- **`ARMFVP-NOTFOUND`** — Zephyr's `armfvp.cmake` does `find_program(... PATHS ENV ARMFVP_BIN_PATH)` at **configure** and caches it, so exporting at run time is too late and the run dies after a build that succeeded. This is issue 1134's family one layer out, and the first *confirmed* instance of it. The build recipes export it too now.
- **`failed to read SDK index`, twice per run** — `nros sdk-path` resolves the index relative to CWD, and all four recipes called it *after* `cd "$workspace"`. It fell back to the host `make`/`ninja` silently, in the recipes whose whole point is that local equals CI.

## #1134 — fixed

`knob_usize` ended in `dotconfig_usize(k).unwrap_or(default)`, and that `Option` carried three situations meaning different things. The third — **DOTCONFIG set, file unreadable** — silently took the crate default, so the image compiled, linked, and behaved as though its Kconfig said nothing. `KnobSource` separates them; that case now panics with the path, the error, and the likely cause. Mutation-tested: collapsing the arm back fails the test.

## #1133 — retracted, not fixed

I could not reproduce it, and measurement refutes the stated cause — both candidate triggers clear on the **first unpatched run**. The real explanation is CLI resolution across checkouts, already documented and guarded. I reverted the retry I had written rather than ship machinery for an undemonstrated condition.

The episode justified one thing: the STALE message now names the **checkout**, not just the binary. (It earned that twice — the first FVP build in this branch failed because `nros` on PATH was the main checkout's binary, behind `origin/main`, rejecting the `front` field the current index uses.)

## Testing

`just ci gate`: **PASSED** — 1139 tests, 1138 passed, 1 skipped for a host capability. `just check fast`: 239 green.

The FVP index row was verified by direct download + digest + install + run. The full `verify-index.py` sweep re-downloads every asset in the index and exceeded 40 minutes locally; CI's `sdk_index` job runs it.


jerry73204:
Correcting the testing note above: the full `verify-index.py` sweep **did** complete locally, after ~50 minutes.

```
ok   arm-fvp/linux-x86_64  (68450956 bytes)
verify-index: index coherent; 36 prebuilt asset(s) verified
```

The PR body says the sweep exceeded 40 minutes and was left to CI. That was true when written and I pushed without it; the run finished afterwards and is green, so every asset in the index resolves and hashes with the new `[tool.arm-fvp]` row included — not just the three hand-verifications of that row.

Net effect on CI: the `sdk_index` job gains one 68 MB download, on index edits only.

PR #620, opened 2026-09-06
docs(#0999): file the issue phase-416 cites, and gate the archive boundary

jerry73204:
Two things with one cause: **a record that says less than the commit that made it.**

## Issue 0999

The id is reserved (`refs/issue-ids/0999`), the fix is on `main` (`c003ae608`, 2026-09-03), and **no `docs/issues/0999-*.md` was ever written.** phase-416 cites it four times, so a reader following those arrives nowhere.

Reconstructed from the fix commit and the phase doc: `preflight::check` received the **workspace** root while `config/rust-targets.txt` lives in the **checkout**, so the read found nothing, fell back to `Provisioning::Rustup`, and every build-std target was reported as an uninstalled rustup one — failing the nightly nuttx cell with `rustup target add armv7a-nuttx-eabihf` for a target that has nothing to install.

Filed as **resolved** in `archived/`, because it is. Its sibling `#0998` also has no file; not reconstructed here because nothing cites it.

**Correcting myself:** I said earlier that 0999 was "cited from five places". It is cited from **one file, four times**. The other four hits were the digit string — a port range (`60999`) and test-fixture ids in `gen-issue-index.py`.

## The archive-boundary gate

Issue 0937's archiving commit was **`R100`** — a 100% similarity rename, zero content change. The correction its own message described had been clobbered by a `git checkout origin/main -- <file>`, so the file sat in `archived/` for a day reading `status: open` and blaming a cause that had been disproved. `c7f2b09a0` restored it, `+81/−72`. Rename detection hides that from review: a reviewer sees a delete plus an add, not a diff.

### The rule was chosen by measurement

Two better-sounding rules were rejected on the numbers:

| candidate | result |
|---|---|
| *"an archive move must not delete body text"* | fires on **127 of 668** archive renames (19%), most legitimate — an archiving commit routinely rewrites an `Options` list into a `Resolution`. A gate firing on a fifth of a normal operation teaches people to ignore it. |
| *"the archived content equals an older committed version"* | fires on **zero of 668**, including 0937 itself, and costs **27 s**. It cannot work — the reverted text was the *first* version, so there was no older one to match. |

What an `R100` rename **necessarily** leaves behind is frontmatter that still says `open`. One cheap tree-state read, firing exactly on the damage.

### Mutation-tested against the real thing

Restoring 0937 to its state at `4f1305788` reds the gate:

```
0937-generated-entry-omits-global-allocator.md
    frontmatter says `status: open` — an archived issue is not open
```

`resolved_in:` is deliberately **not** required: 648 of 991 archived issues lack it, so requiring it would be a 648-entry ratchet that says nothing about whether the record is honest.

### The two files it found, of 991

- **0364** — no frontmatter at all (it predates the convention) while its own body reads *"**Status:** Resolved (phase-326, `5f784d5da` …)"*. The frontmatter now says what the body already said; nothing invented.
- **0203** — archived with `status: open` and no resolution recorded anywhere in the file. Either it was fixed and the record was never updated, or it is still live and sits in the wrong directory. **Baselined with that reason rather than guessed at** — inventing a status for it would be the 0937 defect committed deliberately. The baseline may only shrink, and a stale entry fails the gate.

`archived-issue-status`, `gate-lists`, `gate-selftests`, `default-gates-run-somewhere`, `lane-contracts`, `just-recipe-refs`, `issue-ids`, `issue-index`, `doc-refs`, `markdown-links`, `book-links` all pass.

### DIFF
--- a/docs/issues/archived/0203-mixed-workspace-cpp-interface-overgeneration.md
+++ b/docs/issues/archived/0203-mixed-workspace-cpp-interface-overgeneration.md
@@ -1,10 +1,11 @@
 ---
 id: 203
 title: "mixed-workspace cpp codegen over-generates the interface set — cross-language service pairs blocked"
-status: open
+status: resolved
 type: bug
 area: codegen
 related: [phase-263, phase-269, rfc-0026]
+resolved_in: "e2c413582 (2026-07-16)"
 ---
 
 ## Summary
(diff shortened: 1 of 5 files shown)


jerry73204:
**0203 is resolved, not baselined** — the gate's stale-entry check forced the question, and the answer was already in git.

`e2c413582` (2026-07-16): *"fix(#203): mixed cross-LANGUAGE service pair (C server + C++ client); close #203"*. That commit archived the file and never flipped `status:`. It is 0937's shape **three weeks earlier**, unnoticed until this gate went looking — which is a better argument for the gate than the one I wrote in the PR body.

**The title is a misread, and the fix commit says so.** There was no over-generation: `example_interfaces` genuinely depends on `action_msgs` (it ships `.action` files), so the resolved closure was always correct. What was broken, and only historically, was the cross-scope cpp compile — fixed en route by phase-263 A4's `NanoRosGenerateInterfaces` repairs and phase-269's header-mirror work. The issue's "Direction" section proposes two fixes for a defect that turned out not to be the one that existed; I left it in place with the correction beneath it rather than rewriting history.

Its own Repro sketch asked for *"a cpp service pkg (e.g. a `cpp_add_client_pkg` mirroring the C one)"*. That package is in `examples/workspaces/mixed/src/` today, paired with `c_add_server_pkg` in `demo_bringup/system.toml`.

**One thing worth knowing, recorded in the file:** the compile is still guarded — the mixed workspace has fixture rows and two `Lang::Mixed` cells in `matrix::CELLS`. But the three runtime tests that commit cited as its verification (`mixed_service_roundtrip_xprocess_e2e`, `mixed_action_roundtrip_xprocess_e2e`, `mixed_multihost_e2e`) no longer exist, and the surviving mixed cells are `EntryPubsub`. So the cross-language service *round trip* has no cell. Not a regression of #203, but the file's "verified" line shouldn't be read as current.

**`BASELINE` is now empty**, which is the intended end state. Leaving 0203 in it would have failed the gate's own stale-entry check — that is what made this land rather than sit behind an exemption. The self-test drives a synthetic entry so the exemption mechanism stays covered rather than untested-because-unused.

Both known cases mutation-tested against the empty baseline: restoring 0937 to `4f1305788` and 0203 to its pre-fix state each red the gate.

PR #631, opened 2026-09-06
docs(#1163): compile-smoke checks nros-c in a shape nothing ships

jerry73204:
Files issue 1163. PR #329 landed `nros_node_get_fully_qualified_name` defined twice in `packages/api/nros-c/src/node.rs` and the required `CI` context stayed green: `compile-smoke` and the queue's `test-unit` compile `nros-c` under its default feature only, and the duplicate lives in an RMW-gated region — the shape `check-c` builds and every consumer selects with `default-features = false`. `check-api-parity` refused the header at the right layer, but has reported post-submit only since issue 1040.

Measured on main `2bcdfa5d0`: default and `std` green; `rmw-cffi`, `std,rmw-zenoh`, and `check-c`'s `std,rmw-cffi,platform-posix,ros-humble` all E0428. Symptom fix is #612; this issue names the gate: the shipped set in `compile-smoke` (read from `check-c`'s one spelling), `api-parity` on the PR and queue lanes once #557's clang image lands, and a `check-lane-contracts` rule for the class.

### DIFF
--- a/docs/issues/1163-compile-smoke-checks-nros-c-under-default-features-only.md
+++ b/docs/issues/1163-compile-smoke-checks-nros-c-under-default-features-only.md
@@ -5,7 +5,7 @@ status: open
 type: bug
 area: ci, c
 severity: high
-related: [1040, 1059, 1080, 1102, 0417, RFC-0061]
+related: [1040, 1059, 1080, 1102, 1162, 0417, RFC-0061]
 found: 2026-09-06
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #678, which carries these two commits (identical patch-ids) plus the compile-smoke fix and the lane-contracts rule the issue asks for.

PR #639, opened 2026-09-06
fix(#1167): the guard 1015 added sits above the knob it guards

jerry73204:
**A red on `main`, introduced today.** Every Zephyr zenoh build fails to compile:

```
zpico.c:386:2: error: #error "ZPICO_MAX_SESSIONS must be >= 1: it sizes a fixed C array (issue 1015)"
```

## The defect

The C preprocessor evaluates an **undefined** identifier in `#if` as `0`. Issue 1015's guard sits at line 385; `ZPICO_MAX_SESSIONS`'s `#ifndef`/`#define` default sat at line 484. Nine sibling guards are fine because the defaults block is above them:

| knob | guard | default | |
| --- | ---: | ---: | --- |
| `ZPICO_MAX_QUERYABLES` | 373 | 205 | ok |
| `ZPICO_MAX_PENDING_REPLIES` | 382 | 265 | ok |
| **`ZPICO_MAX_SESSIONS`** | **385** | **484** | **guard first** |
| `ZPICO_MAX_PUBLISHERS` | 400 | 199 | ok |

**It is exactly this knob because it is the one guarded knob with no Kconfig row and no cmake bridge** — `nros-zpico-build/src/runner.rs` says so in as many words. Every other guarded knob arrives as a `-D` on the Zephyr C lane's command line, so the ordering never mattered for them. This one does not, so the in-file default was the only definition the TU would ever get, and it came ~100 lines too late. The compile line carries `-DZPICO_MAX_QUERYABLES=8`, `-DZPICO_MAX_PUBLISHERS=8`, `-DZPICO_MAX_SUBSCRIBERS=8`, `-DZPICO_MAX_LIVELINESS=16` … and nothing for sessions.

Both facts are correct in isolation and only wrong together, which is why reading the diff does not catch it.

## Why it reached `main`

Three independent gaps, all open:
- no merge-gating event builds Zephyr;
- the tier-2 lane has produced no runtime verdict in six consecutive runs (#1158);
- the one cortex-m Zephyr build that did run on 2026-09-06 died earlier, on the cross-checkout fault (#1166).

## Fix

The default moves up beside its neighbours, with a comment at both ends saying the order is load-bearing.

**Gate `check-c-knob-guard-order`** (fast lane, buildless, self-testing with six cases): every `#if KNOB < 1` guard must have the knob defined earlier in the same file, or unconditionally by a tracked header under the component's `include/`.

## Verification

**Mutation-tested against the real tree**, not a fixture — reverting the one-block move turns the gate red naming the exact site, and restoring it turns it green.

**Verified by a build, not only by the gate.** The same `west build -b mps2_an385 examples/zephyr/rust/talker` (zenoh) that produced the `#error` now links:

```
FLASH:  505 KB / 4 MB (12.33%)
RAM:    1789196 B / 4 MB (42.66%)
```

`check-fast` 248/248 green via the pre-push hook. That is the fast tier — it does not compile, which is the whole reason this class needed a build to confirm.

One thing worth noting: the first draft of this gate used `rglob` to find headers, and **`check-no-tracked-file-find` caught it** on the pre-push hook. It now goes through `git ls-files`, which is also what keeps it from walking into other checkouts' worktrees (#1157, #1166).

## Not claimed

That any of the other nine guards was ever wrong, or that `ZPICO_MAX_SESSIONS` should gain a Kconfig row — that is phase-392 W5's open question about this knob, and a separate decision.

https://claude.ai/code/session_01MyCVgGEqVRY71PYDHKmWSv

### DIFF
--- a/docs/issues/archived/1167-zpico-knob-guard-precedes-its-default.md
+++ b/docs/issues/archived/1167-zpico-knob-guard-precedes-its-default.md
@@ -1,0 +1,101 @@
+---
+id: 1167
+title: "`#if ZPICO_MAX_SESSIONS < 1` sits ABOVE the knob's default, so the guard issue
+  1015 added refuses every Zephyr zenoh build instead of protecting the array"
+status: resolved
+type: bug
+area: [rmw, build, embedded]
+severity: high
+found: 2026-09-06
+related: [1015, 0460, 1029]
+---
+
+## What
+
+Issue 1015 gave every knob-sized C array in `zpico.c` a compile-time refusal:
+
+```c
+#if ZPICO_MAX_SESSIONS < 1
+#error "ZPICO_MAX_SESSIONS must be >= 1: it sizes a fixed C array (issue 1015)"
+#endif
+```
+
+The C preprocessor evaluates an **undefined** identifier in `#if` as `0`. Nine
+of the ten guards were safe because the knob's `#ifndef`/`#define` default sits
+in the defaults block above them. `ZPICO_MAX_SESSIONS`'s default was ~100 lines
+**below**:
+
+| knob | guard | default | |
+| --- | ---: | ---: | --- |
+| `ZPICO_MAX_QUERYABLES` | 373 | 205 | ok |
+| `ZPICO_MAX_LIVELINESS` | 376 | 208 | ok |
+| `ZPICO_MAX_PENDING_GETS` | 379 | 247 | ok |
+| `ZPICO_MAX_PENDING_REPLIES` | 382 | 265 | ok |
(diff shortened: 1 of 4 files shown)


jerry73204:
Superseded. The code fix landed on `main` first as `91f3ce7c9` from another session, which moved the GUARD down rather than the DEFINE up — both correct, and rebasing the two together produced a file where main's comment ("`ZPICO_MAX_SESSIONS` is defined further down the file") had been made false by the other fix.

What this PR actually contributed is the **gate**, which `main` does not have — `91f3ce7c9`'s own commit message names the gap it leaves. Reopened as #NEW with the code change dropped and the gate mutation-tested against main's fix instead of this branch's.

PR #642, opened 2026-09-06
feat(#1015): ask the COMPILER whether each array guard can fire

jerry73204:
The compile probe, as asked. Stacked on `#637` (the static rule) which is stacked on `#636` (the fix); GitHub retargets as each merges.

`check-c-array-pool-floors` reasons about the **source**: a guard must sit after the `#ifndef` fallback that defines its knob. That catches the two shapes that shipped this week and nothing else. This asks the **compiler** — the only way to answer *"can this guard fire"* for a guard unreachable for a reason no static rule enumerates: nested inside an unrelated `#if`, disabled by a config header, or in a file the translation unit never reaches.

## Two probes per guarded knob

- compile at **defaults**, require **nothing** to fire — catches `#607`, where a guard before its `#define` read the macro as undefined and fired on every build;
- compile with **`-D<KNOB>=0`**, require **that** guard to fire — catches the first attempted fix, where a guard inside the `#ifndef` is skipped by the very `-D` it exists to catch.

Measured on the real file: **8 guards, 1 file, ~0.4 s**. Against `main` as it stands:

```
[FAIL] zpico.c fires 2 guard at DEFAULT knob values.
    First: … #error "ZPICO_MAX_SESSIONS must be >= 1 …"
```

## Where it runs — and why not the build tier you suggested

`build-serial` was my first home for it, and **`check-gate-visibility` refused it**:

> *"a gate no pull request runs is a gate that rots (issue 0981: a red sat on `main` for a day behind a green required check)"*

That lane is `schedule`/`workflow_dispatch` only. Putting a probe of a defect that shipped **this week** somewhere no pull request looks is the wrong half of the trade.

`gate.yml` provisions `--source zenoh-pico` on `pull_request`/`merge_group` at a step **before** `just check fast`, so the fast lane runs this on both merge-gating events and **skips** it on push, where the source is absent. The skip is recorded, so the lane summary says what did not run rather than implying it passed.

## One thing to weigh, stated rather than buried

**It compiles, in a lane documented as buildless.** One `cc -fsyntax-only` per guarded knob, over a source the *same job* provisioned — which is the affordability rule as `check-lane-contracts` states it (an artifact the job itself builds), and that gate passes. If it still reads as too much for the fast tier, the right answer is a dedicated pull-request step like `rustdoc-links` has, not a build tier nobody runs. Say so and I'll move it.

Non-guard compile failures are ignored deliberately: this asks one question about the preprocessor; whether the TU type-checks is `check-c`'s job, and coupling them would make this red for reasons that say nothing about guards.

## Self-test

Three synthetic files — good, guard-before-define, guard-inside-ifndef — asserting the *default* and `-DK=0` outcome for each, so the probe is shown to **distinguish** all three rather than merely to pass.

`just check fast`: 240 gates ran, 8 skipped, none failed. `gate-lists`, `gate-visibility`, `gate-selftests`, `default-gates-run-somewhere`, `lane-contracts`, `just-recipe-refs` all pass.

### DIFF
--- a/just/check/abi.just
+++ b/just/check/abi.just
@@ -240,3 +240,32 @@ weak-symbols:
 # The fast source-level half is `weak_symbol_audit.rs` (in `just test`).
 weak-symbols-image:
     @bash scripts/check-weak-symbols-image.sh
+
+# The BUILD-tier half of `check-c-array-pool-floors` — issue 1015.
+#
+# That gate is on the fast line and reasons about the SOURCE: a guard must sit
+# after the `#ifndef` fallback that defines its knob. This one asks the
+# COMPILER, which is the only way to answer "can this guard fire" for a guard
+# unreachable for a reason no static rule enumerates.
+#
+# It needs the zenoh-pico SOURCE, and where that is provisioned decides where
+# this runs. `gate.yml` provisions `--source zenoh-pico` on
+# pull_request/merge_group at a step BEFORE `just check fast`, so the fast lane
+# runs it on both merge-gating events and SKIPS it on push, where the source is
+# absent. That skip is recorded, so the lane summary says what did not run.
+#
+# NOT `build-serial:`. That lane is schedule/workflow_dispatch only, and
+# `check-gate-visibility` refuses a gate no pull request reaches -- "a gate no
+# pull request runs is a gate that rots (issue 0981)". Putting a probe of a
+# defect that shipped THIS WEEK somewhere no pull request looks would have been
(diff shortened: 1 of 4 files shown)


jerry73204:
Collapsed into #637, which now carries both commits against `main` — the stack could not be armed until each base merged, which was the whole cost. Nothing dropped: this commit is cherry-picked there with its `-x` trailer, and both gates are mutation-tested on the combined branch.

PR #645, opened 2026-09-06
fix(#1038, phase-413 W2): bake the zenoh ROS router the freertos/threadx_linux nightly cells need

jerry73204:
phase-413 W2. Both cells report `Real failures: 0 / 0` with all 9 e2e cases skipped on one shared precondition — which `_check-skip-budget` correctly refuses to call a pass.

**Cause, confirmed in #1038:** the CI image is `FROM ros:humble-ros-base` plus an apt list whose only ROS entry is `example-interfaces`. A built image has **no** `rmw_zenoh_cpp` prefix, no `rmw_zenohd`, no `libzenohc*`. No router, so `zenohd_unavailable_reason()` fails `require_e2e()` and every case skips.

## Two changes, because installing the package alone would swap a skip for a SEGV

1. **Install `ros-humble-rmw-zenoh-cpp`, with the name derived from the index.** `[prereq.ros-rmw-zenoh-cpp]` is the SSoT and the name is distro-parametric, so the Dockerfile asks `prereq-packages.py` rather than hardcoding. That needs `python3-tomli` in an earlier layer — the helper must parse the index *before* the package it names is installed, which one apt command cannot resolve.

2. **Re-capture the static `LD_LIBRARY_PATH` `ENV` block.** `rmw_zenohd` links `libzenohc.so` by SONAME, and `<prefix>/opt/zenoh_cpp_vendor/lib` is on the loader path only when `setup.bash` was sourced — which a baked `ENV` replaces. Paired with the wrong `libzenohc.so` it **SEGVs mid-startup** rather than failing to load (#0774). `AMENT_PREFIX_PATH` is unchanged; only `LD_LIBRARY_PATH` gains that directory. The Dockerfile's own comment already said to re-run its capture command after changing the ROS package set — the command and its exact one-line diff are now recorded beside the block.

## Verified by building and running the image, not by reading it

- `dpkg -l` shows `ros-humble-rmw-zenoh-cpp` and `ros-humble-zenoh-cpp-vendor` 0.1.9.
- `rmw_zenohd` and the vendored `libzenohc.so` exist at the expected paths.
- The baked env matches a real before/after `setup.bash` diff in a fresh `ros:humble-ros-base` with the same package set.
- **`rmw_zenohd` starts a router with no SEGV** — #0774's failure mode not happening.

**Not verified:** the nightly cells themselves, which need the published image and the self-hosted runner. W2's acceptance line ("each lane green at least once, and the date recorded here") stays unticked.

`ros-humble-example-interfaces` beside it stays a literal deliberately — it is not an indexed prereq, so there is nothing to derive it from.

### DIFF
--- a/ci/docker/ci-base/Dockerfile
+++ b/ci/docker/ci-base/Dockerfile
@@ -63,22 +63,60 @@ ENV DEBIAN_FRONTEND=noninteractive
 # holds for anything installed into the single `/opt/ros/humble` prefix (the apt
 # packages below are), but a ROS package that installs to a DIFFERENT prefix, or
 # one whose own setup extends these variables further, would need this block
-# updated — `ros-humble-rmw-zenoh-cpp` adding
-# `<prefix>/opt/zenoh_cpp_vendor/lib` to `LD_LIBRARY_PATH` is the case issue
-# 0774 is about. Re-run the capture command above after changing the ROS package
-# set rather than assuming.
+# updated. `ros-humble-rmw-zenoh-cpp` is exactly that case (issue 0774 /
+# issue 1038): it vendors zenoh into `<prefix>/opt/zenoh_cpp_vendor/lib`, an
+# ADDITIONAL directory under the same prefix, and `rmw_zenohd` links its
+# `libzenohc.so` by SONAME — paired with the wrong one, it SEGVs mid-startup
+# rather than failing to load. `AMENT_PREFIX_PATH` does NOT change (still one
+# prefix); only `LD_LIBRARY_PATH` gains that directory. Re-captured 2026-09-06
+# after adding the package below, with the same command:
+#
+#     docker run --rm --entrypoint bash ros:humble-ros-base -c \
+#       'apt-get update -qq && apt-get install -y --no-install-recommends \
+#          ros-humble-rmw-zenoh-cpp ros-humble-example-interfaces \
+#          >/dev/null && env|sort >/tmp/a; . /opt/ros/humble/setup.bash; \
+#          env|sort >/tmp/b; diff /tmp/a /tmp/b'
+#
+# which prints exactly one changed line:
+#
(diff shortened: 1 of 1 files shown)


jerry73204:
Closing: **redundant, and I verified that against the lane rather than assuming it.**

phase-413 W2 describes the `freertos`/`threadx_linux` cells as skipping all 9 e2e cases on the router precondition, and prescribes two changes to the CI image. That symptom is **already fixed**, by a different mechanism that landed on 2026-09-05: `575dc4372` (phase-413 W3) added *"Provision the declared system closure"* — `nros setup --system --sudo` — to nightly's `platform` job, which is the job those two cells run in.

Measured on last night's run (07:10Z):

```
freertos       Real failures: 1 / 1   check-skip-budget: 8 ran, 0 skipped for an unmet precondition
threadx_linux  Real failures: 1 / 1   check-skip-budget: 8 ran, 0 skipped for an unmet precondition
```

Both went from "0/0, everything skipped" to **8 tests actually running**, and the system-closure step reports `success`. The lane has signal capacity again; what remains is one genuine failure per cell (`test_rtos_pubsub_e2e … Freertos/Rust`), which is a real bug to chase rather than a provisioning hole.

Baking the package into the image would still be *faster* than a per-run `sudo apt`, but it would be a **second mechanism installing the same package** — and the runtime path is the better of the two on this campaign's own terms, because it takes the name from the index at run time instead of freezing a copy into an image that then has to be rebuilt when the index moves. One path, which is what phase-422 spent eight work items establishing.

The ENV re-capture half is also not needed for these lanes: `zenohd_router.rs`'s `paired_zenoh_library_dir` pins the router to its own `libzenohc.so` for the process it spawns, so issue 0774's trap does not apply to a harness-started router. It still applies to one started **by hand** in that container — that is 0774's remaining edge, and it is documented in nightly.yml rather than fixed here.

Work not wasted: the finding that these cells are already recovered is what updates W2, and it was only visible because someone went looking at the actual lane.

PR #648, opened 2026-09-06
fix + phase-436: release jitter judged against the declared cadence, and the deadline-driven executor (W1–W5)

jerry73204:
> **Scope changed after opening.** #691 was stacked on this branch and has merged into it, so this PR now carries **both** the original jitter fix **and** all five phase-436 work items — 768 insertions across 7 files. The title and this description have been updated to match; the original text is preserved below.

## Part 1 — the original fix: cadence is declared, not inferred

`spin_once(timeout)`'s argument is a blocking **bound**, not a cadence. nros-cpp's tier loops pass a hardcoded 10 ms and then pace themselves with `platform_sleep_us(period_us)`, so `release-jitter-runtime` judged every tier against 10 ms whatever `spin_period_us` the contract declared. **A 1 kHz tier could run nine periods late and register as on time.**

Passing the period to `spin_once` would be wrong — the loop already sleeps the period afterwards, so the effective cadence would become up to **2×**. That changes production pacing to fix a measurement.

Instead the cadence is declared: `set_spin_nominal_us()`, preferred by the rule, falling back to the timeout when unset. Both loops declare theirs — the boot thread and the tier trampoline, which had the identical bug.

This corrects the claim in #379/#418, which said the rule measured the tier loops' actual cadence. It did not.

## Part 2 — phase-436 W1–W5 (was #691)

The executor asked *"how long may I block?"*. It now asks *"when is the next thing I owe anybody?"*

| | | |
|---|---|---|
| **W1** | #1192 | park bounded by the next **timer** deadline; `WakeSourceId` attribution |
| **W2** | #1193 | budget carried in µs; the park rounds **up**, never truncates to a spin |
| **W3** | #1194 | jitter states the granularity it was measured at |
| **W4** | #1195 | a poll never overshoots the caller's deadline |
| **W5** | #1196 | unbounded `condvar_wait` marked and confined by a gate |

**W1 nearly introduced a regression, caught by its own test.** Capping the park by the timer deadline meant issue 0515's quantization audit was auditing the *capped* value — comparing the timer period against itself, an exact multiple every time. The warning would have gone silent on exactly the images that have a timer to quantize. The audit now runs against the declared cadence.

**W5 corrected one of my own issues.** Issue 1196 claimed the unbounded `condvar_wait` had no callers. It has two — the grep behind that claim searched `nros_platform_cond_wait`, and the symbol is `nros_platform_cond**var**_wait`. A clean false negative. Both callers are correct (they bridge zenoh-pico's own unbounded primitive), so the fix became "confine by path, mark the header" rather than "delete".

## Verification

- `cargo test -p nros-node --features std`: **359 passed, 0 failed**
- Default-feature `cargo test`: **20 errors, identical to the base** — this adds none
- `cargo fmt --check` clean; `check-no-unbounded-condvar-wait` OK
- Every new test confirmed RED first. The W1 wiring test, the audit-regression test, and both W4 tests were each re-run against the reverted change and confirmed to fail.

## Not attempted

A `wake_wait_ns` platform slot (would make five ports grow a required symbol; the rounding contract fixes #1193's actual harm without it), and making the wait loops wake off the backend's listener rather than a fixed grid (#1195's larger half). Both recorded in the phase doc.


jerry73204:
Dequeued: this conflicts with #529, which is ahead of it in the merge queue. Both add an export to `nros-cpp/src/lib.rs` at the same anchor and both regenerate `nros_cpp_ffi.h`, so the queue's merge group can't build.

No conflict with `main` itself — a trial merge onto `main` is clean. Will rebase onto `main` and regenerate the header once #529 lands, then re-arm.

jerry73204:
Rebased onto `main` and the conflict with #529 is resolved at the source rather than by serialising the queue: both PRs were inserting a new export immediately after `nros_cpp_bind_group_sched`, an adjacent-insert conflict. `nros_cpp_executor_set_spin_nominal_us` now sits beside `nros_board_native_run_tiers`, next to the two tier loops that call it — which reads better anyway. Trial merge against #529's head is clean, so the two can now land in either order.

Also: `cargo test -p nros-node --features std` is now **341 passed, 0 failed** — the `backing::tests::the_static_is_handed_out_once_and_only_once` red this PR was reporting as pre-existing is gone, because #672 fixed it and the rebase picked it up.

jerry73204:
Rebased onto `main` (was `92c3f3799`, now `6dad26c93`). 6 commits before, 6 after; nothing dropped.
The `comm` no-loss check reports one line, and it is a RETITLE of a commit that is still here, not a
loss — see the `#1196` item below.

## Conflicts and resolutions

**1. `packages/core/nros-node/src/executor/spin.rs` — phase-427 vs phase-436 W2, both live.**

`main` split `spin_once` into a thin wrapper over a new
`spin_once_capturing(timeout, first_error: &mut Option<TransportError>)` (phase-427, so
`SpinOptions::stop_on_first_error` can return the error rather than a count). This branch changes
the same first line of that body from `as_millis()` to `as_micros()` (W2 / issue 1193 — the
truncation that made `from_micros(500)` a zero timeout, i.e. a 100 % CPU loop). Neither supersedes
the other: kept `main`'s split, and the microsecond budget now opens `spin_once_capturing`. The
rest of the branch's body (`timeout_us` through `next_wake_bound_attributed_us`, the park, the
`park_achieved_us` → `timeout_ms` conversion at the transport call) auto-merged and is unchanged.

Also corrected the doc comment directly above, which `main` wrote and which this branch falsifies:
it said `timeout` is *"saturated at `i32::MAX` ms (~24 days) for the underlying transport call"*.
That is the pre-W2 statement. It now says the budget is carried in microseconds to the park and
converted (rounding up) once, at the park. Leaving it would have been a doc that contradicts the
line under it.

`cargo check -p nros-node --features std --all-targets` clean.

**2. `docs/issues/1196-*` — an issue-id collision, resolved by DELETING our copy.**

`just check issue-ids` was RED: `id 1196` named two files. A parallel session filed the same issue
against the same id and landed it on `main` first
(`1196-platform-condvar-wait-is-unbounded.md`), which is the file **this branch's own roadmap
already links** — so `docs(#1196)` was filing a document that existed.

Dropped ours rather than renumbering. `just issue-new` reserves an id so the second filing of a
*different* problem gets its own; this is the same problem, and two documents for one is worse
than the collision. `main`'s version is also the better text — it carries the 2026-09-07 correction
(the "no callers" claim came from grepping `nros_platform_cond_wait`, which is not the symbol) and
the mutex carve-out checking priority inheritance on all three ports. The one section ours had and
`main`'s did not — **what would close this** — was folded into `main`'s file, so the commit still
carries content and is retitled accordingly. `just check issue-ids` → `[OK] ... (1211 issues)`.

## Gates

`check doc-refs` **OK** · `check issue-ids` **OK** · `check prose-issue-refs` **OK** ·
`check doc-recipe-refs` **OK** · `check-doc-commit-citations.py` **OK** ·
`cargo +nightly fmt --check --all` **OK** · `cargo check -p nros-node --features std --all-targets`
**OK**.

`just check fast`: **284 gates, 3 failed.** One is environmental; **two are pre-existing branch debt
that I did not introduce and am not fixing without your call.**

### Not from the rebase — reproduced on the pre-rebase tip

I checked out `92c3f3799` (the tip this PR had before I touched it) in a clean worktree and ran
both gates there. Both fail identically. `gh pr checks 648` also shows the `check` job red on the
previous run, 20m11s, before any of my pushes.

**`check-platform-abi-mirror` — and this one is a design call, so I stopped.**

    drift detected between .../include/nros/platform.h and nros-platform-cffi/src/lib.rs
      missing from unsafe extern "C" block:      nros_platform_wake_park_until_us
                                                 nros_platform_wake_park_granularity_us
      missing from nros_platform_export_*! macro emission:   (the same two)

W7 added both symbols to `platform.h` and wired Zephyr's C port, but neither reaches
`nros-platform-cffi` — no `generated.rs` entry (so `scripts/gen-abi-bindings.sh` was not re-run;
CLAUDE.md's "header edit ⇒ regenerate + commit both"), and no `nros_platform_export_*!` emission.
`PlatformThreading` has no `wake_park_until_us` / `wake_park_granularity_us` method at all.

The judgment I will not make for you: W7's message says both are **OPTIONAL** — *"a port that
cannot beat a millisecond need not define them and keeps the path it already had"* — and the
export macro is the opposite shape: it emits a definition for EVERY Rust-side port that invokes it.
So satisfying this gate means picking one of

  * a defaulted `PlatformThreading` method pair whose default says "unsupported" (the macro then
    defines the symbols everywhere, and optionality becomes a runtime answer rather than a link-time
    one — which may be exactly right, since the C side already probes `has_park_primitive`), or
  * teaching `check-platform-abi-mirror` an OPTIONAL-symbol concept (it has no exemption mechanism
    today; `HEADERS_EXTERN_ONLY` is per-header, not per-symbol).

Both are real changes to this phase's ABI story, and picking wrong is the kind of thing this
repo's own gates exist to prevent me from doing quietly.

**`check-api-parity` — 8 ledger rows owed by W6/W7.**

    8 item(s) differ with no ledger entry.
      rust:Executor::has_park_primitive              rust:Executor::park_granularity_us
      rust:Executor::last_park                       rust:Executor::register_wake_source
      rust:Executor::last_park_achieved_us           rust:Executor::release_jitter_granularity_us
      rust:Executor::set_park_primitive              rust:Executor::set_spin_nominal_us

Green on plain `origin/main` (exit 0). All eight are the phase-436 seam's ours-only surface, so the
verdict is almost certainly `extension` — but the ledger's `why` fields are authored rationale, not
a classification checkbox, and CLAUDE.md's own record is that an AUTHORED parity map drifting into
plausible-but-wrong is the failure mode there. Eight of those written by someone who did not design
the seam is worse than none.

**`check-xrce-vendored-versions`** — environmental: the vendored XRCE trees are not checked out in a
fresh worktree. Reproduced identically on a plain `origin/main` worktree.

State: **MERGEABLE**, auto-merge armed. It will not leave the queue until those two lanes are green.

jerry73204:
Cleared one of the two red gates. The other is a design call and I have left it.

## `check-api-parity` — fixed, and the verdict was measured not chosen

W6 and W7 add eight `Executor` methods with no ledger rows, so the gate reports them UNLEDGERED. Red here, green on `main` — it belongs to this branch.

The verdict is settled by the same measurement the `Executor::release_jitter` pair beside them already used: searched the recorded upstream surfaces (`docs/reference/api-surface/{rclc,rclcpp,rclrs}.json`) for `park`, `wake_source`, `spin_nominal`, `granularity` — **zero hits in any of the three**. No upstream name to align to is what `extension` means, and the two siblings on this same surface are already `extension` for that reason.

The `why` fields are condensed from **the doc comment already above each method**, not written fresh. A ledger row owes the rationale the code states; inventing a second one is how a ledger starts disagreeing with the tree it describes. So `park_granularity_us` carries the hardcoded-`1_000` history (right for FreeRTOS by coincidence, wrong for the other two, and it made the W3 jitter report over-claim on ThreadX — W3's own failure mode reproduced one layer below it); `has_park_primitive` carries why an unbreakable park makes latency *worse*; `register_wake_source` carries why it refuses past capacity rather than dropping silently; `set_spin_nominal_us` carries the nros-cpp tier case where pacing and blocking bound are different numbers.

`just check api-parity` now prints "every divergence carries a ledger entry". Duplicate-key scan across all 17 shards with an `object_pairs_hook` counter: clean — that is the failure a hand-edited shard invites, so it is checked rather than assumed.

## `check-platform-abi-mirror` — left for whoever owns the seam

```
missing from unsafe extern "C" block:
  - nros_platform_wake_park_granularity_us
  - nros_platform_wake_park_until_us
missing from nros_platform_export_*! macro emission:
  - (the same two)
```

This is a genuine design question, not a mechanical gap. W7 calls both symbols **OPTIONAL** — *"a port that cannot beat a millisecond need not define them"* — while `nros_platform_export_*!` is all-or-nothing: it defines the symbol for every Rust-side port.

Two ways to close it, and both change this phase's ABI story:

- **A defaulted pair.** Attractive because the semantics are already written — `has_park_primitive` exists, and `park_granularity_us` documents its fallback as *"1 ms, which IS the honest floor of the `wake_wait_ms` ABI every port exports today"*. But it turns optionality into a runtime answer, and an exported default risks colliding with a C port that does define the symbol.
- **Teach the gate a per-symbol OPTIONAL concept.** It has none — `HEADERS_EXTERN_ONLY` is per-header.

Both were reproduced on the pre-rebase tip `92c3f3799` in a clean worktree, and `gh pr checks 648` shows `check` red on the previous run too, so neither is rebase fallout.

jerry73204:
Recovered a **stranded commit** that answers the design question this PR was blocked on.

## Both red gates are now green

`01c3602d9 fix(phase-436 W7): mirror the optional park symbols through the platform ABI` was pushed to #814's branch at 11:27 — **after** #814 had already been squash-merged into this branch at 03:09. So it sat on no branch and no PR: not on `main`, not here, and #814 is closed. Verified both directions with `git merge-base --is-ancestor` before touching it. Cherry-picked here, where its subject says it belongs.

It is the author's own answer to the question I left open, and it picks the option I was not willing to pick for them:

> Both trait methods carry DEFAULTS, so no port is forced to implement a primitive it cannot offer — `wake_park_until_us` defaults to `-1` ("cannot park", leaving the executor on the millisecond path it already had) and `wake_park_granularity_us` to `0` ("no finer park offered"), which is already the executor's "fall back to the ABI floor" sentinel. Only Zephyr and ThreadX override; the other four ports are unaffected.
>
> That default is what makes these symbols genuinely optional rather than a new obligation on every port, which is the property the header claims for them.

That resolves the tension I described: optionality survives because the default is a *runtime* "no", and `-1` / `0` are sentinels the executor already reads, so nothing new is imposed on the four ports that do not override. Four sites move in lockstep — the `PlatformThreading` trait, the `generated.rs` extern block, `CffiPlatform`'s forwarding impl, and the `nros_platform_export_*!` emission.

**Verified on this branch, not taken from the commit message:**

```
platform C ABI mirror clean: 98 symbols total across 3 headers
platform crate macro invocations clean: 3 crates
```

`just check api-parity` OK (the eight W6/W7 ledger rows added earlier close that half), `cargo test -p nros-node --lib --features std` — 384 passed 0 failed, `cargo check -p nros-platform-cffi --all-targets` clean, `cargo +nightly fmt --check --all` clean.

No-loss check against the recorded remote sha `6911afb0e`: empty. 8 commits → 9.

Nothing in this PR now needs a decision from you.

jerry73204:
Rebased onto `main` (`e802a796f`). BEFORE `710e7e610` (12 commits) -> AFTER `6538ed221` (13, one added below).

**No loss.** `comm -23` of the two subject sets over `origin/main..` is EMPTY — all 12 replayed, nothing dropped.

**The two fixes applied earlier in this session survived the rebase**, checked by hand: the `port_park` re-export in `packages/api/nros/src/lib.rs` still carries `#[cfg(feature = "rmw-cffi")]` on the line above it, and `cargo +nightly fmt --check --all` is clean (the `executor/storage.rs` rewrap).

## The one conflict

`docs(ledger): the eight W6/W7 executor rows check-api-parity was owed` collided in `docs/reference/api-parity-ledger/exec.json`. Resolved with a row-wise 3-way merge off git's index stages rather than a side of the line diff: **25 rows only `main` had touched, 8 rows only this branch adds, 258 untouched, and ZERO rows both sides edited** — so nothing needed a judgement call. All 17 shards re-parsed afterwards with an `object_pairs_hook` duplicate counter (git's line merge can create duplicate JSON keys in files it auto-merges, and JSON keeps only the last): no duplicates, all 17 parse, and the result round-trips byte-for-byte through `json.dumps(indent=2, ensure_ascii=False)`.

## One new commit — a real defect the rebase surfaced

`fix(phase-436 W7): the W7 ABI verbs were inserted INSIDE another function's attributes`

`just check cpp` was red, and not for a formatting reason. The three new C-ABI verbs (`register_wake_source`, `set_park_primitive`, `set_spin_nominal_us`) were inserted between `nros_board_native_run_tiers`'s doc comment and the function it documents. A doc comment IS an outer attribute, so the insertion split `run_tiers`'s attribute list, and the detached `#[cfg(all(feature = "rmw-cffi", feature = "env"))]` re-bound to the NEXT item. Introduced by this branch's first commit (`b37257869`) and grown by W6/W7; a second `#[cfg(all(rmw-cffi, env))]` was added lower down so `run_tiers` still compiled, which is what kept it looking fine.

Three consequences, one of them functional and one of them shipped:

1. **`nros_cpp_executor_register_wake_source` silently required `env`.** Measured with a three-line control, not argued:

   ```
   #[cfg(all(feature_a, feature_b))]
   /// doc
   #[cfg(feature_a)]
   pub fn victim() {}
   pub fn user() { victim(); }
   ```
   `rustc --cfg feature_a` -> `E0425: cannot find function victim`, with `note: found an item that was configured out`. With `--cfg feature_a --cfg feature_b` it compiles; with the `#[cfg]` reattached to the fn and only `feature_a`, it compiles. `env` is the process-ENVIRONMENT capability, and W7.a exists precisely so FreeRTOS, NuttX and the C arm of Zephyr can reach the wake-source seam through this ABI — none of which build with `env`. The verb was absent from exactly the images it was written for.

2. **The contamination shipped in the C header.** cbindgen emits whatever doc is attached, and `nros_board_native_run_tiers` is not exported into `nros_cpp_ffi.h` at all (`git show HEAD:...nros_cpp_ffi.h | grep -c nros_board_native_run_tiers` -> `0`), so the committed header carried 21 lines about spawning tier threads immediately above `nros_cpp_executor_register_wake_source` — describing a function a C reader cannot call and has no other reference to. `just regen-c-headers` removes exactly those 21 lines and nothing else.

3. `run_tiers` lost its `# Safety` section (`clippy::missing_safety_doc`), and the orphaned `#[cfg]` + blank line is `clippy::empty_line_after_outer_attr`. Those were the two `-D warnings` errors.

The fix is placement only — the doc block moves back onto `run_tiers`, the orphaned duplicate `#[cfg]` and its blank line go, and a duplicated `#[cfg(feature = "rmw-cffi")]` in front of `set_spin_nominal_us`'s doc goes with them. No signature, body or feature-list change.

## Gates

| gate | result |
| --- | --- |
| `just check fast` (290 gates) | **1 FAIL: `xrce-vendored-versions`** — environmental, see below |
| `just check platform-abi-mirror` | `98 symbols total across 3 headers`, clean |
| `just check issue-ids` | OK (1219 issues) |
| `python3 scripts/check-doc-commit-citations.py` | OK — 207 citations resolve |
| 17-shard duplicate-key scan | no duplicates |

`api-parity` and `platform-abi-mirror` were the two that had been red on this branch earlier in the session; both are green here, re-verified after the attribute fix as well as before it.

`xrce-vendored-versions` is NOT this branch: it needs `packages/rmw/xrce/xrce-sys/{micro-xrce-dds-client,micro-cdr}` checked out via `nros setup --source`, which a fresh worktree has not run. Reproduced identically on a control worktree at plain `origin/main` (`e802a796f`) — same three lines, same exit 1.

jerry73204:
Superseded by #849, which merged. #849 carried every commit on this branch, rebased onto main (checked by patch-id with `git cherry`), plus the fixes for why this PR was ejected from the merge queue twice. That ejection wasn't phase-436: #462's inline per-SC tables pushed the `Executor` header over its budget, and those tables now live in carved storage. #849 also fixed a C-shim link break for the optional park symbols and regenerated the hand-edited ABI bindings. Closing at the maintainer's request.

PR #686, opened 2026-09-06
fix(ci): host-tests compiled the workspace twice, and two lanes were red for process-shape reasons

jerry73204:
Three changes from a session that started at "why is `host-tests` slow" and turned into "what is this job actually compiling". Each one is measured; two things I intended to do are *not* here because measuring killed them, and those are recorded below rather than dropped silently.

## What landed

**`refactor(check)`: drop ten `--no-default-features` that opt out of nothing.** `panic-platform` has exactly one `cfg` site in the tree, gated `all(panic-platform, not(std), not(panic-halt))`, and no build script reads it. Every one of these rows names `std`, so the feature is already suppressed and dropping it changes nothing compiled — while still forking a second `-C metadata` unit for identical output. Two rows keep the flag deliberately (documented in the message). Emptying `default` instead is **not** an option: that was tried and reverted, and `check-feature-contract.py` now carries a `final-artifact-default-exempt` case with a negative control (issue 0615).

**`feat(check)`: report the gate profile the runner already measures.** `results.tsv` has always carried a per-gate duration; the summary printed one line of it and the `trap` deleted the rest. Now prints the top ten plus:

```
check-build:  busy 2510s over wall 387s at -P24  =>  6.5x effective parallelism
check-fast:   busy  154s over wall  14s at -P24  => 10.7x
```

**`fix(ci)`: host-tests built the workspace twice and capped itself at half the runner.** Two independent defects. The two fixture steps differed by exactly one thing — `CARGO_PROFILE_RELEASE_LTO: "off"` — so every crate they share was compiled twice across the ~53 of the job's ~91 minutes they occupy; it was also an env mutation of `profile.release`, so artifacts landed in `target/release/` while not being release builds. And `NROS_BUILD_JOBS: "2"` was measured on the 2-vCPU/7 GB runner of #57, which this job no longer gets — `just ci tier1` in the same job reports `-P4` because the gate runner asks `nproc`. Now derived once into `$GITHUB_ENV`, bounded on **memory** rather than CPU because #57's actual symptom was the OOM killer:

```
 2 cpu,  7 GiB -> 2    (the #57 runner: same as the literal it replaces)
 4 cpu, 16 GiB -> 4    (today)
16 cpu, 16 GiB -> 8    (memory bites first)
```

Issue #71 is unaffected — it forbids *adding* `lto = "off"` to the workspace step; this removes one.

## Two things deliberately not done

**No warm-up build before the gate fan-out.** I was going to add one, then the new profile line answered it: the `build` lane tops out near 6.5x however many workers it gets, and CI runs it at `-P4`, *below* that ceiling. The lock convoy is real but is not the binding constraint in CI — four cores is. A warm-up would have added serial wall-clock for an inactive bottleneck.

**No sccache, because the prerequisite is not met.** The repo sits at **9.77 GB of its 10 GB Actions cache limit**, and 8.7 GB of that is two `nros-cli-Linux-*` entries at 4.4 GB each — caching an ~80s build step whose own comment says the cache is unreliable enough to need re-verification. Adding sccache today evicts something, possibly `gate.yml`'s, which serves the required PR check. That cache should be shrunk first; then sccache is worth aiming at the 53 minutes of fixture builds.

## Two commits dropped during rebase, both because main got there first

**The backing latch fix.** This branch carried a commit giving
`the_static_is_handed_out_once_and_only_once` its own cargo invocation. Main
landed a better fix in the meantime: it marks the test `#[ignore]` (so the main
invocation skips it with no `--skip` needed) and runs it as `--ignored --exact`
while **asserting `1 passed`** in the output.

That assertion matters, and it caught my version. With the test now `#[ignore]`d
on main, my `--exact` row without `--ignored` reported:

```
test result: ok. 0 passed; 0 failed; 1 ignored; 367 filtered out
```

— a gate passing having run nothing, which is the vacuity class this repo gates
against. Dropped in favour of main's.

**The executor fix.** An earlier commit here fixed the `nros-c` `lending` red — the executor's cancel/spin observables were gated on `alloc` they never used. **Main landed the same fix as issue 1177 while this was in progress**, with a cleaner `HaltFlag` type alias, so my version was dropped during the rebase. Main's is the one here.

## Verification

`just check c`, `just check cpp`, `just check compile-smoke` and `just check node-std-tests` all green on the rebased tree. `node-std-tests` and `workspace-features` were also confirmed green inside a full `NROS_GATE_LANE=build` run. The px4 rows were verified by feature resolution (`--unit-graph`) rather than by building, since PX4 SITL is not provisioned locally — the delta is `panic-platform` and nothing else.

Tier run: **`just check` lanes only** — not a full `just ci gate`. Two gates are red in my worktree for pre-existing environment reasons unrelated to this branch (`submodule-pinned-locks` and `launch-resolve-builds`, both from a dirty `play_launch` pointer whose lock `--locked` refuses to update).

### DIFF
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -169,6 +169,34 @@ jobs:
           "$GITHUB_WORKSPACE/packages/cli/target/release/nros" source-stamp >/dev/null
           echo "$GITHUB_WORKSPACE/packages/cli/target/release" >> "$GITHUB_PATH"
 
+      # phase-437 — the fixture builders' rustc fan-out, DERIVED.
+      #
+      # Every consumer below used the literal "2". That number was measured on
+      # the 2-vCPU / 7 GB runner of issue #57 and this job no longer runs on
+      # one: `just ci tier1` in this same job sets no cap and its own summary
+      # reports `249 gate(s) ran at -P4`, because `run-gates-parallel.sh` asks
+      # `nproc`. So the cap was halving exactly the two steps that are most of
+      # this job's wall clock, to respect a machine that is gone.
+      #
+      # The bound is MEMORY, not CPU, and that is why this is not simply
+      # `nproc`. What issue #57 saw was the OOM killer taking rustc mid-compile;
+      # cores were never the scarce thing. ~2 GB per concurrent rustc is the
+      # rule of thumb this uses, so the runner is free to grow in either
+      # dimension and the cap follows whichever runs out first.
+      #
+      # Written to $GITHUB_ENV rather than repeated per step: the fixture builds
+      # and `just ci tier1` must agree, and three literals that must match is
+      # how they stop matching.
+      - name: Derive the build parallelism
+        run: |
+          cpus="$(nproc)"
+          mem_gb="$(awk '/MemTotal/ {print int($2 / 1024 / 1024)}' /proc/meminfo)"
+          by_mem=$(( mem_gb / 2 ))
+          [ "$by_mem" -lt 1 ] && by_mem=1
+          jobs=$(( cpus < by_mem ? cpus : by_mem ))
+          echo "NROS_CI_BUILD_JOBS=$jobs" >> "$GITHUB_ENV"
+          echo "runner: ${cpus} cpu, ${mem_gb} GiB -> NROS_CI_BUILD_JOBS=${jobs}"
+
       # phase-413 W5 — provision with the SCOPE VERB, not four hand-rolled
       # steps. `just setup native` is what a user runs, and it expands to the
       # same sequence this lane spelled out: `nros setup --build-sources` (the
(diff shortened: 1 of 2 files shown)


jerry73204:
Dropped `refactor(check): drop ten --no-default-features that opt out of nothing` — it broke `check-lane-contracts`, and the gate is right.

My argument was that `--no-default-features` is inert once `std` is named, because `panic-platform`'s only cfg site is `all(panic-platform, not(std), not(panic-halt))`. That is true of the emitted code and I verified it. It is not the question the gate asks.

Issue 1163 (`scripts/check-lane-contracts.py:665`) requires that a crate whose default features are not what ships be compiled by a merge-gating lane **in the shape its consumers take it** — `--no-default-features --features "{{C_API_SHIPPED_FEATURES}}"`. My spelling compiles a SUPERSET (the shipped shape plus `panic-platform`), which is identical today only because that one cfg site happens to be `not(std)`-gated. Add a `panic-platform` block that is not, and the superset compiles it while the shipped shape does not — a defect would pass the gate and break every consumer. That is exactly the shape 1163 was filed for: PR #329 landed a duplicate `#[no_mangle]` in an `rmw-cffi` region, `cargo check -p nros-c` green, the shipped shape `E0428`.

I had already retracted the time-saving claim for this commit — measured with `--unit-graph`, it moves one feature set to another and shares nothing. So it bought nothing and cost a landed invariant. Removed rather than patched.

`just check lane-contracts` now: `2 crate(s) whose default is not what ships (nros-c via check::compile-smoke, nros-cpp via check::compile-smoke) are compiled in their shipped shape by a merge-gating lane`.

The two remaining commits — the gate profile and the host-tests flag/parallelism fix — are unchanged.

PR #687, opened 2026-09-06
fix(phase-432 W3.1): the C emitter refuses the boards it cannot serve

jerry73204:
Top of the stack: #674 (W2.5) → #683 (W3.2–W3.5) → this.

W3.1 is "give every board a C-ABI `run_components`". Measuring what that costs found **a prerequisite the assessment missed**, and **a defect that could be fixed without waiting for it**. This PR is the second; the first is recorded, not built.

## The defect, fixed

`emit_c` hardcodes `nros_board_native_*` in every board call it renders, because native is the whole C-ABI board surface that exists — native has `run_components`, `run_components_named` and `run_tiers`; the three RTOS boards have `run_tiers` alone; ThreadX has neither.

Called with an embedded board it emitted source **naming a symbol the target does not have**: `c_freertos_one.c.golden` called `nros_board_native_run_components_named` for `board = freertos`, and `c_nuttx_tiers` called `nros_board_native_run_tiers` for `board = nuttx`. Five goldens recorded that as if it were board coverage.

It now refuses, naming the board and what would change the answer. **Exactly those five goldens moved and no others** — the check that the refusal is scoped to what it claims.

Nothing shipping changed, and that is why this survived: the dispatch already routed an embedded C entry to the C++ emitter, so only the golden harness ever called `emit_c` with an embedded board. The bytes were wrong in a file nobody compiled.

## The ThreadX `FATAL_ERROR` was the wrong prescription

The phase doc asked for a configure-time refusal, because "a ThreadX C entry silently becomes a `.cpp` with no diagnostic, which is worse than a refusal". The silence was the defect; the conclusion was not. **The routing works** — the C++ board runner drives the entry and reaches each C node through its `extern "C"` seam — so refusing would break something that ships. ThreadX is not special either: every embedded board routes the same way, so a ThreadX-only refusal would be arbitrary as well as breaking.

So `nros codegen entry` now prints why the TU is C++ and what would change it, and `entry-pack` reports `routed=1` so CMake and a human both see it. The diagnostic the item wanted, without the regression it asked for.

## The third prerequisite — measured, NOT built

W3.1's stated justification says *"every primitive it calls (`nros_cpp_init`, `nros_cpp_spin_once`, `nros_cpp_fini`) is already C-ABI and already called from C by the sibling `run_tiers.c`"*. That enumerated three primitives and **missed the two that matter**.

`run_components` also calls `nros::ok()` — through `component_spin_loop`, which is its **exit condition** — and `nros::shutdown()`. Neither is C-ABI, and neither can be fixed by *moving a definition*, which is what closed the other two prerequisites (#668, #682).

Measured: `nros::ok()` is `Node::global_initialized()` → `GlobalStorageHolder<>::initialized`, a C++ **template static emitted by the header** and COMDAT-collapsed across TUs. `nros-cpp` ships **no C++ translation unit at all** — its `src/` is Rust, the CFFI implementation; the only `.cpp` files are tests — so there is nowhere in the crate to put an `extern "C"` accessor without adding one. `main.hpp`'s own comment already said this and the assessment did not read it: the loop *"stays in the header because the per-tick yield and the `nros::ok()` check are platform / C++-global coupled — there is no Rust-side yield or shutdown primitive to move it behind the CFFI"*.

**Why `run_tiers.c` does not hit this**, which is what made the omission easy to make: its spin loops are bare `for (;;)` with no exit check — embedded firmware runs forever — so the sibling cited as proof never calls the two primitives that are missing.

Three options are written into the phase doc with their trade-offs: move the flag behind the CFFI (right shape, touches a hot header path every C++ consumer inlines); add a shipped C++ TU exporting `nros_cpp_global_ok` (ten lines, but puts a C++ TU in a crate that deliberately has none); or give the C runner a different exit condition (cheapest and worst — a second answer to "should this loop stop", this phase's own defect class).

**This needs a decision, not a move.** The ~35-50 lines the item quotes are for the runner; the runner is not the work.

`just check fast` 259/259; `cargo test -p nros-cli-core` 986.

https://claude.ai/code/session_01Vi3bKPpmrys3J3foFUH9oF


jerry73204:
**Second commit: W3.1's third prerequisite is closed — and measuring inverted the recommendation.**

I proposed (b) — a shipped C++ TU exporting `nros_cpp_global_ok` — as the unblock, with (a) as the follow-up. Two of the numbers behind that were wrong, so this is **(a), and (b) is rejected**.

- **(b) is not "ten lines".** `NanoRos::NanoRosCpp` is an INTERFACE (header-only) library, so adding a translation unit changes the library *kind* for every consumer: link lines, PUBLIC/INTERFACE semantics, installs. That is a build-system change.
- **(a)'s objection does not survive the call sites.** I said `ok()` is on a hot path every C++ consumer inlines. Every `ok()` in the tree is a spin-loop condition paired with a millisecond-scale blocking `spin_once` — a load became a call once per tick.

**(a) also turned out to be a duplicate-state fix, not just an accessor.** The context's tag already recorded whether the storage was live; `GlobalStorageHolder<>::initialized` was a second answer maintained by hand at four sites, and the two could disagree — a direct `nros_cpp_fini`, which `run_tiers.c` does, tore the context down without touching the flag, so `ok()` kept saying yes over a dropped executor. The tag is the flag now: `init*` stamps it, `fini` clears it, `nros_cpp_context_is_live` reads it, `Node::global_initialized()` is derived.

**`nros_cpp_fini` had to be hardened for that to be correct, and was wrong on both sides independently.** It blind-cast on the way in — alone among the entry points taking a `void*` executor handle, all of which go through `cpp_ctx_checked` because issue 0436 was memory corruption seen as PX4 dumping core. And it left the tag stamped on the way out, because `drop_in_place` runs field drops and `tag: u64` has none, so a finalised context still read as live. Neither is known to have been hit; a tag that outlives its object is still worse than no tag, since it turns a null check into a confident wrong answer.

**Not verified, and why.** No affordable lane links an archive carrying `nros_cpp_fini`, so there is no runtime probe for the refusal or the unstamp. Building one needs `nros-cpp` with `rmw-cffi`, which fails on the **same pre-existing feature break that has `check-workspace-features` red on main** (`TransportError::BackendDynamic` is `alloc`-gated; adding `alloc` then wants a global allocator). A Rust unit test is unavailable too — `nros-cpp` links `nros-c`, which is `no_std` panic=abort, so `cargo test` fails with "unwinding panics are not supported without std". Verified instead by `just check c`, `just check cpp`, the full fast line (259/259), and reading.

With the flag reachable from C, the ~35–50 lines W3.1 quotes for the runner itself are the remaining work — and the measurement nobody has yet is whether it links and boots on FreeRTOS.

PR #688, opened 2026-09-06
docs(phase-436): the poll/wake revision — a deadline-driven, platform-agnostic executor

jerry73204:
Review of the executor and the `spin*()` surface against four properties: a platform-agnostic core, per-platform wake sources, analyzable real-time bounds, no busy waiting.

## The finding

Three of the four are close to met. The wake machinery is already platform-agnostic and **runtime-probed** — `NodeWake` over five ports (Zephyr `k_sem`, ThreadX, FreeRTOS, POSIX, ESP-IDF), each with an ISR-safe signal — and it degrades to blocking in the transport's own `recv`, not to a spin.

The fourth is not met, for one reason. The executor asks **"how long may I block?"** when it should ask **"when is the next thing I owe anybody?"**. Today the answer is the caller's timeout narrowed by exactly one input, `session.next_deadline_ms()`. Nothing lets a **timer** deadline shorten the sleep — grep `spin.rs` for any next-timer computation and you get zero hits — so `spin_default()`'s 50 ms sleeps straight past a 10 ms timer it owns.

## The design

A `WakeSource` seam. Core members (no platform knowledge): the caller's budget, the timers, the session. Platform members registered per port: `k_poll` event sets, `epoll`, FreeRTOS queue sets, ThreadX event flags — or, on bare metal, a hardware-timer compare plus `WFI`.

The wait becomes `park_until(min over sources)`. Two consequences worth the change:

- **Bare metal becomes the primary case, not a degraded one.** "Park until the earlier of a timer compare and any interrupt" *is* `WFI`. That idiom is unreachable through a `sleep(duration)` seam and natural through a `park_until(deadline)` one.
- **The wake time becomes a function of declared periods**, so it can be enumerated statically — the property a schedulability argument actually needs.

Plus one time base in ns end to end, and a per-port `wake_granularity_ns()` so rounding is **declared and upward** instead of silent and downward.

## Issues filed

| | |
|---|---|
| [1192](docs/issues/1192-executor-wait-ignores-next-timer-deadline.md) | the wait is not bounded by the next timer deadline |
| [1193](docs/issues/1193-spin-timeout-truncates-to-milliseconds.md) | a sub-ms timeout truncates to zero and becomes a busy loop |
| [1194](docs/issues/1194-jitter-measured-in-microseconds-waited-in-milliseconds.md) | jitter measured in µs against a cadence waited in ms |
| [1195](docs/issues/1195-promise-wait-polls-instead-of-using-the-waker.md) | `Promise::wait` polls a 10 ms grid, overshoots shorter timeouts |
| [1196](docs/issues/1196-platform-condvar-wait-is-unbounded.md) | the platform's unbounded `condvar_wait` |

**1192 is deliberately distinct from resolved 0505 and 0515.** The reservation tool surfaced both, and they are close: 0505 is the overrun *policy* for a backlog, 0515 is grid *quantization*. Both take the spin boundary as fixed and describe what happens around it. This one moves the boundary. The issue says so explicitly so a reader doesn't re-close it as a duplicate.

**1193 is latent, not live** — nros-cpp clamps `max(1_000)` and ASI runs `spin_period_us = 5000`. But `spin_period_us` is a microsecond field that accepts values it cannot honour, and `Duration::from_micros(500)` → `as_millis()` → `0` → a 100 % CPU loop with no warning. The one diagnostic that would catch it, `audit_spin_quantization`, is gated `if timeout_ms > 0`.

## API review

Three units — C takes **ns** (and paces correctly in ns, then truncates at the wait), C++ takes **ms** and cannot express sub-ms, Rust takes `Duration` — all funnelling into one ms-truncating core.

And two meanings wearing one shape: `spin_once(timeout)` is a *bound*, `spin_period(period)` is a *cadence*. That confusion is what produced #648's bug. The generated workspace entry is the shape that got it right — it declares the cadence as data (`NativeTierSpec{…, 5000ull, …}`) and never calls `spin_once` — and the hand-written C++ examples are what taught the tier loops the wrong idiom.

## Also

Corrects **three** comments asserting Cyclone is poll-only. Cyclone fills the slot (`vtable.cpp:322`) with a real `dds_set_listener`/`on_data_available`. I nearly filed a sixth issue against zenoh off the same stale reasoning before checking — zenoh doesn't override `supports_wake_callback`, so it takes the `false` default and never enters the path.

## Verification

`check-issue-index` OK (120 = 120), `check-roadmap-status` OK (73 phases), `issue-ids-check` OK (1113 issues). IDs reserved through `reserve-issue-id.sh` / `reserve-phase-id.sh`. `cargo test -p nros-node --features std`: 340 passed, 1 failed — `backing::tests::the_static_is_handed_out_once_and_only_once`, the known red already fixed by #672 in the queue.

Docs and comments only; no behaviour change.

### DIFF
--- a/docs/issues/1192-executor-wait-ignores-next-timer-deadline.md
+++ b/docs/issues/1192-executor-wait-ignores-next-timer-deadline.md
@@ -1,0 +1,75 @@
+---
+id: 1192
+title: "The executor's blocking wait is not bounded by the next timer deadline, so it sleeps past a deadline it owns"
+status: open
+type: bug
+area: executor
+related: [issue-0505, issue-0515, phase-436]
+---
+
+## Problem
+
+`spin_once` bounds its blocking wait by two things and only two: the caller's
+`timeout`, and `session.next_deadline_ms()` — the *backend's* next internal
+event (lease keepalive, heartbeat, ACK-NACK).
+
+```rust
+let timeout_ms = match self.session.next_deadline_ms() {
+    Some(next) => timeout_ms.min(next.min(i32::MAX as u32) as i32),
+    None => timeout_ms,
+};
+```
+(`packages/core/nros-node/src/executor/spin.rs:6412`)
+
+Nothing caps it by the next **timer** expiry. Grepping `spin.rs` for any
+next-timer computation — `next_timer`, `earliest`, `until_next`, `next_expiry`,
+`soonest` — returns zero hits.
+
+So the executor's wake set is transport-only, while the one class of deadline
+the executor itself *owns* — its registered timers — cannot shorten the sleep.
+
+## Evidence
+
+`spin_default()` uses a 50 ms timeout (`spin.rs:7361`). Register a 10 ms timer
+under it and the executor blocks up to 50 ms in `NodeWake::wait_ms` (or in the
+transport's `recv`), during which the 10 ms deadline passes unserved. Timers
+are only evaluated in the readiness scan *after* the wait returns, driven by
+the accumulated `delta_us` (`spin.rs:6847`).
+
+The consequence at the callback is then the backlog replay that **issue 0505**
+described and fixed policy for: several activations back to back once the
+executor runs again.
+
+## Why this is not 0505 or 0515
+
+* **0505** (resolved) is the *overrun policy* — what to do about a backlog that
+  has already accumulated. It takes the stall as given.
+* **0515** (resolved) is *grid quantization* — a timer period that is not a
+  multiple of the tier's spin period alternates between two values. Its model
+  is explicitly "a timer fires on the first spin boundary at or after its
+  period elapses", and the fix was a runtime warning naming the two values.
+
+Both accept the spin boundary as fixed and describe what happens around it.
+This issue is that **the boundary should move**: a wait that knows a timer is
+due in 10 ms must not sleep for 50 ms. Fixing it removes the input to 0505's
+backlog and shrinks 0515's grid to the timer's own deadline, without replacing
+either — 0515's warning still catches hand-written loops, and 0505's policy
+still governs a genuine preemption stall.
+
+## Fix direction
+
+Compute the wait as `min(caller timeout, next timer expiry, next_deadline_ms)`.
+The timer headers are already walked once per executor in
+`audit_spin_quantization` (`spin.rs:2340`), so the traversal exists; it needs to
+become a per-spin `min` over `period_us - elapsed_us` for live timer entries.
+
+Two constraints on the fix:
+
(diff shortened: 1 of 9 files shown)


jerry73204:
Updated: issue 1196 carried a factual error, found while implementing W5 in #691.

The issue claimed the unbounded `nros_platform_condvar_wait` had no callers. It has two. The grep behind that claim searched `nros_platform_cond_wait` — the symbol is `nros_platform_cond**var**_wait`, so it was a clean false negative and a live call site got filed as latent. W5's gate found them on its first run.

Both are correct as written (`platform_aliases.c` bridges zenoh-pico's own `_z_condvar_wait`, whose contract is an unbounded wait), so the issue stays about the API and the fix became "confine by path, mark the header" rather than "delete the symbol".

The phase status line now also records what was **not** built and why — the `wake_wait_ns` platform slot is deferred in favour of the rounding contract, which fixes 1193's actual harm without making five ports grow a required symbol.

jerry73204:
Added the port survey — four parallel read-only passes, one per port domain — plus **issue 1242**.

Three claims this doc already carried were wrong, and were wrong because they came from one reading rather than a check:

**The porting axis is C entry vs Rust entry, not per-RTOS.** ThreadX, bare metal and Zephyr's Rust arm hold a typed `Executor`. FreeRTOS, NuttX and Zephyr's C arm hold an opaque `void*`, and there is no `nros_cpp_executor_register_wake_source` export — **including the arm the ASI FVP lane takes**, so the seam is unreachable from the lane we ship.

**No port multiplexes.** `k_poll` appears nowhere (the one grep hit is `nros_platform_network_poll`, a name collision with a no-op body). FreeRTOS queue sets are unused and disabled where configured. ThreadX event flags are only zenoh-pico's task-join. Every port waits on a socket with blocking `recv` + `SO_RCVTIMEO`. The "natural extra sources" in the old table were aspirations.

**The bare-metal argument stands; the claim that any board can do it today does not.** No board has a one-shot deadline-programmable timer. This doc's own `arm_timer_compare(deadline_us)` sketch presented as available something that must be written per board — corrected in place.

Two unused facilities worth having, and the cheapest real work available: Zephyr's `nros_zephyr_timer_create_oneshot` takes µs and has **zero callers**; smoltcp's `Interface::poll_at()` is never called.

**Issue 1242** — `park_granularity_us()` is a hardcoded 1 ms: wrong high on ThreadX (10 ms tick), wrong low on POSIX (timespec-native), and it never consults the installed primitive. W3's jitter report inherits that, making it the failure W3 exists to prevent, one layer below W3, in its own dependency. Both are mine.

PR #629, opened 2026-09-06
phase-428 W5/W6/W10 + Q1/Q2: the QoS table has one SSoT, and every divergence carries a disposition the gate can check

jerry73204:
## What lands

**W10 — QoS SSoT.** The eleven preset profiles are authored once, in `nros-rmw`'s `qos_profiles!` fence, and the five transcription sites (Rust core, C header, cffi, C++ header, the Zenoh shim) are checked against it by `just check qos-profile-ssot` (63-mutation self-test). `QoSProfile::build` is deleted — it hid five of ten fields. Liveliness `None` now maps to `SYSTEM_DEFAULT` in every copy, and `PARAMETER_EVENTS` is `KEEP_LAST 1000` like upstream instead of `KEEP_ALL 0`.

**W5 — 82 ledger rows** for the W1–W4 user-API audit findings; five remainder findings filed as issues 1148–1152.

**W6 — four gates**: `#[must_use]` on every Rust `bool`/`Result` that reports failure, `check-ledger-orphan-refs` (a row may not cite a file that does not exist), the correlator's return-type comparison promoted to a verdict (`same:ret` — it had never fired: `_ret_of` read a key the flattener never wrote), and `--require-disposition` extended to same-shaped divergences (Q2: 33 ported / 23 native rows authored).

**Q1 — refuse-loud audit.** 141 rows promised a compile error; 3 delivered one. 130 relabelled `absent` (the name does not exist, so nothing refuses), 8 exists-and-differs rows rewritten to what the compiler actually points at, `nros::init_with_args` made loud on `--ros-args` (mirrors the C++ `argv_has_ros_args` abort), 13 live refusals that had no row given one, and the gate now rejects a `refuse-loud` on a `theirs-only` key and any `NROS_RCLCPP_REFUSE_*` message no row cites. Census: refuse-loud 144 → 20, absent 518 → 650.

**Also**: eager-created C entities resolve their names and a failed rclcpp create aborts (`require_created`); Zenoh's `service_is_ready` returns `Err(Unsupported)` on the never-set `server_seen` latch instead of a false `Ok(false)`; Cyclone reserves the correlation slot before the destructive `take_typed_wire`.

## Rebase note

[...]

PR #704, opened 2026-09-06
feat(nros, phase-427 W9): Context on every target, and the rclrs init family

jerry73204:
Phase-427 W9, as RFC-0089 "Context and `init`, settled" states it. Stacked on `phase-428-w10-qos-ssot` (#695 merged there); retarget to `main` after #629 lands. No auto-merge.

## What landed

- **`init.rs` out from behind `env`.** `Context` compiles on every target with **`alloc` as its floor** (`locator` / `rmw` are owned `String`s — a `&'static str` cannot hold the harness-supplied override); `ContextSource`, `InitError`, `InitOptions`, `Context::config` need nothing. `env` gates only the process-environment reader (`read_env_context`, `init`, `init_with_args`, `init_with_launch*`) and `Context::new`.
- **One source of the baked shape.** `Context::baked()` is the single `option_env!("NROS_LOCATOR")` / `NROS_DOMAIN_ID` site; `from_baked(locator, domain)` is the injectable parse the unit tests exercise; `with_locator_override` carries the harness's `-testargs --nros-locator` override. Because a leaf's `cargo:rustc-env` never reaches a dependency's build, `nros/build.rs` now calls `nros_zephyr_build::bake_nros_config()` (the issue-0460 `$DOTCONFIG` channel). Both entry macros (`zephyr_component_main!`, `nros::main!`'s Zephyr arm) call `$crate::Context::default_from_env()` then `ctx.config(name)` — no `ExecutorConfig` literal. **Behaviour change:** the `main!` Zephyr arm had dropped the domain half (issue 0161's class) and ran domain 0 whatever Kconfig said; it now honours `CONFIG_NROS_DOMAIN_ID` like the C lane.
- **The rclrs family.** `InitOptions { domain_id }` (`new`, `Default`, `with_domain_id(self, Option<usize>)`, `set_domain_id`, `domain_id`); `Context::default_from_env()` (hosted = `init()`, freestanding = `baked()`); `Context::from_env(InitOptions)` with the override range-checked (`InitError::DomainIdOutOfRange`, new); hosted-only `Context::new(args, opts)` refusing `--ros-args` through the one `refuse_ros_args` site shared with `init_with_args`. `init()` stays as the C++-symmetric anchor.

## Measured

- `cargo test -p nros --features env,std --lib init`: 19 passed. `--no-default-features --features alloc --lib init`: 9 passed (the floor; `freestanding_default_from_env_is_the_bake` among them).
- `cargo clippy -D warnings`: clean for `env,std --all-targets`, `alloc`, no features, `thumbv7em-none-eabi` with `alloc,rmw-cffi,macros`, and `nros-macros`.
- **Freestanding leaf:** `examples/zephyr/rust/talker` (`zephyr_component_main!`) checked and clippy'd (`-D warnings -D clippy::undocumented_unsafe_blocks`, the west build's own flags) on `aarch64-unknown-none` with the `build-rust-talker-zenoh` configure's recorded environment, against this tree — the workspace's `nano-ros` symlink names the outer checkout, so a plain `west build` would compile the wrong tree. `nros`'s build-script output carries `rustc-env=NROS_LOCATOR=tcp/127.0.0.1:7400` / `NROS_DOMAIN_ID=0`, so the bake reaches the constructor.
- Ledger: `rust:Context::default_from_env` `rename` → `divergence`+`adopt`; `rust:init` → `extension`; `rust:Context::from_env` / `rust:Context::new` / `rust:InitOptions` → `divergence`+`adopt-bounded`; new `rust:Context::baked`, `rust:ContextSource::Baked`, `rust:Context::with_locator_override` (`extension`). All 17 shards re-parsed with a duplicate-key hook. `just check api-parity-ledger`, `ledger-orphan-refs` green; `just check api-parity` ends "every divergence carries a ledger entry"; `just check fast` green after the rebase onto `phase-428-w10-qos-ssot`.


jerry73204:
Not restacked yet, deliberately — the right moment is after #629 lands.

This PR is stacked on `phase-428-w10-qos-ssot` (#629), and that branch has been rebased twice today, so its base is no longer an ancestor: their common ancestor is `1976727d8`, an old `main` tip, and the base has since moved **110 commits** past it. This branch's 54 commits are an old snapshot of #629's line plus its own phase-427 W9 work, so a restack now replays every duplicated commit against its own rebased twin — I tried, and it conflicts immediately on `docs/design/0089-*.md` and `just/check/lanes.just` while applying `bd4b087c1`, which is one of #629's own commits.

**#629 is at merge-queue position 9.** Once it lands, the clean move is to retarget this PR to `main` and rebase onto `main`: the duplicated commits then drop by patch-id and only the genuine phase-427 W9 work is replayed. #699 did exactly that this session and went 22 commits to 1, with every drop verified present on main.

Restacking onto a branch that is about to disappear would have to be redone, so this is waiting on that rather than churning.

jerry73204:
Follow-up on the ledger judgment call the rebase flagged for confirmation — **measured, and the resolution that shipped was wrong in a way neither side was.**

The rebase reconciled two independently-authored versions of four rows. `main` said `declined`/`adopt`; this branch said `divergence`/`absent`. The resolution took this branch's side whole. Counting the tree says that combination is unprecedented:

```
across all 17 shards on main:
  ('declined',  'absent')  516
  ('divergence','absent')    0
```

Four rows on this branch carried it — `c:node_init`, `c:service_init`, `c:client_init`, `c:subscription_init`, the same four the previous commit had to de-duplicate.

**SCHEMA.md decides which half moves.** `divergence` is *"we changed it and a PLATFORM CONSTRAINT is why — `why` must name the constraint (no_std, no exceptions, no allocator, no runtime env, single-threaded transport), not a preference. This is the only sanctioned reason to differ (RFC-0036)."* These rows name no such constraint. Their reason is that phase-417 stage 6 step B moved the callback to registration, where rclc puts it — a **conformance** move toward rclc, not something a platform forced. That is `declined`: *"ROS 2 has it, we deliberately do not, with the reason."*

The disposition stays `absent`, which is where `main` was wrong. `adopt` means *"same name, same observable contract"*, and each row states in its own `why` that **no symbol answers to the key**. A porting user gets nothing; `absent` is what tells them so.

So the rows keep this branch's evidence and disposition, and take `main`'s verdict. `scripts/api-parity.py` is explicit that these answer different questions — *"A verdict says WHY we differ; a disposition says what a PORTING USER GETS"* — and the merge treated them as one choice.

`just check api-parity-ledger` OK, `just check ledger-orphan-refs` OK.

---

Separately, and worth recording because it nearly cost work: my first push of this fix **force-pushed over `96a9ded1e`** (the duplicate-key fix). `--force-with-lease` did not stop it — my worktree was created from a stale local branch ref, so the lease compared against what my clone had last fetched rather than what was on the remote. Recovered by rebuilding on `96a9ded1e` and re-applying; `git merge-base --is-ancestor 96a9ded1e HEAD` now confirms it is preserved, and the branch is 37 commits as before plus this one. `--force-with-lease` protects against a remote you have SEEN move, not one your clone never fetched.

jerry73204:
## Rebased onto `main` (`632bdbeba`)

`053922b0c` → `d9b0168e8`. **39 commits before, 40 after** — nothing dropped, one added.

**No-loss verification** (the check that matters, since `--force-with-lease` only compares against what this clone fetched):

```
comm -23 <(git log --format=%s origin/main..053922b0c | sort -u) \
         <(git log --format=%s origin/main..HEAD       | sort -u)
```
→ **empty**. No commit was dropped by patch-id or otherwise; the 39 all replayed.

### Conflicts, and what was measured

**1. `packages/api/nros-cpp/include/nros/nros.hpp`** — three hunks, all from `437945b0f` ("a failed rclcpp create aborts") landing on top of `main`'s phase-427 node merge.

*Mechanical half:* `main` deleted the shim `class Node` from `nros.hpp` — `rclcpp::Node` is now an unconditional `using Node = ::nros::Node`, and the seven `create_*` verbs are out-of-line `nros::Node::create_*` definitions. Every member the branch side re-declared (`get_name`, `get_namespace`, `get_clock` ×2, `now`, `get_logger`, `create_publisher` ×2, `create_subscription` ×2, `create_wall_timer`, `create_service` ×3, `create_client` ×3) was verified present on `main` by name before dropping. `main` wins; the shim is superseded.

*Non-mechanical half:* the branch's actual payload is behavioural, not structural — the seven verbs stop writing `(void)node_.create_…(…)` and go through `rclcpp::detail::require_created`, which aborts with `NROS_RCLCPP_ABORT_FAILED_CREATE` instead of handing back a valid-looking `shared_ptr` to a dead entity. That was **ported onto `main`'s shape**, not dropped: `abort_failed_create` / `require_created` land in the same `rclcpp::detail` block (before the `nros::Node::create_*` definitions, so the ordering resolves), the `shutdown()` ADOPT→ADOPT-BOUNDED doc revision is kept, and all seven call sites carry the refusal.

**2. An EIGHTH site, which is the one finding this rebase produced.** `main`'s phase-430 W6 landed `rclcpp::create_timer(node, clock, period, callback)` as a free function *after* the branch's sweep ran, with the same discarded `Result` (`nros.hpp:865`). Per the fix-the-CLASS rule it goes through `require_created` too — and it is the quietest member of the class, because a timer handle is stored and never dereferenced, so a dead one produces no downstream symptom at all.

**3. `docs/reference/api-parity-ledger/timer.json` — `cpp:create_timer` gained a `disposition`.** Not a conflict; a gate failure the rebase created. This branch widens `--require-disposition` to `same_shaped_divergences` (`c60bfaee8`, `05b1d566d`), and `main` supplied a subject for it: with `rclcpp::create_timer` now present with upstream's argument order and a returned `shared_ptr`, the row's shape *corresponds* on the ported-file surface, so a ported line COMPILES and nothing states the envelope — exactly the class the widening is for. Confirmed the rule is this branch's, not `main`'s: `git show origin/main:scripts/api-parity.py | grep -c 'SHAPE corresponds'` → `0`.

The value is **measured from the header, not chosen**: `main`'s own comment above the function reads *"the same ADOPT-BOUNDED trade `get_clock()` itself makes"* (`nros.hpp:815`). So `adopt-bounded`, with the `why` recording the four bounds the header already documents — `nros::Clock*` where upstream takes `Clock::SharedPtr` (no allocator, RFC-0022), millisecond truncation of the period, no trailing `group` parameter, and `shared_ptr<nros::Timer>` rather than `TimerBase::SharedPtr`. **No re-verdict**: `divergence` and its `why` still describe the native out-ref `nros::create_timer`, which is unchanged.

**4. `.config/prose-issue-ref-baseline.txt`** — resolved BY HAND, as `8ee910d3d` itself insists. `main`'s copy of this file is *sorted*, header comments interleaved among the rows: someone ran `--write-baseline` on `main` and it destroyed the CLASSIFIED block's ordering. That damage is pre-existing on `main` and was **not** repaired here (out of scope, and there is a separate branch for it); the resolution kept `main`'s file byte-for-byte and removed only the five lines this branch discharges — the two `1099` rows plus the three-line `1099 IN FLIGHT` classification note. `main`'s new `docs/issues/1231-…:1164` row is preserved.

Discharge verified rather than assumed: `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md` **is** in the rebased tree, and both citations (`phase-428-…-sweep.md:414,503`, `lanes.just:634,647`) still exist. One correction to that commit's own message: it says *"7 known in-flight, down from 9"*, which was true against its old base — `main` has since added a row, so the measured figure is now **8, down from 10**. `just check prose-issue-refs` reports `8 known in-flight`.

### Gates

| gate | result |
|---|---|
| `just check fast` | **284 / 285** |
| `just check cpp` | OK — including `failed_create_aborts: OK` (the branch's own abort probe, running against `main`'s node shape) |
| `just check api-parity-ledger` | OK (self-test 0 failures) |
| `just check ledger-orphan-refs` | OK — 17 shards, every cited path and bare filename resolves |
| `scripts/check-doc-commit-citations.py` | OK — 206 citations across 899 files |
| ledger duplicate-key scan (`object_pairs_hook` across all 17 shards) | **none** |

The one `check fast` failure is `xrce-vendored-versions`, and it is **environmental, reproduced on a plain `origin/main` worktree** rather than assumed: `check-xrce-vendored-versions: no vendored tree checked out — nothing verified`. `capability-conditionals` passed here.

State: **MERGEABLE**, auto-merge armed.

jerry73204:

`d9b0168e8` → `63e9345f3`. **40 commits before, 40 after.** No-loss verified:

```
comm -23 <(git log --format=%s origin/main..d9b0168e8|sort -u) \
# (empty)
```

Nothing dropped, nothing skipped by patch-id.

Only two conflicts, both narrow — 38 of the 40 replayed clean, including the whole `nros.hpp` create-verb block, because this branch's previous rebase had already ported those onto main's out-of-line `nros::Node` shape.

**1. `packages/api/nros/src/init.rs` — a doc line, and it is main's rename that owns it.**

phase-427 W11 took upstream's spin names and migrated every `spin_blocking` call site to `spin`. This branch's `21d535273` rewrites the same module-doc paragraph to introduce the rclrs `Context` family. The two edits overlap on one line. Resolution keeps **both** payloads: main's `Executor::spin` spelling, and the branch's new sentence naming `Context::default_from_env` / `Context::new` beside `init()` / `init_with_args`. The branch text is the one that changed here, so this is the branch's paragraph with main's verb name in it — not a choice of side.

**2. `.config/prose-issue-ref-baseline.txt` — reapplied by hand onto main's repaired header.**

main's `fix(#1241)` repaired the scrambled CLASSIFIED header, and its new `check_baseline_shape` refuses a baseline whose comments follow the first data row. `8db657ccf` was written against the scrambled file, so git's line merge produced garbage. Resolution took main's repaired file verbatim (verified byte-identical to `origin/main`'s before editing) and removed only this commit's payload: the two `…:1099` rows (`docs/roadmap/phase-428-api-porting-principle-sweep.md`, `just/check/lanes.just`) and the 1099 IN-FLIGHT note, since this branch files `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md`. main's 9999 "TWO documents cite it now" paragraph is kept. `--write-baseline` was not used — it emits rows only and would delete the header.

The resulting file is **byte-identical** to the resolution taken on #629, which carries the same commit.

#### Ledger

No ledger conflict arose in this rebase. All 17 shards nevertheless scanned for duplicate keys with a `json.load(object_pairs_hook=…)` counter, because git's line-based merge can create them silently: **0 duplicates**.

#### Gates

| gate | result |
|---|---|
| `python3 scripts/check-doc-commit-citations.py` | OK (205 citations resolve) |
| `just check cpp` | OK |

The one `check fast` failure is `xrce-vendored-versions`, environmental rather than pre-existing on this branch: reproduced identically on a plain `origin/main` worktree (`no vendored tree checked out — nothing verified`), because a fresh worktree has no `nros setup --source micro-cdr --source micro-xrce-dds-client`.

jerry73204:
Rebased onto `main` @ `cf4b0048a` (phase-438 W1). **40 commits before, 40 after** — no drops:

```
comm -23 <(git log --format=%s origin/main..63e9345f3|sort -u) \
         <(git log --format=%s origin/main..HEAD|sort -u)   # empty
```

Two files needed non-mechanical resolution; everything else replayed clean.

### 1. `docs/reference/api-parity-ledger/param.json` — ROW-WISE, then FIELD-WISE

The ledger is a flat dict keyed by symbol, so two branches adding different rows do not actually
disagree even where git's line merge reports conflicting text. Each conflict was re-merged per
row and then per field (`verdict` / `disposition` / `why` are independent), with both sides'
changes verified present afterwards.

One genuine field collision, on `why` for `cpp:Node::declare_parameter`, `cpp:Node::get_parameter`
and `cpp:Node::has_parameter`. Measured, not guessed:

* `ours` (main) = `base` + a 1124-char appended paragraph (the phase-426 W4 retraction: the
  node-local C++ store is gone; both facades now forward through `nros/node_parameters.hpp`).
* `theirs` (this branch) = `base` with **one** internal substitution repairing a citation to
  `rclcpp_compat.hpp`, which phase-417 stage 6 step A deleted.

The edited spans are **disjoint** — asserted programmatically (`ours.startswith(base)`, and
`theirs` agrees with `base` on both sides of the substituted clause) — so the merge is
`base-with-theirs-substitution + ours-appendix`: both sides kept whole, nothing paraphrased.

One judgement call inside that. The branch's replacement cited `nros.hpp:834` / `:851` / `:875`,
measured against this branch's own tree. On `main` those numbers are **stale** — phase-427 moved
`class Node`, and the out-of-line forwarders now sit at `602` / `615` / `633`:

```
$ git show origin/main:packages/api/nros-cpp/include/nros/nros.hpp | grep -n 'Node::declare_parameter\|Node::get_parameter\|Node::has_parameter'
602:template <typename T> inline T Node::declare_parameter(const char* name, T default_value) {
615:template <typename T> inline bool Node::get_parameter(const char* name, T& out) const {
633:inline bool Node::has_parameter(const char* name) const {
```

Rather than transcribe numbers that go stale on the next header edit, the citation names the
**symbols**. Same intent — say where the code actually lives now — with no drift surface. Checked
that `nros.hpp` exists on `main` and that `rclcpp_compat.hpp` does not, which is exactly what the
branch's edit asserted.

### 2. `.config/prose-issue-ref-baseline.txt` — row-wise, respecting `fix(#1241)`'s shape rule

The header auto-merged correctly (this branch's `1099` block gone, `main`'s new `1110` block
kept). Only the data rows collided: `main` **added** `docs/roadmap/phase-431-ship-the-nros-binary.md:1110`,
this branch **removes** both `1099` rows (which is what `fix(#1099)` is for). Row-wise result is
both — 10 rows → **8**, and a ratchet may only shrink ✅. Verified against `check_baseline_shape`:
zero comments after the first data row, rows sorted. Edited by hand, header above all rows.

### The duplicate-key trap

Git's line merge silently creates duplicate JSON keys when both sides add the same key at
different offsets; `json.load` keeps only the last, so a verdict disappears with no gate firing.
Every resolution was followed by an `object_pairs_hook` duplicate scan over all 17 shards. It
fired mid-rebase on `node.json: c:node_init` — repaired later in this same series by
`fix(ledger): drop 4 silent duplicate keys the rebase's line-based JSON merge introduced`.
**Final tree: zero duplicates across all 17 shards.**

### The payload `main` could have superseded, and did not

phase-427's node merge means there is no `class Node` in `nros.hpp` any more. This branch's seven
`create_*` verbs routing through `detail::require_created` are **not** superseded, so they were
ported onto main's out-of-line shape rather than dropped, and each was verified present by name:

```
$ grep -c require_created packages/api/nros-cpp/include/nros/nros.hpp
9                       # 1 definition + 8 call sites
$ git show origin/main:packages/api/nros-cpp/include/nros/nros.hpp | grep -c require_created
0                       # main has none — this is entirely this PR's payload
```

Eight call sites, i.e. the seven verbs plus the eighth `main` added — `create_publisher`,
`create_subscription`, `create_subscription_raw`, `create_service`, `create_client`,
`create_timer`, `create_wall_timer`. Not merely grepped: `just check cpp`'s
`failed_create_aborts.cpp` probe compiles and passes.

Also checked that this branch does **not** re-add an inline `NROS_CPP_HAS_*` capability
predicate — phase-438 W1 moved those to `nros/std_detect.hpp`, and a branch side re-declaring one
would be superseded. `git diff origin/main..HEAD -- packages/api/nros-cpp/include/` adds no such
line.

### Gates

| gate | result |
|---|---|
| `just check fast` | **286 / 287** |
| `check xrce-vendored-versions` | FAIL — **environmental**, reproduced identically on a plain `origin/main` worktree (`no vendored tree checked out`; needs `nros setup --source micro-cdr --source micro-xrce-dds-client`) |
| `cargo +nightly fmt --check --all` | PASS |
| `just check api-parity` | PASS |
| `just check api-parity-ledger` | PASS |
| `just check ledger-orphan-refs` | PASS |
| `just check doc-refs` | PASS |
| `just check issue-ids` | PASS |
| `just check prose-issue-refs` | PASS |
| `python3 scripts/check-doc-commit-citations.py` | PASS |
| `cargo check -p nros-node --all-targets --features std` | PASS |
| `just check cpp` | every C++ compile probe PASS; fails only on `clippy::missing_safety_doc` for `nros_cpp_declare_remap` at `packages/api/nros-cpp/src/lib.rs:3262` — **pre-existing on `main`**, measured not assumed: a plain `origin/main` worktree fails the same gate with the same lint at the same line, and this branch's diff for that file is empty |

jerry73204:
## Rebased onto `main` (b5097a460) — 40 commits before, 40 after

`origin/phase-427-w9-context-every-target` was at `d25842376`; it is now `c6bdcafd9`.
No-loss check over commit subjects vs `origin/main` is EMPTY — nothing dropped.

### One conflict, the ledger row-wise class

`docs/reference/api-parity-ledger/log.json`: main added `rust:resolve_logger` (phase-427
W5), this branch's W5 pass added `c:log_severity_t`. Distinct keys — not a semantic
conflict, only adjacent text. Resolved ROW-WISE from git stages `:1`/`:2`/`:3`,
field-wise within any shared row (no field-level conflicts arose): base 72, main 73,
branch 73, merged **74**, one row from each side kept.

Nothing else conflicted. This branch's own
`eacd036bc fix(ledger): drop 4 silent duplicate keys…` and
`c6943ca44 fix(ledger): four rows carried a verdict/disposition pairing that exists
nowhere else` replayed unchanged, and the tip's
`fix(rebase): the eighth create verb main added…` still applies against main's current
out-of-line `node.hpp` shape — `just check cpp`'s "one node type: hosted shape + both
create families" and "one node type: FREESTANDING construct + create_publisher" cases both
compile, so no ninth verb appeared on main since the last rebase.

**Duplicate-key sweep after the rebase:** all 17 shards re-parsed with a
duplicate-raising `object_pairs_hook` — zero duplicates, zero conflict markers. This is
the check that matters here, because the discard happens in files git reports as cleanly
auto-merged.

### Gates (measured in a clean worktree at the rebased tip)

| gate | result |
| --- | --- |
| `just check api-parity` | every divergence carries a ledger entry |
| `just check api-parity-ledger` | self-test 0 failed |
| `check-doc-commit-citations.py` | OK — 206 citations across 913 live files |
| `just check cpp` | All C++ checks passed |

`xrce-vendored-versions` is ENVIRONMENTAL: reproduced identically on a plain
`origin/main` worktree — a fresh worktree has no `nros setup --source micro-cdr --source
micro-xrce-dds-client` tree, so the gate verifies nothing and reports rc=1.

jerry73204:
## Rebased onto `main` after #629 landed — 40 commits → 9

`c6bdcafd9` → `19cbe9482`, base still `main`. #629 put 35 commits on `main`, most of
them this branch's own earlier work re-authored there, so the rebase is mostly
subtraction. Every drop was verified individually rather than assumed.

### No-loss verification

`comm -23` over commit subjects, `origin/main..<old tip>` against `origin/main..HEAD`,
left **31** subjects. **30 of them resolve on `main` by subject** (`git log origin/main
--grep=<subject> --fixed-strings`). The one that does not:

* `docs(phase-428): six commit citations named nothing — replace them with what survives`
  — subsumed by `cc428cf8d` on `main` ("restore the citation fix the rebase dropped"),
  which re-applies it under a different subject. Measured rather than trusted: all six
  dead short hashes (`7b3bab8e9 839f69b5c 9c492403d ca70b9f6c e82830bb6 fe807a999`) occur
  **0** times in `origin/main`'s copy of `docs/roadmap/phase-428-api-porting-principle-sweep.md`.

### Non-mechanical resolutions

**1. `docs(phase-428 W5): 82 ledger rows` — skipped as subsumed by `86792f9ea`.**
Not identical patches, so git would not drop it. Diffed the two patches: **12 of 13
files byte-identical**; `timer.json` differs only by a MOVE of three keys
(`rust:DeclaredNode::create_timer_for_callback_name_on_clock`,
`rust:NodeCtx::create_timer_on_clock`, `rust:NodeCtx::create_timer_on_clock_in`) from
one offset to another. All three are present **exactly once** on `origin/main`
(`grep -c` per key). A pure reordering of a dict is not content, so the commit is
subsumed.

**2. `fix(phase-428): eager-created C entities resolve their names` — skipped.**
`main` carries it as `8b231b484` **plus** `eb3339dae`, which is `main`'s own copy of
this branch's tip commit and adds the eighth verb (the free `rclcpp::create_timer`).
`origin/main`'s `nros.hpp` has **9** `require_created` occurrences: the definition plus
all eight call sites. The branch's version also carried three prose corrections to
`detail::abort_failed_create`'s doc block that `main`'s does not; those are **not**
subsumed and are being carried on #773 instead (the file is untouched here), so nothing
is lost — see the note there.

**3. `docs(ledger): the three rows main added since the last rebase` — skipped as
subsumed by `7a0c903df`.** `main`'s version MERGED both authors' prose rather than
picking one. Verified field-wise in Python, not by eye: for
`cpp:LifecycleTransition`, `cpp:shutdown_transition_for` and `rust:throttle_decide`,
the branch's `why` is a literal substring of `main`'s, the `verdict` matches, and
neither side carries any other field.

**4. `fix(ledger): drop 4 silent duplicate keys` — kept, with two of its four hunks
dropped.** This is the shape CLAUDE.md warns about: git's line-based merge creates
duplicate JSON keys with no conflict marker. Measured with an `object_pairs_hook`
counter over all 17 shards at each step:

| tree | top-level duplicate keys |
| --- | --- |
| `origin/main` | 0 |
| rebase HEAD before this commit | 2 — `c:node_init`, `c:subscription_init` |
| after resolution | 0 |

`c:publisher_init` and `rust:throttle_decide` were **already** de-duplicated on `main`
(`198afab46`, `7a0c903df`), each by MERGING the two authors' prose — so this commit's
removal hunks for them would have deleted `main`'s merged row. Both hunks resolved to
`main`'s side. The other two rows keep this branch's phase-428 Q1 row, which is the
resolution this branch already made and which #629 does not touch: both `c:node_init`
and `c:subscription_init` exist unchanged at the **merge base** and on `main`, so the
choice predates #629 and is not a new decision.

**5. `fix(rebase): the eighth create verb main added` — dropped by git as
`patch contents already upstream`** (`eb3339dae`). Its two-line formatting delta over
the branch's copy is `main`'s, and `main`'s is the rustfmt/clang-format-correct one.

### Gates

Run in a worktree detached from `origin/<branch>` (outside the repo tree — a worktree
*inside* `/home/aeon/repos/nano-ros` makes cargo walk up to the root workspace and
breaks `cargo fmt`, which is what `leaf-lockfiles` was tripping on in the first run):

* `just check fast` — **289/290**. The one failure is `xrce-vendored-versions`, which
  fails **identically on a plain `origin/main` control worktree** (vendored trees not
  checked out here). Pre-existing and environmental.
* `just check api-parity`, `api-parity-ledger`, `ledger-orphan-refs`, `doc-refs`,
  `issue-ids`, `prose-issue-refs` — all OK.
* `python3 scripts/check-doc-commit-citations.py` — OK, 206 citations across 914 live
  files resolve. The two "may no longer be needed" baseline entries (`d97c9c606`,
  `e3786749a`) are reported identically on the `origin/main` control.
* `just check cpp` — all C++ checks passed, including `failed_create_aborts: OK`.
* `cargo +nightly fmt --check --all` — clean.
* `cargo check -p nros-node --all-targets --features std` — clean.
* All 17 ledger shards re-parsed with an `object_pairs_hook` counter: **zero** duplicate
  top-level keys.

PR #707, opened 2026-09-07
fix(api-parity): 43 of the 48 "missing" C symbols are ones we export

jerry73204:
`check-api-parity` was red with 51 unledgered rows, which withdrew `test-unit` and `test-lane-contracts` from `just ci gate` — the lane stops at the first failure, so nothing below it had run for a while.

**The obvious reading was "classify 51 rows". That would have been wrong, and permanently.** A ledger row is AUTHORED and nobody re-derives it, so writing *"gap: rcl has it, we do not"* for a function we ship would have been believed by every later reader. I had 22 such rows drafted — `declined` for the options getters, `gap` for the `is_valid` and `fini` families — before noticing this in our own generated header:

```c
NROS_PUBLIC bool rcl_clock_time_started(const struct nros_clock_t *);
```

## The actual defect

phase-417 stage 6 step A ("the ROS 2 spellings are ours") adopted the upstream names directly wherever our contract matches: `nros/nros_generated.h` declares **39 `rcl_*` and 8 `rclc_*` entry points as `NROS_PUBLIC`**. The tool could not see any of them, in **two independent places** — and fixing either alone changes nothing:

- `ours_c()` harvested with `prefixes={"nros_", "NROS_"}`, so they were never collected;
- `correlate.LIB_PREFIXES["c"]` stripped only `nros_`/`NROS_` on our side, so anything collected keyed as `rcl_client_is_valid` while upstream keyed as `client_is_valid`.

Both now carry `rcl_`/`rclc_` on our side too. This preprocesses *our own* header, so a `rcl_*` name found there is ours by construction — upstream's headers are never on that include path.

| | before | after |
|---|---|---|
| `same` | 39 | **101** |
| `theirs-only` | 416 | 354 |
| unledgered | 48 | **5** |

Sixty-two symbols move from "we lack this" to "we have this", which is what was true all along.

## The 8 rows that are real

- **5 × `rename`** — `c:{client,service,publisher,subscription,node}_init`. We export both `rcl_*_get_default_options` and the options struct (they correlate `same` now); what is missing is the plain `init` that consumes them. Ours is `nros_*_init_with_options`, and for node `nros_node_init_ex`, which takes `nros_node_options_t*` exactly as `rcl_node_init` takes `rcl_node_options_t*`. Same shape, different name, no platform reason — so `rename` per the schema, and they join the campaign's work list rather than being excused. The node row cites its ours-only half, `c:node_init_ex`.
- **3 × `extension`** — `cpp:LifecycleTransition` (an id enum; rclcpp's `Transition` *object* is separately ledgered as a gap, so this is not that under another name), `cpp:shutdown_transition_for` (rclcpp resolves it inside a state machine a C-ABI caller cannot reach), `rust:throttle_decide` (the throttle macros are `no_std` and expand in the caller's crate, so the predicate has to be a callable item there).

The ledger edits are inserted textually rather than re-serialised: the shards preserve insertion order, not a sort, so a `json.dump` round-trip rewrote 5,800 lines for 8 rows. The diff is **+32 lines, nothing else touched**.

## Verified

`just check api-parity` green; `--self-test` 0 failures; `just check fast` 258/258; and **`just ci gate` passes all six steps** — `test-unit` and `test-lane-contracts` run again.

## Worth knowing

`api-parity` is not on the fast line, so nothing on the push path or the required PR context runs it. It appears only at `just ci gate`, which is why a report that had been wrong about 43 symbols went unnoticed — the same shape as the `check-build` problem CLAUDE.md already records. Not changed here.

### DIFF
--- a/docs/reference/api-parity-ledger/lifecycle.json
+++ b/docs/reference/api-parity-ledger/lifecycle.json
@@ -554,5 +554,13 @@
   "c:lifecycle_callback_t": {
     "verdict": "extension",
     "why": "`uint8_t (*nros_lifecycle_callback_t)(void *context)` -- the named typedef for a lifecycle transition callback, added by issue 0792. rclc has no counterpart because it spells its callback inline as `int (*)(void)` at each of its four `rclc_lifecycle_register_on_*` declarations and never names the type. We name it for two reasons: it is the shape ALL TWELVE of our entry points share (the six standalone and the six executor-scoped), and naming it is what makes it expressible in C at all -- the unnamed cross-crate `Option<LifecycleCallbackFnCtx>` it replaced reached the header as an opaque struct passed by value. Same convention as `nros_timer_callback_t` and `nros_guard_condition_callback_t`. The `void *context` parameter itself is the `callback-bound-at-creation` divergence recorded on `c:lifecycle_register_on_configure`."
+  },
+  "cpp:LifecycleTransition": {
+    "verdict": "extension",
+    "why": "An ENUM naming the lifecycle transition ids, which the C ABI carries as a `uint8_t` across `nros_executor_lifecycle_change_state`. rclcpp models a transition as a `Transition` OBJECT (constructed, carrying label and id) — a different shape, and its absence here is recorded separately as `cpp:Transition` (gap). This is not that object under another name: it is the id vocabulary an embedded caller passes by value, with no allocation and no lifetime."
+  },
+  "cpp:shutdown_transition_for": {
+    "verdict": "extension",
+    "why": "Maps a lifecycle state onto the shutdown transition legal FROM that state. rclcpp has no such helper: its `LifecycleNode::shutdown()` resolves the same thing inside the state machine, which is reachable because the node owns it. Ours is a free function because the state machine lives behind the C ABI and a caller holding only the current state cannot ask it."
   }
 }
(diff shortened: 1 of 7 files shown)


jerry73204:
Closing: **superseded on main, and by better rows than mine.**

While this was in the queue, the issue-1188 work landed the same two script fixes and all eight ledger rows. Checked line by line:

- `scripts/api-parity.py` `ours_c()` prefixes and `correlate.LIB_PREFIXES["c"]` — main's versions are a **superset** of mine (they carry the uppercase `RCLC_`/`RCL_` variants too) and cite issue 1188. My half added nothing.
- All 8 rows (`c:{client,service,publisher,subscription,node}_init`, `cpp:LifecycleTransition`, `cpp:shutdown_transition_for`, `rust:throttle_decide`) already exist.

And main's rows are **more accurate**. I classified `c:node_init` as `rename` — same capability, ours is the name that should change. Main has it as `declined` with `disposition: adopt`, citing `c:node_init_with_options` as the pair and recording that `nros_node_init` *existed and was deliberately removed*. That is a fact I did not have, and it makes `rename` the wrong verdict: the absence was an act, not a spelling choice.

`just check api-parity` is green on `origin/main` with none of this branch applied, so nothing here is needed to unblock `just ci gate`.

The one thing worth keeping from this branch is an observation, not code: `api-parity` runs on neither the fast line nor the required PR context — only at `just ci gate` — which is why a report that had been wrong about 43 symbols went unnoticed. Same shape as the `check-build` problem CLAUDE.md records. Left alone here.

PR #724, opened 2026-09-07
phase-426 W4: delete both C++ parameter stores

jerry73204:
phase-426 **W4 only**, on top of #719 (W1+W2). Do not merge before it.

## What this does

Three stores existed for one concept. The executor's `nros_params` table is the one the six `rcl_interfaces/srv/*` servers read — so it is the one `ros2 param get` sees — and `rclcpp::Node` and `nros::ComponentNode` each carried an inline `nros::ParameterServer` of their own. A parameter declared through either C++ facade was invisible to `ros2 param get`, and a sibling node did not share it. Both members are deleted; both facades forward to the executor's store.

They forward through **one** header, `nros/node_parameters.hpp`, not two spellings — the two facades already flagged their duplication against themselves, and a shared helper is what the tree's own rule asks for.

## The FFI it needed

`nros_cpp_*_param_*` took a bare executor handle (so every call landed on the primary node) and had no setter and no existence check. #719's report named the missing writer; this adds the rest: 19 node-scoped entry points keyed by the `nros_cpp_node_t*` a C++ node already holds, whose `node_id` **is** W1's key. Declare / get / set for bool, integer, double, string; `has`; declare / get for the three array element types (`ComponentNode::declare_parameter<std::vector<double>>` is a real capability — deleting its store without an array channel would have turned a weight matrix into a silent default).

They are defined **unconditionally**, which is a link-time decision: the facades live on node types every C++ image compiles, whether or not its bringup declares `param_services`. Gating the symbols would make a ported rclcpp node fail to link on an image that never asked for parameter services. Without a store they answer `NROS_CPP_RET_UNSUPPORTED` — a code the facades report (`ComponentNode` makes it boot-fatal, naming the capability), never a default that pretends to have been stored.

Every set routes through `ParameterServer::apply`, so the C++ facade and the wire cannot disagree about whether a write may happen.

## `NROS_RCLCPP_MAX_PARAMS`

**Gone**, not redocumented — with the member deleted there is no per-node arena for a number to bound. The same is true of `NROS_COMPONENT_MAX_PARAMS` / `MAX_SEQ_PARAMS` / `SEQ_POOL_BYTES` / `NROS_PARAM_SEQ_DEFAULT_CAP`; a `ComponentNode` was carrying 4 KiB of sequence pool and 256 scalar slots whether or not it declared a parameter. The image's one arena is `NROS_MAX_PARAMETERS` on `nros-params`, shared across every node rather than multiplied by the node count. Both header sites say so where the knob used to be.

## Capability layout

Measured before and after, and the gate now measures the macro that matters here:

| | before | after |
| --- | --- | --- |
| `sizeof(rclcpp::Node)` | **3752** | **240** |
| ...under `-DNROS_SYSTEM_PARAM_SERVICES=1` | 3752 | **240** |
| `sizeof(::nros::Node)` | 192 | 192 |
| `sizeof(::nros::QoS)` | 36 | 36 |

`check-cpp-capability-layout`: `OK — 3 type(s) x 7 capability macro(s)` before, `OK — 3 type(s) x 8` after. The eighth is `NROS_SYSTEM_PARAM_SERVICES`, added here: it is the one capability macro no header defines for itself (`NanoRosCapabilities.cmake` sets it per directory), which makes it exactly the shape the gate exists for, and it had never been measured against a layout while gating the branch that decides whether a node has a parameter store.

## Two things nothing was compiling

* **`NROS_CPP_STD` parameter overloads.** The `std::string`-valued and `std::vector<T>` facades on both node types were built by **no lane**: the header sweep is `-std=c++14 -ffreestanding` without the macro, the `-nostdinc++` probes are the opposite configuration, and the only in-tree definer is one px4 bridge module no gate builds. New probe `param_hosted_overloads.cpp`, wired into `check cpp`. Negative control run, not assumed: renaming `node_param_get_array_ffi` fails the probe at `declare_parameter [with V = std::vector<double>]` while the c++14 sweep still passes.
* **`params_shim.rs`'s test module never compiled.** It asserts with `assert_eq!` on `ParameterValue`, which has no `PartialEq`; no lane builds it (`test-unit` is `--workspace` with no features, so the cfg is off). Five tests read as coverage while asserting nothing. Fixed in place, and eight new tests added beside them.

## Acceptance, and which half is provable here

*"A parameter declared through `rclcpp::Node::declare_parameter` is visible to `ros2 param get`."* The **store** half is what this PR can prove: the facade declares into the table the six services read, keyed by the declaring node. The **wire** half — the services answering per node FQN — is W3, and an end-to-end `ros2 param get` cell is W6. Stated rather than claimed.

One behaviour narrows, deliberately and in the code: launch seeds are written before any node exists, so they land on the primary. A single-node C++ image is unaffected (its node **is** the primary, so issue 0745's adoption now happens for free rather than through a helper). In a multi-node image only the primary adopts, where before every node adopted the primary's value for a shared name — the collision W1 exists to remove. Keying the seed per node needs the launch resolver to say which node an override belongs to; the phase doc puts that in RFC-0060 territory.

## What ran

```
just check cpp                     All C++ checks passed
just check c                       All C checks passed
just check cpp-capability-layout   OK — 3 type(s) x 8 capability macro(s)
just check ffi-struct-mirrors      in sync (component.h + nros_native_tier_spec_t x8)
just check cbindgen-headers        OK (3 committed headers match a fresh generation)
just check api-parity              every divergence carries a ledger entry
just check api-parity-ledger       0 checks failed
just check fast                    264/265 (see below)
cargo test -p nros-node --lib --features std,param-services      373 passed, 0 failed
cargo test -p nros-cpp  --lib --no-default-features \
    --features std,rmw-cffi,platform-posix,ros-humble,param-services   26 passed, 0 failed
```

`check fast`'s one red is `doc-commit-citations`, and it is **not from this branch**: four baseline entries resolve because this worktree fetched extra refs. Verified by `git stash`-ing the entire diff and re-running the gate on the pristine base — identical failure, same four hashes.

## Notes for the reviewer

* W3 is in flight on the same base. This branch does not touch `register_parameter_services`, `executor_sizing.rs` or `packages/api/nros/src/node.rs`.
* No repo lane compiles `nros-cpp`'s lib tests with `param-services`, which is how the broken test module survived. The feature combination above is the one that builds them; worth a lane, and worth its own issue rather than a drive-by here.
* `nros::ParameterServer` itself is untouched — `examples/native/cpp/parameters` still uses it directly. What changed is that no node type owns one. Whether the class should survive is a W5 question and the ledger row says so rather than answering it.


jerry73204:
Rebased onto current main — 13 commits to 7. The six drops are #719's set, which has landed; each verified present on main by subject (`d1f3c7432`, `3d156b4cf`, `30f5e9f94`, `f5efbbc57`, `e6a2f7f71`, `e727af142`).

The one real conflict was three `why` fields in `docs/reference/api-parity-ledger/param.json` (`cpp:Node::{declare,get,has}_parameter`), and it looked like competing narratives until the structure was measured:

    base   1385 chars
    branch 2508   startswith(base) = True   -> pure APPEND
    main   2089   startswith(base) = False  -> edited the base AND appended

So the two sides are not in competition. Main rewrote the body (the umbrella header now unconditionally includes `component_node.hpp`, and the `gap` verdict still validates because this row keys on `nros::Node`) and this branch appends a different fact on top of the old body — that phase-426 W4 DELETED the node-local C++ stores, so both facades now forward through `nros/node_parameters.hpp` to the one `nros_params::ParameterServer`.

Resolved as main's newer body plus this branch's appended block, chronologically. Both annotations survive — 6 `RETIRED 2026-09-07` and 8 `PHASE-426 W4` mentions in the file. Merged row-wise then field-wise over the git stages, so nothing outside those three fields was touched; 147 rows, `param.json` parses.

Verified: `api-parity-ledger`, `c-array-guard-probe`, `c-array-pool-floors` pass, `cargo +nightly fmt --check --all` clean.

jerry73204:
Rebased onto `main` after phase-427 landed, and the rebase was a re-port rather than a replay. Recording it because the description above now measures a tree that no longer exists.

## What moved underneath this branch

phase-426's plan said W4 gates phase-427 W1 — "the parameter store is not hosted-only, so the node merge's `void* hosted_` does not move it, and one type means one store size" — and the queue landed them the other way round. #755 merged first, so `rclcpp::Node` is now the merged node type in `nros/node.hpp`, and the store came with it: `NodeHosted` carried `ParameterServer<NROS_RCLCPP_MAX_PARAMS> params`, described in place as "NODE-LOCAL: `ros2 param list` reads the EXECUTOR's store, which is Rust-side (RFC-0089 decision 3, issue 0793)". The defect this branch deletes was moved, correctly labelled, and kept.

So the merge was not textual. `git rebase` conflicted in five regions of `nros.hpp`, and in three of them the code this branch edits had left the file entirely.

## How it was resolved

- **`nros.hpp`** keeps the out-of-line definitions phase-427 left there. `Node::declare_parameter` / `get_parameter` (both forms) / `set_parameter` / `has_parameter` now call `::nros::detail::node_param_*` against `this->ffi_handle()` — the accessor phase-427 already exposes — instead of `this->hosted().params`. `adopt_executor_param_seed` and its four overloads are deleted as before: with one store there is nothing to reconcile.
- **`node.hpp`** loses the three things that were the store: the `ParameterServer<NROS_RCLCPP_MAX_PARAMS> params` member on `NodeHosted`, the `parameters()` accessor pair, and the `NROS_RCLCPP_MAX_PARAMS` knob, each replaced by the note saying what went and where the one arena now is.
- Two conflicting regions were taken from `main` unchanged — the `rclcpp::create_timer` block and the services banner — because this branch never touched them; they only collided because the class moved out from between them.
- `component_node.hpp`, `node_parameters.hpp`, the FFI, the shim and the compile probes applied unchanged.

## What the description above no longer describes

The `sizeof` table (`rclcpp::Node` 3752 → 240) was measured when the store was an inline member of the class. phase-427 had already moved it into the heap-allocated hosted block, so that table's before-value is not this tree's, and I did **not** re-measure it — the acceptance clause is that `check-cpp-capability-layout` still passes, and it does, now over five types and eight capability macros, with its own three-case self-test including the mutation control. The saving after this rebase is in `NodeHosted`, not in `sizeof(rclcpp::Node)`.

Also stale: "W3 is in flight on the same base. This branch does not touch `register_parameter_services`, `executor_sizing.rs` or `packages/api/nros/src/node.rs`." W3, W5 and W6 have all merged; the statement about which files this branch touches is still true.

## Verified after the re-port

```
just check cpp                     All C++ checks passed
just check c                       All C checks passed  (executor_param_node_keying: two nodes, two parameter tables, one store)
check-cpp-capability-layout        OK — 5 type(s) x 8 forced capability macro(s), self-test 3/3
just check build                   1 red, rmw-xrce, pre-existing and unrelated (see below)
just check api-parity              pass
just test-unit                     1232 passed; the one non-pass is a zenoh router precondition the junit rewrite classes as a skip
just test-lane-contracts           17 passed
```

Before the rebase I also ran `just ci gate` over the full phase-426 stack (W3 + W6 + W4 on one tree) and the `params_per_node_interop` cell against it live — 3 passed, 0 skipped — so W4 and the per-node services were checked together rather than only apart.

Two reds are not from this branch and are stated rather than hidden:

- **`check-rmw-xrce`** fails wherever the micro-XRCE submodules are checked out. `nros-rmw-xrce/CMakeLists.txt` has required `NROS_XRCE_CFFI_OUT_DIR` since phase-420 W9 (`56cb10f45`) and no recipe sets it, so `just check rmw-xrce` configures and stops. It is invisible to CI because `pull_request` runs `check-fast` and not `check-build`, and invisible to anyone whose submodule is uninitialised because the recipe then takes its loud-skip arm. No phase-426 file is involved.
- **`check-doc-commit-citations`** was red on a full local clone for a dead `AGENTS.md` hash, which is fixed in its own commit here with the reasoning in the message. It came in on the rebase; the `pre-push` hook runs that gate, so it refuses every push from a full clone until it resolves, while CI's shallow clone skips it.

jerry73204:
Rebased onto `main` (`7ff68c777`). 8 commits before, 7 after — one deliberate drop, below. No-loss check (`comm -23` of subjects against pre-push sha `ebd6bbf2d`) shows exactly that one subject.

## Conflicts and resolutions

**1. `scripts/check-cpp-capability-layout.sh` — modify/delete.** `main`'s `fix(#1225)` (`d93cc31e6`, "the layout gate measured three names; derive all ninety instead") deleted the shell gate and replaced it with `scripts/check-cpp-capability-layout.py`. This branch's W4 commit added `NROS_SYSTEM_PARAM_SERVICES` to the `.sh`'s `CAPS` list. Ported the addition — the macro plus its rationale comment — into the Python gate's `CAPS` tuple and deleted the `.sh` leftover. The rationale still holds verbatim under the new gate: it is the one macro on that list no header defines for itself (`NanoRosCapabilities.cmake` sets it per DIRECTORY), which is exactly the shape the list exists for.

Measured after the port: `check-cpp-capability-layout` self-test 6 cases OK, then **90 derived subjects x 8 forced capability macros** hosted plus the freestanding `-nostdinc++` arm against the ThreadX shim — 8, not 7, so the new macro is in the sweep, and no baseline entry moved.

**2. `AGENTS.md` — content conflict, superseded.** `main` landed its own fix for the same defect first: `7d7d04025 docs(AGENTS.md): the stacked-branch paragraph stranded its own citation, and it is red on main`. `main`'s version is the more complete answer — it names the stranded commit by SUBJECT rather than describing it, and adds the "a citation to a commit on an UNMERGED branch is a citation with an expiry date" lesson. Resolved to `main`'s text, which left this branch's commit empty.

**Deliberately dropped:** `docs(AGENTS.md): the stranded-stack paragraph cited a commit that had been rebased away`. Verified present on `main` under `7d7d04025`; `AGENTS.md` is no longer in this PR's diff at all.

**No conflict in `docs/reference/api-parity-ledger/param.json`** — it merged clean, but I checked it the way the ledger requires rather than trusting that:

- All **17 shards** parsed with an `object_pairs_hook` counting keys per object: **zero duplicate keys**, at the top level or nested.
- `param.json` row-wise against `origin/main`: 153 keys before, 153 after, **0 added, 0 removed, 9 changed** — `cpp:Node::{declare,get,set,has}_parameter`, `cpp:Node::parameters`, `cpp:ComponentNode::{declare,get,has}_parameter`, `cpp:ParameterServer`. Exactly the surface this wave reworks, and no key gained a second offset.

## Gates

- `just check cpp` — **All C++ checks passed**, including this PR's new `hosted parameter overloads: std::string + std::vector<T> on both node types (c++17)` probe and every phase-427 W8 `Result`/`Expected<T>` probe `main` added.
- `just check c` — **All C checks passed**, including `executor_param_node_keying: two nodes, two parameter tables, one store`.
- `just check cpp-capability-layout` — OK, self-test 6 cases, 90 x 8.
- `just check api-parity-ledger` — self-test 0 failures; `gen-api-comparison` self-test 0 failures. (`check ledger-orphan-refs` is not a recipe in this tree; the ledger-integrity gate that exists is `check-ledger-doc-single-home.py`, run: **OK, 17 shards point at one SCHEMA.md**.)
- `just check c-array-guard-probe` — OK, 21 guards across 9 files fire at 0 and are silent at defaults (this PR teaches it `node_parameters.hpp`).
- `just check c-array-pool-floors` — OK, 24 knob-sized C arrays: 21 guarded, 3 zero-legal, 0 unclassified.
- `just check doc-refs` / `issue-ids` (1211) / `prose-issue-refs` / `doc-recipe-refs` / `check-doc-commit-citations.py` (196 citations across 907 files) — all OK.
- `cargo check -p nros-params --all-targets` — clean. `cargo +nightly fmt --check --all` — clean.
- `just check fast` — **1 of 284 gates failed: `xrce-vendored-versions`**, which is environmental in a fresh worktree (the micro-XRCE vendored trees are not checked out). Reproduced identically on a plain `origin/main` worktree, same message, same rc — pre-existing, not this PR.

PR #734, opened 2026-09-07
fix(#1161): a capability skip fails unless the lane declared it

jerry73204:
Issue 0584 made an absent in-lane **fixture** a hard failure. `check-skip-budget` already modelled
the three skip classes and printed the actionable ones — but never *failed* on any, so a lane
could run 7 of its 20 tests and report pass. Same sentence, one class over.

## The rule

A `capability` skip whose reason matches no line in `.config/capability-skip-baseline.txt` now
fails and names what was missing.

A **ratchet, not an allowlist**: each line states what is absent, why a host may legitimately lack
it, and what would retire it. The set may shrink freely; it cannot grow without someone writing
that paragraph. A **missing** baseline declares *nothing* rather than allowing everything — the
direction that keeps a ratchet ratcheting when someone deletes it, and it is self-tested.

Two entries, neither of which is what this issue was filed about:

- **`rmw_zenoh_cpp/rmw_zenohd`** — ships as a ROS package outside a base install, resolved through
  `AMENT_PREFIX_PATH` rather than PATH (RFC-0075), so "installed but not sourced" correctly reads
  as absent (issue 0774).
- **`ZPICO_MAX_SESSIONS=1`** — the default single-session shim, which is right for a shipped
  target; the multi-session paths have their own lane.

## What it caught first was not what I filed

The lane this issue is named after now runs **20 of 20** — the router got installed on this host
between filing and fix — so the rule never fires there. It fired on `test-unit`, twice:

1. **A misclassification.** *"default locator port 7447 … already in use"* was reported as
   `capability`, because a bare `skip!` **defaults** to that class and the site never said
   otherwise. A bound port is a `resource` — a runtime prerequisite a rerun or a quieter host
   resolves — and the two have opposite remedies. Mislabelled, it would have demanded a baseline
   entry for a condition nobody can provision away. Now `skip_class!(resource, …)`, and the
   breakdown reads `capability=1 resource=1` where it read `capability=2`.
2. **The `ZPICO_MAX_SESSIONS` cell**, which CLAUDE.md already records as *"skipped on every host in
   every tier"* — sitting in a budget line nobody was required to read.

The first is the argument for the rule: the class was wrong **by default**, in a green lane, and
nothing would have said so.

## Verified

Both directions against crafted junit, rather than by waiting for a host to lack something:

```
declared capability   -> rc 0
undeclared capability -> rc 1, naming the reason and the first test
```

plus four self-test cases on the declaration itself (declared allowed, undeclared refused, empty
declaration refuses everything, missing file declares nothing).

```
just test-unit                     1214 ran, capability=1 resource=1, rc 0
just check required-features-tests PASS (20 of 20 here)
just check fast                    275 ran, 1 ledger skip, 0 failed
```

## Also: #1131 sharpened, not closed

Its one remaining knob needs a Zephyr build and this host has none — exactly as the issue says.
The search for one turned up a fact that needs no Zephyr and changes what the question is worth:
`NROS_ZEPHYR_MAX_TIERS` and `NROS_ZEPHYR_TIER_STACK_SIZE` have **no producer at all** — no Kconfig,
no cmake define. Every Zephyr image links `4 × 16384 = 65,536 B` of tier stacks and **no image can
set otherwise**, whatever the zero ruling turns out to be.

So the reachable question is "can an image state this?", not "is zero legal?" — and the C side is
already `#ifndef`-shaped for a producer. Recorded on the issue rather than attempted, because
verifying a producer needs the same Zephyr build the ruling does.


jerry73204:
Ejected from the merge queue by the gate this PR adds — **which is the gate working**, so the fix is here rather than a retry.

```
ERROR: 3 test(s) skipped for a capability this lane never declared it may lack:
   1x zenohd not found — cannot open a session to observe acceptance
   1x zenohd not found — run `just build-zenohd`
   1x zenohd unavailable (looked at the nros setup store and PATH) — ...
```

Before #1161 that run reported **pass** having done less than it claimed. That is the whole point of the change, and the first thing it caught was itself.

## What is actually missing

One thing: the ROS zenoh router, absent on a GitHub-hosted runner because there is no ROS install there at all. Counting the tree, that one absent router has **five** spellings — two from resolvers (`fixtures::zenohd_router`, `process::zenohd_unavailable_reason`) and three invented at the call site. The baseline could match none of the three.

## The call sites are the defect, and the fix already existed

`nros_tests::zenohd_unavailable_reason` was added for issue 0982 for precisely this — *"a caller that turns unavailable into a SKIP must be able to put the cause in the skip message"* — and these three restated it instead of using it. They now report what the resolver says, so the wording has one owner and cannot drift from the thing that decides it.

That also retires a dead instruction: two of the three told the reader to run **`just build-zenohd`**, which does not exist. RFC-0075 / phase-362 W4 deleted the vendored router and `just --list` has no such recipe. The resolver's text names **sourcing** first (issue 0653), which is the actual common cause — `rmw_zenohd` is invisible to `command -v` even when installed.

## One baseline line, not a widened one

Added `no ROS zenoh router`. Deliberately **not** widening the existing `rmw_zenoh_cpp/rmw_zenohd` entry to cover both: those are two resolvers with two texts, and this file's own rule is that a pattern must be narrow enough to name ONE missing thing — a broad pattern re-creates the hole the file closes. Both entries carry what would retire them, and they collapse into one the moment the two resolvers share a string.

## Verified

`check-skip-budget --self-test` OK (5 extraction cases + grouping); the recorded reason is matched by the new pattern and by no other; `cargo check -p nros-rmw-zenoh --all-targets` clean; `cargo +nightly fmt --check --all` clean.

PR #737, opened 2026-09-07
docs(RFC-0093, phase-437): a board name states the reach of its build

jerry73204:
Phase-422 W7 made both board vocabularies **resolve** and deferred which is canonical — on the grounds that the index namespace is what users type while the cmake namespace is what the build uses, and neither is obviously the one to keep.

That framing has no answer, because it asks for a house style. RFC-0093 asks a different question — **what is this build actually true for?** — and the names follow.

## The names are untrue in both directions

| name | claims | measured |
|---|---|---|
| `riscv64-qemu` | an ISA | one QEMU machine: UART `0x10000000`, CLINT `0x02000000`, virtio-MMIO `0x10001000`, and a write of `0x5555` to the `test-finisher` at `0x100000` — a device **no silicon has** |
| `qemu-arm-freertos` | any ARM under QEMU | links `mps2_an385.ld` |
| `nuttx-qemu-arm` | NuttX on ARM under QEMU | `CONFIG_ARCH_BOARD="qemu-armv7a"`, `CONFIG_ARCH_CHIP="qemu"`, PL011 @ `0x9000000` IRQ 33 |
| `s32z270-freertos` | — | **real NXP silicon**, and not an index key at all: an emulator-first namespace never had a name to give it |

## Five rules

1. Name the narrowest thing the build is true for — **system** / **part** / **machine model**
2. `<where>-<stack>`, stack always last
3. Never name the emulator
4. Borrow the vendor's word — **R4 outranks R5**
5. Leave out what the index already carries as a field (`arch`, `platform`)

R4 over R5 is why `qemu-armv7a-nuttx` keeps a word that looks like an arch: it is NuttX's own `CONFIG_ARCH_BOARD`, and a borrowed name is checkable where an invented one becomes a third vocabulary.

The rule is **not** hardware-first or emulator-first. Both pick a word order and then argue; each is right for some boards and wrong for others. "Keep the index consistent" was never available either — `fvp-aemv8r-smp` is machine-first and already sits among the `qemu-*` keys.

`esp32-c3` is R3's witness, from the tree rather than from principle: its overlay reads *"board overlay for ESP32-C3 (real silicon) + ESP32-C3 QEMU"*. One name, both, because the name is the part.

**Sixteen index keys become twelve**, plus three boards the index cannot currently reach at all.

## Found while measuring — phase-437 W2, not a footnote

Comparing every `[board.*]` **body**, four pairs are byte-identical and `check-board-vocabulary` asserts three. It keys on a `# = [board.X]` marker, and **`qemu-arm-baremetal` / `mps2-an385` never got one** — two names, one build, nothing holding them in step. (`native`/`posix` are also identical and correctly unmarked; `check-host-platform-vocabulary` owns that pair.)

## The phase

phase-437 stages the rename across five namespaces — index, cmake overlay, board crate, fixture coordinate, and 50 `package.xml` exports — with the **gate first, in ratchet form**, because the gates that exist assert a board *resolves* and not that its name is true. Directories move **last**, one `git mv` per commit, so a bisect can land between them.

Blast radius is measured per name in the phase doc: 572 files for `mps2-an385` down to 31 for `esp32-c3`, and six `examples/<board>/` directories.

Its non-goals are as load-bearing as its waves: config axes (`-smp`) are not board names and this phase must not re-cut that axis while renaming; `deploy=` belongs to RFC-0087 D3; `native`/`posix`/`linux` are CLAUDE.md's; and **no package set, linker script or defconfig moves** — the name follows the build, never the other way round.

Nothing keys on a board name as a string — the harness resolves emulators by tool (`qemu_system_riscv32_path()`), never by parsing a board — so every rename is mechanical. That is the hazard, not the reassurance, and it is why this is a staged phase rather than a `sed`.

`just check fast`: 276/276.


jerry73204:
Superseded by #739, which cherry-picked RFC-0093 and the phase doc onto `main` and carries all seven waves. Closing so the two do not both land the same commits.

PR #739, opened 2026-09-07
feat(phase-437 W1+W2): the gate that reads a board's reach, and the pair the table always meant

jerry73204:
Carries RFC-0093 and the phase doc (cherry-picked from #737, which can close if this lands first), plus the first two waves.

## W1 — `check-board-name-reach`

On the fast line, ratcheted at **nine** violations — exactly the nine names RFC-0093 §4 retires, and nothing else. `mps2-an385*`, `mps3-an536-freertos`, `s32z270-freertos`, `esp32-c3`, `freertos-posix`, `threadx-linux`, `native`, `posix`, `zephyr` and `fvp-aemv8r-smp` all pass. That agreement is the wave's acceptance — and it was not assumed: **the first two implementations disagreed with the RFC and both were wrong.**

### Reach is read from the overlay alone, and the richer version is the bug

The obvious implementation walks the `packages/boards/**` directories each overlay names and looks for linker scripts and MMIO literals. It reported `freertos-posix` and `threadx-linux` as pinned to one target — both of which travel, and one of which says so in its own overlay (*"no cross toolchain — nothing here sets `CMAKE_C_COMPILER` or a linker script"*).

Those directories are **shared**: `nros-board-freertos` is referenced by four boards, `nros-board-common` by two. So the FreeRTOS Cortex-M linker script got attributed to the POSIX board, and a peripheral address in `threadx_hooks.c` to the Linux one. Attributing a shared crate's pins to every consumer inverts the answer for exactly the boards the rule exists to sort.

Reading the overlay alone makes the gate **conservative in one direction only** — a board whose linking comes from an external SDK (ESP-IDF) pins nothing and reads as portable. That is the safe direction: the SYSTEM rule then forbids naming a machine or emulator, which catches `qemu-esp32-baremetal` either way.

The vendor exemption (R4, which outranks R5) is **harvested, not listed**: `CONFIG_ARCH_BOARD` is read from the defconfigs the overlays name, so `qemu-armv7a-nuttx` passes because NuttX says that is the board's name — and the same string without the declaration fails. A listed exemption would be the third vocabulary the RFC exists to prevent.

Negative-controlled both ways: un-baselining `riscv64-qemu` reports it by name; adding a passing board to the baseline says the file may only shrink.

## W2 — one board, and the table said so all along

`[board.qemu-arm-baremetal]` and `[board.mps2-an385]` have byte-identical bodies and no `# =` marker. They are **one board**, and the evidence is the board table's own header:

> Keyed by the canonical board id (`examples/<board>/` dir name **or** the board-crate name).

That is the doubling rule, written in the contract. `qemu-arm-baremetal` is the examples directory, `mps2-an385` is the crate, and `719ff9436` added both — adjacent, in one commit. Born together rather than mirrored later, which is why the pair predates the marker convention.

The same commit is the control: it also added `[board.stm32f4]` with `cortex-m4` and `openocd` instead of `qemu`, because that one is real hardware you flash. The author encoded the difference when there was one.

Marked rather than collapsed, so `check-board-vocabulary` asserts it — **five** mirrored pairs now — for the waves before W5 merges them.

Two findings recorded for later waves:

- **`nano-ros-board-mps2-an385.cmake` has zero live consumers.** `NANO_ROS_BOARD` is never set to it; the only occurrences are a `FATAL_ERROR` example string and a doc reference. Its header is also stale about being used under FreeRTOS. So `-baremetal` is a correction, not a narrowing.
- **W6's `git grep <retired name>` acceptance needs namespace scoping here.** `mps2-an385` survives legitimately as the QEMU machine argument, the deploy-token alias, the linker script and the PAC crate. Only the index **key** retires.

## Not enforced, deliberately

**R2** (`<where>-<stack>`, stack last). `mps2-an385` and `esp32-c3` pass the gate while §4 still renames them. R1/R3/R4 catch a name stating something **false**; a missing stack suffix misleads nobody, and enforcing it would need `native`, `posix`, `zephyr` and `fvp-aemv8r-smp` carved out — an allow-list carved out of a shape rule is how a gate stops meaning anything. R2 stays W5's job.

`just check fast`: 277/277.


jerry73204:
Rebased onto `main`. The `modify/delete` was resolvable, and the branch's own W7 rule is what decides it.

**What collided.** This branch's W7 commit annotates `docs/issues/1115-nuttx-config-header-stale-vs-regenerated-msg-headers.md` — a board-name note plus a rewrite of its 13-leaf inventory to the new names. `main` deleted that path: `c8ad7c350` **fixed** issue 1115 and archived it as `docs/issues/archived/1115-nuttx-committed-config-fallback-missing-codegen-version.md`, `status: resolved`, `resolved: 2026-09-07`, with a corrected title (the issue had diagnosed stale build-dir residue; the real cause was that the NuttX config header is a COMMITTED snapshot the phase-429 W1 macros never reached).

**Why main's side wins, from this branch's own rule.** W7 draws the rewrite/leave line per occurrence, and its test is what the sentence is FOR. Two of its clauses apply here and now point the same way:

* *"Annotated, not rewritten"* covered 1115 only because it was **open** and its evidence is verbatim compiler output. It is no longer open.
* *"Left as record"* is the rule for everything archived — and `main`'s archived copy does keep `examples/qemu-arm-nuttx/…` verbatim, correctly, because an archived record's text was true when written.

The other half of the branch's edit — the hand-written inventory of 13 affected leaves, which W7 moved on the grounds that *"that is a list a fixer navigates"* — has no fixer to navigate it any more. So the annotation is moot in both halves, and carrying it forward would mean re-creating a live issue that `main` closed.

Resolved by accepting the deletion. Nothing in the tree references the old path (grepped), and the retired-name ledger this branch adds is where a reader lands instead.

**Two other conflicts, both mechanical.** `.config/doc-commit-citations-baseline.txt` — `main` had already removed the same stale QEMU-citation block; took its side. `docs/design/README.md` twice — both sides appended a new RFC row (RFC-0094 on `main`, RFC-0093 here); kept both, ordered by number to match the rest of the table.

**Verified:** `doc-refs`, `issue-ids`, `board-name-reach` (0 violations), `board-vocabulary` (0 mirrored pairs), `markdown-links` all OK; `check-doc-commit-citations` OK — 196 citations across 894 live files resolve. 21 commits before, 21 after.

jerry73204:
Correction to my rebase above: it dropped this branch's tip commit, `fix(box-sync): six more build-output patterns eating tracked source, in five trees`.

Not a rebase decision — my worktree was created from a **stale local branch ref** that lagged `origin` by one commit, so the rebase never saw it and `--force-with-lease` had nothing to compare against: the lease checks the remote you have FETCHED, and my clone had never fetched that tip.

Restored by cherry-pick. 22 commits, matching the branch before I touched it; verified by comparing every commit subject on the old tip against the new one, with none missing. Gates re-run green: `doc-refs`, `issue-ids`, `board-name-reach`, `board-vocabulary`, `markdown-links`, `doc-recipe-refs`.

The issue-1115 resolution described above is unaffected.

jerry73204:
Rebased onto `main` again — main advanced 26 commits, including #548's rewrite of the page this branch renames a path in.

## The one conflict

`book/src/getting-started/workspace-entry-pkg.md`. Not a competing edit: main's `docs(#1107)` **rewrote the page** — the "one Entry pkg per deploy target" model the build system retired, the Zephyr `west build` section, the `nros new entry` scaffolding verb — while this branch's W6 commit only renames `examples/qemu-arm-baremetal/` to `examples/mps2-an385-baremetal/` inside it.

Git presented ~150 lines of prose against ~30, which reads like a design collision and is not one. Resolved by taking **main's page whole** and re-applying the rename to what survived it: exactly one occurrence, `examples/…/rust/talker-rtic/src/main.rs` at line 385. Verified by grepping main's version for both spellings before and after.

That is the right split of authority — #1107 owns what the page SAYS, phase-437 owns what the paths are CALLED, and neither needs to arbitrate the other.

## Verified

22 commits before, 22 after; no-loss check (every commit subject on the pre-push tip present on the new one) came back empty.

`doc-refs`, `doc-recipe-refs`, `issue-ids`, `prose-issue-refs`, `board-name-reach` (0 violations), `board-vocabulary` (0 mirrored pairs), `markdown-links` all OK. `check-doc-commit-citations` OK — 199 citations across 900 live files resolve. `just check fast` 282/284; the two reds are `capability-conditionals` and `xrce-vendored-versions`, which refuse because a fresh worktree has no submodules and no `nros setup` sources — neither is reachable from a docs-and-renames diff.

jerry73204:
Rebased onto `main` (was `1b584b01c`, now `c50d09a90`). **22 commits before, 21 after.** 21 of the
22 replayed clean; one was dropped as superseded, with the evidence below.

## The one conflict: `scripts/dev/ros2-box-sync.sh`

`main`'s `fix(#1243): the box mirror dropped 2914 tracked files, in four independent causes` covers
every case this branch's last commit covers, and covers them a level up — by EXCLUDING the trees
rather than rescuing directories inside them. Checked rule by rule against
`git show origin/main:scripts/dev/ros2-box-sync.sh`:

| this branch's rule | on `main` |
|---|---|
| `--include '/third-party/*/qemu/target/***'` | dead — `--exclude '/third-party/qemu/qemu/'` and `--exclude '/third-party/esp32/qemu/'` (both QEMU checkouts are provisioned from a released tarball; nothing builds them from the mirror) |
| `--include '/third-party/*/qemu/roms/**/build*/***'` | dead, same two exclusions |
| `--include '/external/**/build*/***'` | dead — `--exclude '/external/'` (RFC-0095 D3 retires the directory) |
| `--include '/esp-idf-workspace/**/build-cmake/***'` | `main` has this exact rule, line 207, same issue-1243 rationale. The auto-merge produced it TWICE |
| `--include '/third-party/px4/PX4-Autopilot/**/pymavlink/**/node_modules/***'` | `main`'s `--include '/third-party/px4/PX4-Autopilot/**/node_modules/***'` is a superset with the same PX4 anchoring |

So the commit is dropped rather than partially applied — keeping any of it would have left a rescue
rule inside an excluded tree, which reads as coverage and is not. **The resulting
`scripts/dev/ros2-box-sync.sh` is byte-identical to `origin/main`'s** (`git diff origin/main --` on
that path is empty).

This is the only line the no-loss `comm` reports, and it is that commit.

## `docs(#1107)` vs the phase-437 renames — the split held

`main` rewrote `book/src/getting-started/workspace-entry-pkg.md`. Took `main`'s prose whole; exactly
one old path survived that rewrite and it was renamed in place:

    -`examples/qemu-arm-baremetal/rust/talker-rtic/src/main.rs`
    +`examples/mps2-an385-baremetal/rust/talker-rtic/src/main.rs`

`#1107` owns what the page says, phase-437 owns what the paths are called. No other occurrence of a
retired board name remains in that file.

## Gates

`just check board-name-reach` **OK** — 10 overlays + 14 index entries, **0 known violations awaiting
W4–W6** · `just check board-vocabulary` **OK** — 5 deploy values, 5 board values, each resolves and
is an index key, **0 mirrored pairs identical** (cmake 10 / index 14 / crate 19 / fixture 10) ·
`just check markdown-links` **OK** (788 live docs, every link resolves; 422 dead in archived, at
baseline) · `check doc-refs` **OK** · `check issue-ids` **OK** (1212) · `check prose-issue-refs`
**OK** · `check doc-recipe-refs` **OK** · `check-doc-commit-citations.py` **OK** (198 citations) ·
`cargo +nightly fmt --check --all` **OK**.

`just check fast`: **285 gates, 1 failed** — `check-xrce-vendored-versions`, environmental (the
vendored XRCE trees are not checked out in a fresh worktree). Reproduced identically on a plain
`origin/main` worktree, so it is not from this branch.

Retired board names still appear in `docs/issues/archived/**`, the audit findings and the RFCs.
That is W7's deliberate split — *"the live docs say the current name; the record keeps its own"* —
and `check-board-name-reach` is green over it.

PR #742, opened 2026-09-07
feat(phase-412 W3b): the arena's QoS depth becomes a knob, and the model splits in two

jerry73204:
**Stacked on #738** (base is `work/412-depth-wiring`), because it changes the same expression.

#738 made the arena read the **declared** per-endpoint depth. That answer needs a
`<stem>.contract.yaml` stating `qos.depth` for *every* endpoint — one unannotated subscription and
cmake refuses, correctly, because a table over the annotated subset sizes the image from part of
itself. **No in-tree image has such a file**, so the 91,080 bytes #738 measured were unreachable
in practice.

`NROS_PUBSUB_QOS_DEPTH` is the image-wide number, on the ladder every derivable knob follows:
`env > Kconfig/board > declared > builtin`. An image whose subscriptions are all KEEP_LAST(1)
states one number instead of authoring a contract, and gets the same arena.

The builtin stays ROS 2's own default — `rmw_qos_profile_default`'s KEEP_LAST(10), what a
subscription created with no QoS argument actually gets — so an image that says nothing is still
sized the way ROS would size it.

## The model now emits two depths

`arena.rs` asks two different questions and #738 had them sharing one answer:

- **`QOS_DEPTH`** — the *restatement* of ROS 2's default, so
  `the_modelled_qos_depth_is_the_runtime_default` can catch it drifting from
  `QoSProfile::default().depth` (a build script can't read that const out of the crate it's
  building). It must stay the builtin whatever the image budgeted, or the drift check starts
  asserting against a configuration instead of against ROS. Mutation-verified: changing the
  restatement to 7 still fails that test.
- **`BUDGETED_QOS_DEPTH`** — what the arena was actually sized for.
  `the_modelled_pubsub_region_covers_a_default_qos_subscription` now measures against it.

That split was **found by running the tests with the knob set**: #738 emitted the resolved depth as
`QOS_DEPTH`, and both tests failed — one of them asserting that a configured image matches ROS's
default, which is not a question that image is asking. What the retargeted test still catches is
the thing that matters: a budget that does not cover its own claim.

## Measured

Reference island entity shape (11 subscriptions, 4 timers, 2 service servers):

| | `ARENA_SIZE` | `BUDGETED_QOS_DEPTH` |
| --- | ---: | ---: |
| builtin (nothing stated) | 144,584 | 10 |
| `NROS_PUBSUB_QOS_DEPTH=1` | **53,504** | 1 |
| `NROS_DECLARED_MAX_QOS_DEPTH=1` | 53,504 | 1 |
| declared 1, knob 4 | 76,472 | 4 |

The knob outranks the declaration, as the ladder requires. `QOS_DEPTH` stays 10 throughout.

Kconfig help states the trade in full, including that **too small is not a smaller buffer**: an
arena that does not fit halts during entity creation, before the first spin, so the advisory issue
0900 added never prints.

```
just check fast   275 ran, 1 ledger skip, 0 failed
just test-unit    PASS
```

### DIFF
--- a/packages/core/nros-node/build.rs
+++ b/packages/core/nros-node/build.rs
@@ -316,7 +316,30 @@ fn main() {
     // variable means "some endpoint said nothing" and the default stands. That
     // guard belongs there rather than here: the count is a property of the
     // declaration, and this lane cannot see it.
-    let pubsub_depth = declared_max_qos_depth().unwrap_or(PUBSUB_QOS_DEPTH);
+    // phase-412 W3b — and this is the whole ladder, in one expression.
+    //
+    // `env_usize` supplies the three rungs above the builtin (an environment
+    // value, `CONFIG_NROS_PUBSUB_QOS_DEPTH` from `$DOTCONFIG`, then the
+    // board/platform descriptor); the DECLARED depth sits between those and the
+    // literal, which is where the tree's stated precedence puts a derived
+    // value: env > Kconfig/board > derived > crate default.
+    //
+    // WHY A KNOB EXISTS BESIDE THE DECLARATION. The declared depth needs a
+    // `<stem>.contract.yaml` stating `qos.depth` for EVERY endpoint — one
+    // unannotated subscription and cmake refuses, correctly, because a table
+    // over the annotated subset sizes the image from part of itself. That is a
+    // high bar for a saving worth 91,080 bytes on the island shape, and an
+    // image whose subscriptions are all KEEP_LAST(1) should not have to author
+    // a contract to say one number. This lets it say the number.
+    //
+    // The BUILTIN stays ROS 2's own default: `PUBSUB_QOS_DEPTH` is
+    // `rmw_qos_profile_default`'s KEEP_LAST(10), which is what a subscription
+    // created with no QoS argument actually gets, so an image that says nothing
+    // is still sized the way ROS would size it.
+    let pubsub_depth = env_usize(
+        "NROS_PUBSUB_QOS_DEPTH",
+        declared_max_qos_depth().unwrap_or(PUBSUB_QOS_DEPTH),
+    );
     let pubsub_region = buffered_region(pubsub_depth, rx_recv_size);
     let pubsub_entry = pubsub_region + PUBSUB_ENTRY_STRUCT;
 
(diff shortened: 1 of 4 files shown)


jerry73204:
Rebased so this targets `main` cleanly and carries #738's **rebased** commit rather than its old one.

## What was wrong

The branch held `ac6e03ae4` — #738's W3 commit as it stood before #738 was itself rebased onto `main`. Replaying that against today's `main` conflicted in three files that #738's own rebase had already resolved: `.config/doc-commit-citations-baseline.txt`, `cmake/NanoRosEntityFacts.cmake` and `scripts/check/config-knob-census.py`. Re-resolving them here would have produced a second, independently-authored answer to a question already settled one PR over — and two different resolutions of the same conflict is how a patch-id stops matching.

So the replay is `--onto origin/work/412-depth-wiring`, which takes #738's post-rebase `59281d30a` as-is and replays only W3b on top. Base stays `main`, so auto-merge arms; the carried commit drops by patch-id when #738 lands, because it is now byte-identical to what will land.

## The one real conflict

`scripts/check/config-knob-census.py` — both sides append a knob to the same dict at the same offset:

- `main`: `ZPICO_GRAPH_CACHE_SIZE` — sized by the PEER graph, so stated rather than derived.
- W3b: `NROS_PUBSUB_QOS_DEPTH` — the arena's per-subscription history depth, which outranks W3's derived per-endpoint `NROS_DECLARED_MAX_QOS_DEPTH`.

Different knobs, same classification (`sizing`), no overlap. Kept both — the git conflict was positional, not semantic.

## Verified

`config-knob-census` OK, `doc-refs` OK, `doc-recipe-refs` OK, `issue-ids` OK, `check-doc-commit-citations` OK (194 citations across 891 live files resolve), and `tests/cmake-entity-inventory-tests.sh` 39/39 including its own "the declared QoS DEPTH crosses the lane boundary (phase-412 W3)" case.

2 commits before, 2 after.

jerry73204:
Rebased onto `main` (`632bdbeba`).

**2 commits -> 1.** `feat(phase-412 W3): the declared QoS depth reaches the arena` was #738's, and #738 has landed — git skipped it by patch-id (`warning: skipped previously applied commit 59281d30a`), and the same subject is on `origin/main` as `e4b576489`. Verified nothing else dropped: the subject diff between the pre-push tip and the rebased tip is exactly that one line.

**Conflicts: none.** The rebase applied W3b cleanly; `scripts/check/config-knob-census.py` did not conflict — the `NROS_PUBSUB_QOS_DEPTH` entry appends to `KNOB_CLASS` at a point `main` did not touch.

**Gates:** `just check fast` — 282 of 283 green; the one red is `xrce-vendored-versions`, which fails identically on a plain `origin/main` worktree ("no vendored tree checked out", it wants `nros setup --source …`), so it is environmental, not this diff. Also green: `doc-refs`, `issue-ids`, `prose-issue-refs`, `doc-recipe-refs`, `check-doc-commit-citations.py`, `cargo +nightly fmt --check --all`, `cargo check -p nros-node --all-targets --features std`.

PR #751, opened 2026-09-07
feat(phase-430 W1+W2): a silent /clock says so once, and use_sim_time is declared everywhere

jerry73204:
**phase-430 W1 and W2 only.** The rewritten phase doc is not on `main` — it rides on PR #629 (`phase-428-w10-qos-ssot`) — and this PR does not touch it. W3 and W5 landed as #712 and #716.

## W2 — declare `use_sim_time = false` on every node

rclcpp auto-declares it in every node's constructor, so `ros2 param list` shows it and `ros2 param set <node> use_sim_time true` works against a node whose author never wrote a line about simulated time. Ours honoured the parameter only if the APP had declared it: the store refuses an undeclared set (`allow_undeclared` is false), so **the runtime switch worked on exactly the nodes that least needed it**. That is the PARTIAL the survey measured (row 5).

`ensure_parameter_store` now seeds `Bool(false)` when `sim-time` **and** `param-services` are both on. Gated on `sim-time` too: an image that cannot act on the parameter should not advertise it — a settable knob that does nothing is worse than an absent one.

### The seed is a placeholder, and that is load-bearing

`ParameterServer::declare` refuses a name it already holds, so a naive seed turns an app's own `declare_parameter("use_sim_time", Bool(true))` into a refusal and **the default silently beats the app**. Measured, not reasoned — with `yield_seeded_use_sim_time` stubbed out, six tests fail, four of which predate this work:

```
use_sim_time_attaches_and_detaches_the_clock_source ... FAILED
use_sim_time_declared_with_a_descriptor_attaches_the_clock_source ... FAILED
use_sim_time_follows_the_store_verdict_not_the_caller ... FAILED
an_apps_own_use_sim_time_declaration_beats_the_auto_declared_default ... FAILED
```

So `sim_time_param_seeded` tracks whether the stored value is ours; the first real declaration removes the placeholder before declaring, with its own value and its own descriptor. A declaration then refused on its own terms (full store, ill-formed range) puts the default back, so the name is never left undeclared.

The seed deliberately does NOT go through `note_reserved_parameter`, so `sim_time_stated` stays false: **an auto-declared default is not a statement by anyone**. Recording it as one would make every executor with parameter services write `set_active(false)` onto a PROCESS-GLOBAL gate on its first spin — the collision `sim_time_stated` was added for (issue 1104).

### The class, swept

```
grep -n 'new_param_state\|ensure_parameter_store' packages/core/nros-node/src/executor/spin.rs
```

Two doors into the store, and only one was the declare path: `register_parameter_services` built its own `ParamState`, so a hook on the declaration path alone would have missed the node W2 is really for — one that declares NOTHING. It routes through `ensure_parameter_store` now (issue 0745's preserve-the-store and issue 0756's never-by-value both intact).

### phase-426 assumption

Written against `main`. phase-426 W1–W3 (store keyed by NODE, six services per node) is on `origin/phase-426-w{1,2,3}-*` and **not merged**. Its three touch points are `declare_parameter_on(node, …)`, `declare_parameter_with_descriptor_on`, and a `NodeKey`-taking `declare`/`get_bool`; this PR's hooks sit at those same three seams, so the rebase is a signature change at each rather than a redesign. The seed becomes per node, which is what upstream does anyway.

## W1 — `use_sim_time` true and no `/clock` ever: say so once

An image told to use simulated time that never receives a sample looks **identical** to one whose simulator is merely paused: both leave every ROS-time timer sitting still, silently. Only the image can distinguish them — "no sample has EVER arrived" is a fact about this subscriber, not about the topic.

```
use_sim_time is true but no /clock sample has arrived in 5s: ROS-time timers are
running on SYSTEM time meanwhile. Publish rosgraph_msgs/msg/Clock (ros2 bag play
--clock), or set use_sim_time false.
```

198 bytes, inside `nros_log`'s 256-byte call-site budget, both actionable identifiers in the first clause — the lesson `arena::report_arena_headroom` paid for. `nros_log`, never stdio (issue 0589). Not a behaviour change: row 15 stands.

### Where the threshold lives

**The executor's own per-spin `delta_us`** — the same quantity every wall timer on the same executor accumulates, including its "no clock, credit the requested timeout" fallback. The other two candidates lose:

* A **spin count** needs no clock and means nothing: a `spin_once(5ms)` loop and a busy poll differ by four orders of magnitude, so any constant is right for one image and absurd for every other. That is per-image folklore, which is what a diagnostic exists to replace.
* A **wall duration** is the obvious choice and has the obvious flaw: on a target running under simulation the wall clock is part of the fiction too — `native_sim` runs uptime as fast as the host allows when idle, QEMU under `-icount` as slowly as the model dictates — so five host seconds are not five seconds of anything the image experienced. Worse, an executor may have **no clock at all** (`now_us()` returns `Option`, `None` on a freestanding build with no injected hook), so a wall threshold would be silently UNREACHABLE on exactly the targets this is for.
* The **first ROS-time timer coming due** is the semantically sharpest trigger and too narrow: a node that only stamps messages off `Clock::ros_time()` is just as misconfigured and would never hear a word.

`delta_us` buys three properties at once: the warning arrives after as much time as the image's own wall timers believe has passed, so it is legible against them; it moves WITH a simulated wall clock instead of against it; and a clockless image is not exempt, because the fallback still advances it. The whole argument lives on `time_source::SILENCE_WARN_US`, not in this PR body.

### Once, not per spin

The report disarms the watch, so the record cannot become the noise it warns about. A sample arriving before the threshold disarms it **permanently**: the configuration is then proven right, and a simulator that stops afterwards is a PAUSE — warning about a pause is the false alarm that teaches everyone to ignore the line. `use_sim_time` going false and back to true re-arms with a fresh interval.

Scope is the source the EXECUTOR installed for `use_sim_time`. An app calling `NodeCtx::install_ros_time_source` directly is not watched, because this line names `use_sim_time` and that app never set it.

## Tests

Asserted through `Executor::sim_time_silence_warnings()`, not a log grep: the contract is "exactly once", so the assertion must distinguish one from two, and `nros_log`'s sink is not present on every target that runs this. The accessor is `sim-time`-gated, so it adds no ledger row — the parity extractor builds the Rust surface without `sim-time` (`NROS_FEATURES` in `scripts/api_parity/extract_rust.py`), which is the survey's finding C.

| test | covers |
| --- | --- |
| `an_unarmed_watch_never_reports` | nothing before attach |
| `a_silent_source_reports_once_at_the_threshold` | the interval, and the one-shot |
| `a_sample_before_the_threshold_silences_the_watch_for_good` | a pause is not a misconfiguration |
| `re_arming_restarts_the_interval` | no carry-over between armings |
| `use_sim_time_with_no_clock_sample_is_reported_exactly_once` | the executor wiring, all four |
| `use_sim_time_is_declared_on_a_node_that_never_named_it` | W2, `ros2 param list` + the wire set |
| `use_sim_time_is_declared_by_registering_the_parameter_services_alone` | W2 via the second door |
| `an_apps_own_use_sim_time_declaration_beats_the_auto_declared_default` | the app wins |
| `an_apps_use_sim_time_declaration_with_a_descriptor_beats_the_default` | …descriptor path too |

The executor W1 test rides a `private_test_clock!` (issue 1105) so it crosses five seconds without sleeping five seconds — `test_clock_us` is free-running and shared, so the interval would otherwise measure how much of the suite ran first.

### Negative controls

Three mutations, each on a temporary WIP commit (never `git stash` — the stash stack is shared), tests unchanged:

**A — the watch is never armed:**
```
---- use_sim_time_with_no_clock_sample_is_reported_exactly_once ----
assertion `left == right` failed: use_sim_time is on, /clock has never spoken, and
the image said nothing -- which is exactly the silence that reads as a paused simulator
  left: 0
 right: 1
test result: FAILED. 381 passed; 1 failed
```

**B — no auto-declare:**
```
---- use_sim_time_is_declared_on_a_node_that_never_named_it ----
assertion `left == right` failed: rclcpp declares use_sim_time=false on every node;
a `ros2 param list` here showed nothing until the app named it
  left: None
 right: Some(false)
test result: FAILED. 379 passed; 3 failed
```

**C — the seed never steps aside** (the failure mode the auto-declare invites):
```
---- an_apps_own_use_sim_time_declaration_beats_the_auto_declared_default ----
the app's own declaration was refused because the auto-declared default had taken the name
test result: FAILED. 376 passed; 6 failed
```

## Ran

| command | result |
| --- | --- |
| `cargo test -p nros-node --lib --features std,sim-time,param-services` | **382 passed**, 0 failed, 2 ignored |
| same, `--features std,sim-time` (no param-services) | 350 passed, 0 failed |
| `cargo build -p nros-node` at `std` / `std,param-services` / `std,sim-time` | clean, no warnings |
| `cargo clippy -p nros-node --all-targets --features std,sim-time,param-services -- -D warnings` | clean |
| `just check no-std-stdio` | OK (192 no_std crates) |
| `just check no-vacuous-tests` | OK (306 test files, 10 self-test cases) |
| `just check api-parity` | OK — every divergence carries a ledger entry, 0 new rows needed |
| `just check fast` | 273 of 277; four reds, all environmental — see below |

The four `check fast` reds read no file in this diff and are properties of this worktree, not of the change:

* `capability-conditionals` — `zenoh-pico` submodule not checked out here (the gate says so and exits 2).
* `xrce-vendored-versions` — no vendored XRCE tree (`nros setup --source …` not run).
* `xrce-source-manifest` — its own self-test does `Path("/nonexistent/…").exists()`, which raises `PermissionError` under this sandbox rather than returning a verdict about the tree.
* `doc-commit-citations` — `docs/issues/README.md:718` cites `22e511442`, a commit this checkout does not have, plus four baseline entries that now resolve. Inherited; untouched by this PR.

Tier: `just ci gate`-shaped (compile + unit, no fixtures), narrowed to the crate this touches. No fixture, SDK, QEMU or cross toolchain involved.

## Book

`book/src/user-guide/simulated-time.md` gains the W2 default under "Turning it on" and the W1 diagnostic under "Limits", which is what both work items' acceptance asks for.

### DIFF
--- a/book/src/user-guide/simulated-time.md
+++ b/book/src/user-guide/simulated-time.md
@@ -58,6 +58,18 @@ talker:
 `use_sim_time` is reserved: nothing reads its value, the runtime acts on it.
 Set it false and `/clock` samples stop being installed.
 
+You do not have to declare it first. An image built with parameter services
+declares `use_sim_time = false` on your behalf, as rclcpp does on every node, so
+`ros2 param list` shows it and
+
+```console
+$ ros2 param set /talker use_sim_time true
+```
+
+reaches a node whose author never wrote a line about simulated time. Declaring
+it yourself still wins — your default, and your descriptor, replace the
+auto-declared one.
+
 The Rust crate needs the `sim-time` feature, because the subscription costs an
 entity slot and an RX buffer:
 
--- a/book/src/user-guide/simulated-time.md
+++ b/book/src/user-guide/simulated-time.md
@@ -137,6 +137,26 @@ deliberate: `Clock::clear_ros_time_override()`.
 model `nros_core::Clock` has always documented. Two nodes in one image cannot be
 on different simulated times.
 
+**A never-connected `/clock` looks exactly like a paused one, so the image says
+so once.** Both leave every ROS-time timer sitting still. Five seconds after the
+source attaches with no sample received, the image logs one warning at `WARN`,
+naming `use_sim_time` and `/clock` and the fact that ROS-time timers are running
+on system time meanwhile:
+
+```text
+use_sim_time is true but no /clock sample has arrived in 5s: ROS-time timers are
+running on SYSTEM time meanwhile. Publish rosgraph_msgs/msg/Clock (ros2 bag play
+--clock), or set use_sim_time false.
+```
+
+Once, not per spin, and never again once a sample has arrived — a simulator that
+publishes and then pauses is the case the line exists to be distinguished from.
+Turning `use_sim_time` off and on again arms it afresh. The five seconds are
+measured on the executor's own per-spin elapsed time, the same one your wall
+timers accumulate, so an image whose wall clock is itself simulated gets the
+warning at a point it can relate to its own timers. `Executor::sim_time_silence_warnings()`
+answers the same question in code.
+
 **No jump callbacks.** `rclcpp::Clock::create_jump_callback` and the
 `rcl_jump_threshold_t` machinery have no counterpart. The behaviour a jump
 produces is fixed (above); you cannot subscribe to the event.
--- a/packages/core/nros-node/src/executor/spin.rs
+++ b/packages/core/nros-node/src/executor/spin.rs
@@ -7559,6 +7559,20 @@ impl<'s> Executor<'s> {
                     self.params = Some(state);
                 }
             }
+            // phase-430 W2 — "declare `use_sim_time` on EVERY node, as rclcpp
+            // does", said where the nodes finally exist. This is the only place
+            // that knows the full set: `ensure_parameter_store` runs before any
+            // node record does, so it can only seed the implicit PRIMARY, and
+            // the further nodes an image composes onto one executor arrive
+            // later. Seeding as each set of six goes up keeps the two in step —
+            // the parameter is declared on exactly the nodes a
+            // `ros2 param set <node> use_sim_time true` can reach.
+            //
+            // Idempotent by construction: `declare` refuses a name the node
+            // already holds and the placeholder flag is only raised on
+            // acceptance, so re-seeding PRIMARY here neither duplicates the
+            // entry nor re-marks an application's own declaration as ours.
+            self.seed_use_sim_time_default(super::node_record::NodeId::from_raw(index));
         }
         Ok(())
     }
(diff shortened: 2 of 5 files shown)


jerry73204:
## Rebased onto `main` (phase-426 W3 landed underneath this) — and W2 needed a port, not a conflict resolution

`main` moved the ground this PR's second commit stood on. Force-pushed
`038f73c9` → `7097f8cb`, same two commits, no drops
(`comm -23 <(git log --format=%s origin/main..038f73c99|sort -u) <(git log --format=%s origin/main..HEAD|sort -u)` is empty; 2 commits before, 2 after).

### What moved underneath

Everything is `94bfdf1fc feat(phase-426 W3): six parameter services per NODE, not one per executor`
(plus phase-426 W1's node-keyed store, which was already under this branch's base and
which **this branch had never been compiled against** — `cargo check -p nros-node
--all-targets --features std,sim-time,param-services` on the pre-rebase tip is
**12 × E0061**, 2 in `spin.rs` and 10 in `tests.rs`. So the PR was red on arity before
the merge conflict, not only because of it):

* `ParamState::services` is a `heapless::Vec<Box<dyn ParamServiceProcessor>, MAX_PARAM_SERVICE_SETS>` — one set of six per node — not an `Option`.
* the six servers are no longer built in `register_parameter_services`; that records `ParamState::requested` and `reconcile_parameter_services` publishes one set per node at the head of a spin.
* `new_param_state()` takes no argument; `ParameterServer::{declare,remove,get_bool,get_descriptor,apply,list_names}` all take a `NodeKey`.

### The two conflicting hunks

1. **`spin.rs` ~7744** — git had mis-aligned two unrelated regions: `main` added
   `parameter_service_node_names()` (W3) exactly where W2 rewrote the registration
   tail. Resolved to **`main`'s side alone**, and that is not a drop. W2's tail rewrite
   existed to route `register_parameter_services` through `ensure_parameter_store`
   instead of building its own `ParamState` — the "two doors into the store" sweep in
   its commit message. **W3 already closed that door** for its own reasons:
   `register_parameter_services` now calls `ensure_parameter_store()` on its first
   line. W2 inherits the single door rather than authoring it, and
   `parameter_service_node_names` survives untouched.
2. **`spin.rs` ~8017** — `ensure_parameter_store`. Kept `main`'s
   `Self::new_param_state()` and re-attached the seed.

### The design question: per node, or `NodeId::PRIMARY` only

**Per node.** The commit is titled "declare use_sim_time on every node, as rclcpp
does", and its own body predicted this: *"The seed would then be per node, which is
what upstream does anyway."* Before phase-426 the store was one flat table per
executor, so "every node" and "the executor" were the same sentence and the
distinction was not expressible. It is now, and the narrow reading would make the
title false.

It is also **measured**, not argued. A new test —
`use_sim_time_is_declared_on_every_node_the_services_are_published_under` — builds a
two-node executor, registers the services, and asserts `parameter_service_node_names()
== 2` plus `use_sim_time == Some(false)` on both keys. Stubbing each half back to the
narrow shape fails it:

```
seed NodeId::PRIMARY only, in the reconcile loop:
  assertion `left == right` failed: node 1 has parameter services but no use_sim_time;
  `ros2 param set` on it would be refused as undeclared, which is the pre-W2 behaviour
  this work item removes
    left: None     right: Some(false)

one SHARED placeholder flag (the original single `bool`):
  beta's placeholder must still be a placeholder: a declaration on a SIBLING node is
  not a declaration on this one
```

That second control is the reason the flag had to move too. With one `bool` per
executor, alpha's own `declare_parameter_on(alpha, "use_sim_time", true)` spends the
flag, and beta's identical declaration is then **refused** — an app that named the
parameter on one node of an image would silently lose it on the others.

### How it is re-expressed

* **`seed_use_sim_time_default(node)`** and **`yield_seeded_use_sim_time(node, name)`** take a `NodeId`.
* **Two seeding sites, because there are two moments a node can exist:**
  `ensure_parameter_store` seeds the implicit `PRIMARY` (it runs *before* any node
  record does — that is the whole of issue 0745, so it cannot see the others), and
  `reconcile_parameter_services` seeds each node key **as its set of six goes up**. The
  parameter is therefore declared on exactly the nodes a
  `ros2 param set <node> use_sim_time true` can reach — the set
  `parameter_service_node_names()` enumerates. Idempotent by construction: `declare`
  refuses a name the node already holds and the flag is only raised on acceptance, so
  re-seeding `PRIMARY` in the reconcile neither duplicates the entry nor re-marks an
  app's own declaration as ours.
* **The placeholder is `ParamState::sim_time_seeded: [bool; MAX_PARAM_SERVICE_SETS]`**,
  not a field on `Executor`. Two reasons: it is a fact about the *store's* contents,
  and it is now a table indexed by node key —
  `storage::tests::the_executor_value_does_not_scale_with_the_knobs` (issue 0961) is
  the test that says a knob-scaled table does not belong in the `Executor` value.
  `ParamState` is boxed, and the bound is the one `services` already uses.

### Invariants checked, not assumed

* **Issue 1104 / `note_reserved_parameter`** — the seed still bypasses it, on *every*
  node rather than just the first: N seeded nodes must still be N statements by
  nobody, or the process-global-gate collision comes back multiplied. Both W2 tests
  that assert `!ros_time_source_installed()` after a seed still pass, and the new
  multi-node one asserts it too.
* **Issue 0745** — `reconcile_parameter_services` still preserves an initialized store; the seed is additive.
* **Issue 0756** — no `ParamState` is built by value anywhere; `new_param_state` is untouched apart from one array initializer.
* **Feature gating** — unchanged: the whole impl block is `param-services`, the seed body is `all(sim-time, any(has_rmw, test))`.

### Tests

No test premise stopped holding, so nothing was deleted. Ten call sites in the four W2
tests and two in the W1 test were re-spelled to the node-keyed API using the
convention the surrounding phase-430 W3 tests already use
(`super::node_record::NodeId::PRIMARY.into()`); one test was added.

```
cargo test -p nros-node --lib --features std,sim-time,param-services
  => 391 passed, 0 failed, 2 ignored     (was 385 before W1 replayed, 379 pre-PR)
cargo check -p nros-node --all-targets --features std          => clean
cargo clippy -p nros-node --all-targets -D warnings, for
  std | std,param-services | std,sim-time | std,sim-time,param-services  => clean
cargo +nightly fmt --check --all                               => clean
just check fast  => 282/283; the one red is `xrce-vendored-versions`, which
  fails identically on a pristine `origin/main` worktree ("no vendored tree checked
  out — run `nros setup --source micro-cdr --source micro-xrce-dds-client`"), i.e.
  environmental. `capability-conditionals` passes once zenoh-pico is checked out.
```

### One commit message edited

W2's body carried a section headed *"Assumption about phase-426"* stating that
phase-426 W1–W3 was unmerged and that the port would be "a signature change at each
[seam], not a redesign". Half of that is now false and the other half understates it,
so the section was replaced with what actually landed — the two seeding sites, the
per-node flag, where it lives and why, and the two negative controls above. The tree
is byte-identical either way (verified against the pre-amend commit).

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01PU5GN5rrqV1fiCwCztA45N

PR #753, opened 2026-09-07
phase-427 W10 [rust]: the executor opens a session, the node is named after

jerry73204:
phase-427 **W10** — the executor opens a session, the node is named after.
Stacked on #704 (W9: `Context` on every target, `InitOptions`, the rclrs `init`
family); base is `phase-427-w9-context-every-target`.

RFC-0089 "Context and `init`, settled" gives Rust the rclrs sequence — context,
then executor, then node. W9 landed the first. This lands the other two, so a
ported `executor.create_node("talker")` finally has a `create_executor` above it.

## What landed

**`Context::create_executor()`** (`alloc` + `rmw-cffi`) → `Result<Executor<'static>, InitError>`.
rclrs's name in rclrs's place, `adopt-bounded` on two counts the doc states and
the compiler names: no runtime argument (RFC-0002 puts one executor per RTOS
task, so `create_executor(rt)` is an arity error and `create_basic_executor()`
is `no method named`), and it returns a `Result`, because ours OPENS THE SESSION
and upstream's does not.

**`Context::create_executor_in(backing)`** — ours, the freestanding form.
Storage supplied rather than leaked, so an entry with a `static mut
[MaybeUninit<u64>; N]` allocates nothing. It is the SAFE wrapper around `unsafe
Executor::open_in`, so it enforces the precondition that makes that one unsafe:
a short backing is `InitError::BackingTooSmall { needed, given }`, naming both
numbers, refused **before** the RMW is consulted — with the order reversed a
caller whose array is short would be told the router is missing.

**`Executor::create_node(name)`** already existed and is unchanged; phase-376
W5/B1 gave it the `(name, namespace)` dedup that makes it a registration rather
than a handle factory. What it gains is the sequence it sits in, and a doc
saying so.

**`ExecutorConfig::node_name`** (the builder method, not the field) deprecated,
message naming `create_node`.

**`NodeError` gains `Display` + `core::error::Error`.** RFC-0089 says the port's
error difference is "invisible under `?` and loud under a `match`"; that was
half true, because `?` into a `Box<dyn Error>` needs `Error`, `Error` needs
`Display`, and this enum had neither — so a ported `main` could not keep one
error type at all.

## The deprecation shim, and why it is feature-armed

`#[cfg_attr(feature = "deprecate-node-name-setter", deprecated(...))]`, forwarded
by `nros/deprecate-node-name-setter`. The same shape `nros-log`'s retired
`deprecate-legacy-names` had.

This workspace sets `warnings = "deny"` (root `Cargo.toml`,
`[workspace.lints.rust]`), so an always-armed `#[deprecated]` is not a nudge
here — it is **65 hard errors across 47 files** that have not migrated
(`grep -rn '\.node_name(' --include=*.rs packages/ examples/`), in tests,
benches, boards, bridges and example leaves. Landing that as
`#[allow(deprecated)]` at every one of those sites would suppress the signal in
exactly the code the deprecation is aimed at. Build with the feature to find
your call sites; the feature and the setter go together when the sweep is done.
`Context::config` carries the one legitimate internal use and says so.

## Acceptance: the count is MEASURED, and it is not two

`packages/testing/nros-tests/tests/rclrs_talker_port.rs` holds the upstream text
and the port in two files and diffs them. The ported file is `include!`d as well
as `include_str!`d from ONE path, so "it ports" is a build-time fact about the
same bytes the count is computed from.

```diff
--- upstream_rclrs_talker.rs.txt
+++ ported_talker.rs
@@ -1,13 +1,13 @@
-use rclrs::*;
-use example_interfaces::msg::String as StringMsg;
+use nros::*;
+use nros_std_msgs_diag::msg::String as StringMsg;

-fn main() -> Result<(), RclrsError> {
+fn main() -> Result<(), Box<dyn core::error::Error>> {
     let context = Context::default_from_env()?;
-    let mut executor = context.create_basic_executor();
-    let node = executor.create_node("talker")?;
+    let mut executor = context.create_executor()?;
+    let mut node = executor.create_node("talker")?;

     let publisher = node.create_publisher::<StringMsg>("chatter")?;

-    executor.spin(SpinOptions::default()).first_error()?;
+    executor.spin_blocking(SpinOptions::default())?;
     Ok(())
 }
```

**Six lines, not two.** The test records the disagreement rather than the
target, with the wave that owns each:

| kind | n | what |
| --- | --- | --- |
| import | 2 | the crate a name comes from. Any port has these. |
| error type | 1 | the `main` signature — POSSIBLE AT ALL only as of this wave. |
| mutability | 1 | `let mut node`. Ledger `rust:Executor::create_node` records why; rustc names the binding. |
| **predicted** | **2** | the two `?`s RFC-0089 names. These are W10's, and they are now the only two W10's half of the port costs. |

The `spin` line additionally renames `spin` → `spin_blocking`, which W10 does
not own: `Executor::spin` is taken here by `spin(Duration) -> !`, the body of an
RTOS task, and RFC-0089's target shape moves the `SpinOptions` form onto that
name in a later wave. It is on the same line, so it costs no extra one.

Upstream provenance is in the test's header: the `rclrs/src/lib.rs` "Basic
Usage" skeleton on `ros2-rust/ros2_rust@main`, plus the `create_publisher` line
from `NodeState::create_publisher`'s own doc example (the crate-level one
subscribes; a talker publishes).

## Ledger

* `rust:Context::create_executor` — `declined`/`absent` → `divergence` /
  **`adopt-bounded`**.
* `rust:Context::create_executor_in` — new, `extension`. No disposition, for the
  reason `rust:init_with_launch` gives.
* `rust:Executor::create_node` — gains **`adopt`**.
* `rust:InitError`, `rust:NodeError` — what each gained.

## Run

| lane | result |
| --- | --- |
| `cargo test -p nros --features env,std --lib` | 84 passed (`--test-threads=1`; see note) |
| `cargo test -p nros-node --lib --features std` | 338 passed, 2 ignored |
| `cargo test -p nros-tests --test rclrs_talker_port --test context_create_executor --features rclrs-port-test` | 9 passed |
| `cargo clippy -D warnings` — `nros` (`--all-targets`), `nros-node --lib` (armed and unarmed), `nros-tests --tests` | clean |
| `cargo check -p nros --no-default-features --features alloc` | clean |
| `cargo check -p nros --no-default-features --features alloc[,rmw-cffi] --target thumbv7m-none-eabi` | clean |
| `just check api-parity` / `api-parity-ledger` | green |
| `just check fast` | **green**, 253 gates, 7 environmental SKIPs (no in-tree `nros` CLI, no zenoh-pico submodule in this worktree) |

Two pre-existing reds, neither touched by this branch:

* `cargo test -p nros --lib` **parallel** flakes on
  `init::ros_args_refusal_tests::plain_args_pass_through_to_init` —
  `src/env.rs`'s tests call `std::env::set_var("ROS_DOMAIN_ID", ...)` and race
  `init.rs`'s readers. Green at `--test-threads=1`.
* `cargo clippy -p nros-node --all-targets` fails
  `clippy::assertions_on_constants` at `executor/tests.rs:424`, which this
  branch does not touch. `--lib` is clean.


jerry73204:
## Rebased onto `main` (CONFLICTING/DIRTY → MERGEABLE)

**58 commits before, 41 after.** 19 dropped, 2 added.

### The 19 dropped commits

17 are patch-id drops the queue's rebase-merge already made — each subject is on
`origin/main` (`git log origin/main --grep=<subject>` returns exactly 1): the eight
`design(RFC-0089)` commits, the RFC fold, the three `fix(#417)` commits, and
`docs(#1087, #1088)` / `docs(#1092)` / `docs(phase-426…)` / `docs(phase-427, phase-428…)` /
`docs(phase-430, RFC-0089…)`.

Two are **not** on `main` by subject and were dropped deliberately, because `main`
answered the same question later and differently:

| dropped | why |
| --- | --- |
| `feat(phase-428 W6): [[nodiscard]] on Result` | `main` landed it as part of phase-427 W8's `ResultOf<T>` refactor: `NROS_NODISCARD` is on `ResultOf<void>`, `ResultOf<T>` and the deprecated `Expected<T>` (`result.hpp:181,320,367`). Resolving the conflict to `main` left the commit byte-identical to `HEAD`, i.e. empty. |
| `docs: correct the [[nodiscard]] claim — it is a signal here, not an enforced gate` | Its claim is now **false**. It says "a discarded `Result` is a `-Wunused-result` WARNING and the lane stays green through it". `main`'s `cpp` lane compiles `packages/api/nros-cpp/tests/compile/result_nodiscard_probe.cpp` and **fails the build if it compiles** (`just/check/lanes.just:1039-1040`). Landing the correction would have re-introduced a statement the gate disproves. |

### Non-mechanical conflict resolutions

Most conflicts were "`main` rewrote the surrounding text, this branch added one line" and
resolved to `main` + the branch's line. These needed judgement:

**1. `#[must_use]` × `main`'s node-keyed parameter store.** phase-428 W6 marks
`ParameterServer::declare` / `declare_with_descriptor` and `Executor::declare_parameter*`
`#[must_use]`; `main` changed those signatures to take a `NodeKey`/`NodeId` first
(`declare(node, name, value)`). Kept `main`'s signature and the branch's attribute + doc,
and re-wrapped every conflicting call site as `assert!(server.declare(NODE, …))`. Two of
the resulting lines cross 100 columns, so there is a separate `style(rebase)` commit that
is `cargo +nightly fmt` and nothing else.

**2. `executor/tests.rs` `use_sim_time` test.** The branch wrapped a second
`declare_parameter` in `assert!(!…)`. `main` (phase-430 W3 / issue 1202) **rewrote the
test** to go through the store's `apply()` instead, and the second `declare_parameter`
call the branch was annotating no longer exists. Kept `main`'s test verbatim.

**3. RFC-0089.** The branch's fold commit and `main` both restructured this file. `main`
carries content the branch never had and in three places directly supersedes it —
`ResultOf<T>` (the branch still said `Result<T>`, which `main` records as ill-formed on
gcc 12.3 and clang 14), "The C++ std surface is REQUESTED", and "Decision: the base goes".
Verified none of those three exist in `bd49aca4c` or its parent, so the branch could not
have been *moving* them; resolved to `main` in all four hunks. No duplicate headings in
the result.

**4. `cpp:LifecycleTransition` ledger row.** Both sides rewrote it. Kept `main`'s text and
appended the two facts only the branch's version carried, both verified in the tree: the
compile-test pin (`packages/api/nros-cpp/tests/compile/lifecycle_transition_ids.cpp`
exists) and the type-safe-overload rationale (`trigger_transition(LifecycleTransition)` at
`lifecycle.hpp:241` forwards to the `uint8_t` one).

**5. `rclcpp::shutdown()` doc.** Both sides describe the same fix (the discarded
`nros::shutdown()` `Result`); the code below was already identical. Kept `main`'s and
appended the branch's *bound* paragraph, which `main` lacks.

**6. `c:client_init` (service.json) and `cpp:Node::*_parameter` (param.json).** `main`'s
rows are strictly newer and one of them explicitly **retracts** what the branch's version
asserts (`main`: "`component_node.hpp` IS part of the umbrella now"; branch: "…is still
not part of the umbrella"). Resolved to `main`.

**7. `docs(phase-428): six commit citations named nothing`.** Took the branch's side for
the two hunks that replace a hash with prose (that is the commit's purpose) and `main`'s
for the third, an unrelated table-header rewording ("the pre-sweep census").

### One follow-up commit: four gates the rebase turned red

All four are green on `origin/main` and were green pre-rebase — each is a collision the
rebase created.

* **Five duplicate ledger keys** (`check-api-parity-ledger`, `check-api-parity`). `main`
  and this branch's phase-428 sweep each authored a row for `cpp:shutdown_transition_for`,
  `c:node_init`, `rust:throttle_decide`, `c:publisher_init`, `c:subscription_init`, at
  different file positions — so git merged both copies and JSON kept only the last, which
  is precisely the silent verdict-discard the gate exists to catch. Merged into one row
  each, `main`'s verdict/disposition, **both** reasons.
* `c:publisher_init` needed a correction rather than a concatenation. The sweep's reason
  *measures* the hole it reports — "`grep -n resolve_entity_name
  packages/api/nros-c/src/publisher.rs` returns ZERO" — and that no longer holds: this
  branch's own "eager-created C entities resolve their names" commit put the seam on all
  three publisher entry points. Measured: **10** occurrences on this branch, **0** on
  `origin/main`. The finding is kept for how the hole was found; the false sentence is
  marked corrected.
* **Orphan ref** (`check-ledger-orphan-refs`). This branch *widens* that gate to bare
  filenames, and `main`'s three `cpp:Node::{declare,get,has}_parameter` rows cite
  `rclcpp_compat.hpp` — deleted at phase-417 stage 6 step A — with none of the gate's three
  allow-list words (`deleted`/`removed`/`retired`) in the sentence. Restored the
  pre-rebase branch wording.
* **Dead hash** (`check-doc-commit-citations`). The commit whose whole point is "cite #557
  by PR number, the hash stopped naming it" quoted `00e8f3087` inside the prose explaining
  why — and the gate reads prose. The sentence now describes the hash instead of spelling
  it. Same edit fixes the tense: PR #557 is **MERGED**, so "unsupported until it merges"
  was false.
* **Stale ratchet** (`check-prose-issue-refs`). `docs/issues/1099-*.md` is on `main`, so
  the two baselined 1099 citations resolve. Removed by hand, not with `--write-baseline`,
  which drops the file's classification comment.

### Gates

`just check fast` — **283 of 285 pass**. The two failures are environmental (a fresh
worktree has no submodules) and reproduce identically on a fresh `origin/main` worktree:

```
capability-conditionals   zenoh-pico/include/.../platform.h missing
xrce-vendored-versions    micro-xrce-dds-client / micro-cdr not checked out
```

`cargo +nightly fmt --check --all` — clean.
`cargo check --all-targets` — clean for `nros`, `nros-macros`, `nros-c`, `nros-rmw`,
`nros-rmw-cffi`, `nros-params`, and `nros-node --features std`. (`-p nros-cpp` fails with
`unwinding panics are not supported without std`; identical failure on a plain `main`
checkout, so pre-existing and not from this branch.)

No submodule pin moved and no lockfile changed.

jerry73204:
## Rebased onto `main` @ `e4b576489` — 42 → 42 commits, nothing dropped

Follow-up rebase after `main` advanced by 4 commits past the previous merge base (`6e7779b05`).

**No-loss verification** (subjects reachable from the pre-push remote tip `327ff2098` vs. the new
head `7276da3c3`, both measured against `origin/main`):

```
comm -23 <(git log --format=%s origin/main..327ff2098 | sort -u) \
         <(git log --format=%s origin/main..HEAD      | sort -u)   -> empty
comm -13 ...                                                       -> empty
```

Empty in **both** directions: no commit lost, none silently gained, and no patch-id skip.
Submodule pins are byte-identical to `main` (`git diff origin/main..HEAD` over every `.gitmodules`
path reports no `Subproject` line), so the forward-only rule is not in play.

### One conflict, and it was not the one the diff looked like

Only `.config/prose-issue-ref-baseline.txt` conflicted, in commit `bb67960f5`. The ledger shards
(`param.json` and four siblings) auto-merged — that is the interesting part, so it is measured
below rather than assumed.

#### `.config/prose-issue-ref-baseline.txt` — a ratchet, resolved row-wise, plus a scrambling to undo

The three stages disagreed for two independent reasons, so I merged them separately.

**Data rows.** Base had 9. `main` added one; this branch removed two:

| row | base | `main` | branch | resolved |
| `docs/issues/1231-…:1164` | – | **added** | – | **kept** (main's) |
| `docs/roadmap/phase-428-…:1099` | ✓ | ✓ | **dropped** | **dropped** |
| `just/check/lanes.just:1099` | ✓ | ✓ | **dropped** | **dropped** |
| other 7 | ✓ | ✓ | ✓ | kept |

The branch's drop is a genuine tightening, but **not for the reason its commit message gives.**
The message says `docs/issues/1099-*.md` "landed on `main`". It has not:

```
git ls-tree -r --name-only origin/main -- docs/issues | grep 1099   -> no match
git log --oneline --diff-filter=A -- 'docs/issues/1099-*'           -> 4294de884 (this branch)
```

The file is supplied by **this branch** (`4294de884`), so the two citations resolve *here* and a
ratchet that keeps them stops tightening — the conclusion holds, the stated evidence did not.
I corrected the note in the file to name the branch rather than `main`, and dropped the now-stale
`1099 IN FLIGHT` classification entry with its rows. `1164` is left unclassified, matching how the
three existing `:1110` rows are treated.

**Comments.** `main`'s `0a02601b5` regenerated this file with `--write-baseline`, which sorts
*every* line — comments included — so the classification block came back interleaved with the data
(`# A RATCHET…` at line 12, between two data rows). That is the exact hazard CLAUDE.md warns about.
The merge base had it readable; the branch had restored it. I kept the branch's readable ordering
and added a line to the `Regenerate:` note so the next person does not re-scramble it:

> — but by HAND for anything but a row: the regenerator sorts EVERY line, comments included, which
> scrambles the classification below into the data.

Gate agrees with the resolution: `check-prose-issue-refs: OK (8 known in-flight, no new dangling
reference)` — 8 rows, exactly the union above.

#### Ledger shards — auto-merged, and verified rather than trusted

`git` merged `param.json` textually while both sides edited it, which is the shape that silently
produces duplicate keys (JSON keeps only the last). Checked directly with an
`object_pairs_hook` counter across all 17 shards: **0 duplicate keys, 0 parse failures.**

Then the row-wise question — did the branch's sweep clobber what `main` had just landed
(`1505290ec`, phase-426 W5 node-keyed C parameter surface)? Seven keys changed on `main` since the
merge base; all seven compare **equal to `main`'s value** in the merged file:

```
c:executor_allow_undeclared_parameters      PRESERVED
c:executor_allow_undeclared_parameters_on   PRESERVED
c:executor_declare_param_*                  PRESERVED
c:executor_declare_param_*_on               PRESERVED
c:executor_get_param_*_on                   PRESERVED
c:executor_has_param_on                     PRESERVED
c:executor_set_param_*_on                   PRESERVED
```

153 keys on `main`, 159 after the merge, 0 lost, 6 added by the branch. No hand-edit was needed.

The other two shapes flagged for this branch family did **not** occur: `result.hpp` did not
conflict, and `spin.rs` needed no arity fix — `cargo check -p nros-node --all-targets --features
std` is clean, so the `NodeKey`-first signature was already carried.

### Gates

| gate | result |
| --- | --- |
| `just check fast` | **283 / 285** — the 2 below |
| `just check api-parity-ledger` | 0 checks failed (+ self-test) |
| `just check ledger-orphan-refs` | OK — 17 shards, every path resolves |
| `just check issue-ids` | OK — 1191 issues unique |
| `just check prose-issue-refs` | OK — 8 known in-flight |
| `scripts/check-doc-commit-citations.py` | OK — 203 citations resolve |
| `cargo check -p nros-node --all-targets --features std` | exit 0, 0 errors, 0 warnings |

The two `check fast` failures are `capability-conditionals` and `xrce-vendored-versions`, both
missing-checkout complaints (zenoh-pico headers; the XRCE vendored trees). Not taken on trust —
reproduced on a **plain detached `origin/main` worktree** with byte-identical messages:

```
capability-conditionals   exit 2  "…platform.h is missing — run `git submodule update --init`…"
xrce-vendored-versions    exit 1  "no vendored tree checked out — nothing verified…"
```

Environmental to a fresh worktree, present on `main`, unrelated to this branch.

State: `mergeable: MERGEABLE`, auto-merge armed (`autoMergeRequest` non-null; the queue sets the
actual strategy), waiting on the required `CI` context.

jerry73204:

`7276da3c3` → `8d5107b3e`. **42 commits before, 43 after** (one cherry-picked from #704 — see below). No-loss verified:

```
comm -23 <(git log --format=%s origin/main..7276da3c3|sort -u) \
# (empty)
```

Nothing dropped, nothing skipped by patch-id.

**1. `packages/api/nros-cpp/include/nros/nros.hpp` — the shim `class Node` this branch edited no longer exists.**

`2e6528e90` ("a failed rclcpp create aborts") edited seven `create_*` verbs inside the `rclcpp::Node` shim class. main's phase-427 node merge deleted that class: one unconditional `class Node` in `node.hpp`, `using Node = ::nros::Node`, and the verbs defined out-of-line in `nros.hpp`. The **declarations** are superseded; the **payload** is not. Every one of the seven was verified present on main by name before its shim copy was dropped, and `detail::require_created` was ported onto main's out-of-line shape — `create_publisher`, `create_subscription`, `create_wall_timer`, `create_service` ×2, `create_client` ×2, each `(void)…` becoming `::rclcpp::detail::require_created(…, "<verb>", <name>)`. The branch's own runtime probe still proves it: `just check cpp` reports `failed_create_aborts: OK` (it forks and checks the child dies on SIGABRT with the migration text).

**2. An eighth create verb arrived on main, and the widened gate reached a row — cherry-picked from #704 rather than re-derived.**

`--require-disposition` went red on `cpp:create_timer` after the rebase, and it is a genuine new interaction rather than a latent red: `api-parity.py --check --require-disposition` exits 0 on plain `origin/main` **and** on the pre-rebase tip `7276da3c3`; only the combination fails. main's phase-430 W6/W7 added `rclcpp::create_timer(NodeT&&, ::nros::Clock*, period, CallbackT&&)` returning `shared_ptr<::nros::Timer>`, which upstream's four-required-parameter `create_timer` correlates `same` against once this branch's return-type-comparing correlator sees it.

#704 had already resolved exactly this. Rather than write a second answer to the same question, its commit was **cherry-picked verbatim** (`63e9345f3`), and the result verified byte-identical:

- the `create_timer` function body — `diff` against #704's tree: identical;
- the `cpp:create_timer` ledger row's `why` — `diff` on the decoded JSON string: identical.

Payload: `disposition: adopt-bounded` with the envelope named (no trailing `CallbackGroup::SharedPtr`, `::nros::Clock*` not `Clock::SharedPtr`, millisecond truncation, `shared_ptr<nros::Timer>` not `TimerBase::SharedPtr`), and the discarded `Result` on that eighth verb routed through `require_created` with the other seven.

**3. `.config/prose-issue-ref-baseline.txt` — main repaired the header independently, so this branch's repair is superseded and only its removal survives.**

`069aa3185` did two things to this file: it repaired the scrambled CLASSIFIED header (adding a "regenerate by HAND" warning, because `--write-baseline` sorted comments into the data) and it removed the two `…:1099` rows. main's `fix(#1241)` repaired the same header first and added `check_baseline_shape`, and per main's own account `--write-baseline` no longer destroys the header — so this branch's warning text is now **false** and its repair is redundant.

Resolution takes main's repaired file and applies only the removal: the two `…:1099` rows (`docs/roadmap/phase-428-…md`, `just/check/lanes.just`) and the 1099 IN-FLIGHT note, since this branch files `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md`. main's 9999 "TWO documents cite it now" paragraph is kept. `--write-baseline` was not used.

The result is **byte-identical** to the resolution taken on #629, #704 and #773, all of which carry a version of this commit — so whichever lands first, the rest drop it by patch-id instead of conflicting.

#### Ledger

The 5 shards `069aa3185` touches all merged cleanly. All 17 nevertheless scanned for duplicate keys with a `json.load(object_pairs_hook=…)` counter, since git's line-based merge creates them silently and JSON keeps only the last: **0 duplicates**.

#### Gates

| gate | result |
|---|---|
| `python3 scripts/check-doc-commit-citations.py` | OK |
| `just check cpp` | OK |

The one `check fast` failure is `xrce-vendored-versions`, environmental rather than pre-existing on this branch: reproduced identically on a plain `origin/main` worktree, because a fresh worktree has no `nros setup --source micro-cdr --source micro-xrce-dds-client`.

jerry73204:
Rebased onto `main` @ `cf4b0048a` (phase-438 W1). **43 commits before, 43 after** — no drops:

```
comm -23 <(git log --format=%s origin/main..8d5107b3e|sort -u) \
         <(git log --format=%s origin/main..HEAD|sort -u)   # empty
```

Two files needed non-mechanical resolution.

### 1. `docs/reference/api-parity-ledger/param.json` — ROW-WISE, then FIELD-WISE

The ledger is a flat dict keyed by symbol, so two branches adding different rows are not in
conflict even where git's line merge says the text is. Each conflict was re-merged per row, then
per field (`verdict` / `disposition` / `why` are independent), with both sides' changes verified
present afterwards.

One genuine field collision, on `why` for `cpp:Node::declare_parameter`,
`cpp:Node::get_parameter` and `cpp:Node::has_parameter`. Measured character-for-character rather
than judged by eye:

* `ours` (main) = `base` + a **1124**-char appended paragraph — the phase-426 W4 retraction: the
  node-local C++ store is gone, both facades forward through `nros/node_parameters.hpp`.
* `theirs` (this branch) = `base` + a **50**-char insertion in the middle,
  `, which phase-417 stage 6 step A has since deleted`, marking the `rclcpp_compat.hpp`
  reference as dead.

The two edited spans are **disjoint**, and that is asserted rather than assumed — `ours` must
start with `base`, and `theirs` must still end with `base`'s tail, so the branch's insertion
provably cannot reach the region `main` appended to. The merge is therefore
`theirs + ours[len(base):]`: **both sides kept in full**, nothing paraphrased or dropped.

Worth noting this branch's spelling of the repair carries **no line numbers**, so unlike its
siblings it has no staleness surface after phase-427 moved `class Node`.

### 2. `.config/prose-issue-ref-baseline.txt` — row-wise, respecting `fix(#1241)`'s shape rule

The header auto-merged correctly. Only the data rows collided: `main` **added**
`docs/roadmap/phase-431-ship-the-nros-binary.md:1110`, this branch **removes** both `1099` rows.
Row-wise result is both — 10 rows → **8**, and a ratchet may only shrink ✅. Verified against
`check_baseline_shape`: zero comments after the first data row, rows sorted. Edited by hand with
the header above all rows.

### The duplicate-key trap

Git's line merge silently creates duplicate JSON keys when both sides add the same key at
different offsets, and `json.load` keeps only the last — a discarded verdict no gate would catch.
Every resolution was followed by an `object_pairs_hook` duplicate scan over all 17 shards. It
fired mid-rebase on `node.json: c:node_init`, repaired later in this same series by
`fix(rebase): four gates the rebase turned red — five duplicate ledger keys, an orphan ref, a
dead hash, a stale ratchet`. **Final tree: zero duplicates across all 17 shards.**

### What `main` could have superseded, and did not

* phase-438 W1 moved the `NROS_CPP_HAS_*` capability predicates into `nros/std_detect.hpp`. This
  branch adds no inline predicate block, so nothing here is superseded by it.
* phase-427's node merge (no `class Node` in `nros.hpp`) leaves this branch's `require_created`
  payload intact: 9 occurrences in `nros.hpp` = 1 definition + 8 call sites, against **0** on
  `origin/main`. `just check cpp`'s `failed_create_aborts.cpp` probe compiles and passes, so
  that is verified by the compiler, not by grep.

### Gates

| gate | result |
|---|---|
| `just check fast` | **287 / 288** |
| `check xrce-vendored-versions` | FAIL — **environmental**, reproduced identically on a plain `origin/main` worktree (`no vendored tree checked out`; needs `nros setup --source micro-cdr --source micro-xrce-dds-client`) |
| `cargo +nightly fmt --check --all` | PASS |
| `just check api-parity` | PASS |
| `just check api-parity-ledger` | PASS |
| `just check ledger-orphan-refs` | PASS |
| `just check doc-refs` | PASS |
| `just check issue-ids` | PASS |
| `just check prose-issue-refs` | PASS |
| `python3 scripts/check-doc-commit-citations.py` | PASS |
| `cargo check -p nros-node --all-targets --features std` | PASS |
| `just check cpp` | every C++ compile probe PASS; fails only on `clippy::missing_safety_doc` for `nros_cpp_declare_remap` at `packages/api/nros-cpp/src/lib.rs:3262` — **pre-existing on `main`**, measured not assumed: a plain `origin/main` worktree fails the same gate with the same lint at the same line, and this branch's diff for that file is empty |

jerry73204:
Correction to the base named in my previous comment: `main` advanced while this batch was in
flight, and this branch's rebase actually landed on **`3fbb0bc66`** (`fix(#1249)`), not
`cf4b0048a` — 25 commits further along, including the whole phase-437 board-rename series. Its
siblings #629 and #704 are on `cf4b0048a`.

That is a strictly better base, and everything in the previous comment still holds — the
43-before / 43-after no-loss check, the resolutions, and every gate result were all measured on
this tree. Re-verified after the fact:

```
$ git merge-base origin/main HEAD
3fbb0bc66a51548fdb44807fa684816595d339df
$ git rev-list --count origin/main..HEAD
43
```

Also checked that none of the phase-437 renames collide with this branch: restricted to this
branch's own changes (`git diff --name-only $(git merge-base origin/main HEAD)..HEAD`), it
touches **no** path under `examples/qemu-*`, `examples/mps2-*`, `examples/rv-virt-*`,
`examples/esp32-c3-*`, nor `examples/fixtures.toml` or `nros-sdk-index.toml`. Same for #629 and
#704. GitHub reports all three `MERGEABLE` against the current `main`.

jerry73204:
## Rebased onto `main` (b5097a460) — 43 commits before, 43 after

`origin/phase-427-w10-executor-opens-node-named` was at `6a13be868`; it is now
`6b9f59cf3`. No-loss check over commit subjects vs `origin/main` is EMPTY — nothing
dropped.

### One conflict, the ledger row-wise class

`docs/reference/api-parity-ledger/log.json`: main added `rust:resolve_logger` (phase-427
W5), this branch's W5 pass added `c:log_severity_t`. Distinct keys, so nothing actually
disagrees — git only saw adjacent text. Resolved ROW-WISE from git stages `:1`/`:2`/`:3`
and field-wise within any shared row (no field-level conflict arose): base 72, main 73,
branch 73, merged **74** — one row from each side kept.

The remaining 42 commits replayed clean, including this branch's own
`5063b4c18 fix(rebase): four gates the rebase turned red — five duplicate ledger keys, an
orphan ref, a dead hash, a stale ratchet` and the tip's
`fix(rebase): the eighth create verb main added…`, which still applies against main's
current out-of-line `node.hpp`.

**Duplicate-key sweep after the rebase:** all 17 shards re-parsed with a
duplicate-raising `object_pairs_hook` — zero duplicates, zero conflict markers. Worth
stating explicitly because the silent-discard case lives in files git reports as cleanly
auto-merged, not in the one it flagged.

### Gates (measured in a clean worktree at the rebased tip)

| gate | result |
| --- | --- |
| `just check api-parity` | every divergence carries a ledger entry |
| `just check api-parity-ledger` | self-test 0 failed |
| `check-doc-commit-citations.py` | OK — 206 citations across 913 live files |
| `just check cpp` | All C++ checks passed |

`xrce-vendored-versions` is ENVIRONMENTAL: reproduced identically on a plain
`origin/main` worktree — a fresh worktree has no `nros setup --source micro-cdr --source
micro-xrce-dds-client` tree, so the gate verifies nothing and reports rc=1.

jerry73204:
Rebased onto `main` (`e802a796f`). BEFORE `6b9f59cf3` (43 commits) -> AFTER `bb6f15f62` (15 commits).

## The 28 dropped commits — every one verified on `main` by subject, individually

PR #629 (`phase-428 W5/W6/W10 + Q1/Q2`) merged in the meantime and carries most of this branch's phase-428 half. `git log origin/main --fixed-strings --grep=<subject>` for each dropped subject:

| dropped subject | on `main` as |
| --- | --- |
| design(RFC-0089, phase-427 W9/W10): Context and init settled | `c5f2c2cbf` |
| docs(#428): Context::default_from_env is a rename of nros::init() | `e2c7f823d` |
| docs(#428): the orphan-refs gate, widened to bare filenames | `acdc36697` |
| docs(issues): drop the open copies of 1087/1148/1149 | `8cb901965` |
| docs(phase-428): file five remainder findings as issues | `3971af8e2` |
| docs(phase-428 Q1): 130 refuse-loud rows relabelled absent | `78da2f7d8` |
| docs(phase-428 Q1): every live refusal site has a row citing it | `678ac9d71` |
| docs(phase-428 W5): 82 ledger rows | `86792f9ea` |
| docs(phase-430): measure the delta against phase-425 | `f447547d1` |
| docs(phase-430): rescoped | `d5435aaa1` |
| feat(phase-428 Q1): init_with_args refuses --ros-args loudly | `0b3decbcd` |
| feat(phase-428 Q1): --require-disposition rejects a refuse-loud that refuses nothing | `d5c191ba2` |
| feat(phase-428 W10): one QoS SSoT | `58612b5b4` |
| feat(phase-428 W6): a ledger row may not cite a file that does not exist | `05dbc1ae5` |
| feat(phase-428 W6): #[must_use] on Executor::declare_parameter | `8cb7888f1` |
| feat(phase-428 W6): #[must_use] where a bool reports failure | `469733748` |
| feat(phase-428 W6): --require-disposition reaches the silent class | `6d7d35fe3` |
| feat(phase-428 W6): the correlator compares return types | `36aee46d8` |
| fix(#428): check-ledger-orphan-refs matches a bare filename | `df45703f2` |
| fix(#629): the doc cited this branch's own pre-rebase tip | `7569d510a` |
| fix(phase-428): eager-created C entities resolve their names | `8b231b484` |
| fix(phase-428 W1-W4): the parity extractor sees our adopted names | `f01880443` |
| fix(phase-428 W6): the node-std-tests lane did not compile | `2cdebbbf5` |
| fix(phase-428 W6): the orphan gate runs its own control | `2cb453f2d` |
| fix(rebase): drop the unarchived duplicates of issues 1150-1152 | `06200c3d0` |
| fix(rebase): the eighth create verb main added | `eb3339dae` |
| ledger(phase-428 Q1 leftovers): measure the `env` surface | `9e3dc52f2` |

The 28th, `style(rebase): rustfmt the two call sites the conflict resolution rewrapped`, is the ONE with no counterpart on `main` — and it is subsumed rather than lost. It existed only to re-wrap two `assert!(server.declare(...))` call sites that a PREVIOUS rebase's conflict resolution had joined onto one line. `main` already carries the multi-line, trailing-comma form at both sites (`packages/core/nros-params/src/server.rs`, `packages/core/nros-node/src/parameter_services.rs`), so this rebase took `main`'s formatting at both conflicts and the commit had nothing left to do. Proof it is not a formatting regression: `cargo +nightly fmt --check --all` is clean on the result.

## Conflicts and how they were resolved

Nine commits conflicted. The ledger shards were merged with a **row-wise then field-wise 3-way merge** driven off git's index stages (1=base, 2=ours/main, 3=theirs/branch) rather than by picking a side of a line-diff: a key only one side touched takes that side, a key both touched is merged field by field, and only a field both sides rewrote is decided by hand.

**Every hand-decided row went to `main`'s text, and in each case for the same reason: `main` carries the CLOSED description of the finding this branch wrote OPEN.** #629 landed the closures. Examples, each read before deciding:

* `timer.json cpp:Node::create_wall_timer` — branch: "RANKED-9 also lands here ... discard the inner create `Result` with an explicit `(void)`". `main`: the same paragraph plus "phase-428 Q1 (2026-09-06) CLOSED it ... passes to `detail::require_created`", and it quotes the branch's own sentence as "History, one sentence". Strict superset.
* `lifecycle.json cpp:LifecycleNode::trigger_transition` — branch: `gap`, "Issue 1099, open". `main`: `divergence`, "LANDED 2026-09-06 (PR #602, issue 1099 archived), gap -> divergence".
* `init.json rust:init` — branch: `extension`. `main`: `rename` with a full `rename` object and blocker, which is this branch's own later re-verdict, already landed via #629.
* `pubsub.json cpp:Node::create_publisher`, `service.json cpp:Node::create_{client,service}`, `init.json rust:Context*`, the eight `*get_actual_qos` rows — same shape (the qos rows differ by literally four characters, `main` citing `rcl/rmw` where the branch cited `rcl`).

**Duplicate ledger keys — three found, all cleared.** Git's line merge silently creates duplicate JSON keys, and JSON keeps only the last. `git show :3:` on the branch's own committed blobs showed the class was already live in this branch's history (`c:publisher_init` was duplicated inside commit `797ad0f85`; `cpp:shutdown_transition_for` inside `5215db2ea`). During the replay, three duplicates appeared in the accumulating tree — `node.json c:node_init`, `pubsub.json c:subscription_init`, `other.json rust:throttle_decide`. The branch's own `fix(rebase): four gates the rebase turned red` commit is what clears them, and it did; `rust:throttle_decide` needed one hand decision, recorded below. **Final state: all 17 shards re-parsed with an `object_pairs_hook` counter — no duplicates, all 17 parse, and each round-trips byte-for-byte through `json.dumps(indent=2, ensure_ascii=False)`.**

**`other.json rust:throttle_decide` — the one non-mechanical merge.** Three texts existed: `main`'s 4068-char row, the branch's 1621-char row, and the dedup commit's own 3961-char merge of the two written against an OLDER `main`. Compared PARAGRAPH BY PARAGRAPH (`p in other`): all five substantive paragraphs of `main`'s row appear verbatim in the dedup commit's version and vice versa — the two differ only in their preamble, `main` saying "MERGED 2026-09-07 during the phase-428 W10 rebase" and the branch's saying "THE SECOND READING OF THIS KEY, merged at the 2026-09-09 rebase". The branch's 1621-char row is a single paragraph that appears verbatim inside `main`'s. So `main`'s row is the merge that LANDED and the branch's is what it already absorbed: the row is set to `main`'s, and the branch's re-description of a duplicate that no longer exists is dropped. Nothing textual is lost — checked, not assumed.

**Non-ledger conflicts**, all superseded-by-main and taken as `main`'s:
* `packages/api/nros-cpp/include/nros/nros.hpp` — two wordings of the same `nros::shutdown()` bound; `main`'s adds "one site over from the seven `create_*` verbs and the only one the compiler already pointed at".
* `packages/core/nros-node/src/executor/spin.rs` — `main` had dropped a redundant "Returns `true` if successful." from the same doc comment; `#[must_use]` itself is already on `main` (`git show origin/main:...spin.rs` confirms it on `declare_parameter`).
* `packages/core/nros-{params/src/server,node/src/parameter_services}.rs` — rustfmt line-wrapping, see the dropped-commit note above.
* `docs/roadmap/phase-428-api-porting-principle-sweep.md` — two hunks. One was the dead-hash paragraph, where `main`'s text names `add92bfcf` (PR #630) instead of the branch's dead `e82830bb6`/PR #612 attribution; taken as `main`'s, which is also what `check-doc-commit-citations` wants. The other was pure line-wrap, taken as the branch's since re-wrapping was that commit's whole point.

## Gates

| gate | result |
| --- | --- |
| `just check fast` (290 gates) | **1 FAIL: `xrce-vendored-versions`** — environmental, reproduced identically on a control worktree at plain `origin/main` (`e802a796f`) |
| `just check issue-ids` | OK (1219 issues) |
| `python3 scripts/check-doc-commit-citations.py` | OK — 207 citations across 914 live files resolve |
| `cargo check -p nros --features std,env --all-targets` | clean |
| 17-shard duplicate-key scan | no duplicates |

PR #755, opened 2026-09-07
phase-427 W1-W3/W5/W6/W8 + phase-430 W6/W7: one node type, named rclcpp::Node

jerry73204:
Implements **phase-427 W1, W2, W3, W5, W6, W8** and **phase-430 W6, W7**.
**W4 (`ComponentNode` deletion) is NOT started** and W7's deprecation is blocked
— both are recorded with the measurements behind them, in the phase doc.

## The number W1 asked for

`sizeof`, hosted `g++ -std=c++17`, via `check-cpp-capability-layout`'s own probe:

| | `nros::Node` | `rclcpp::Node` |
| --- | --- | --- |
| `origin/main`, baseline | 192 | 3752 |
| `origin/main`, `-DNROS_CPP_STD=1` | 192 | 3752 |
| **this branch, baseline** | **200** | **200** |
| **this branch, `-DNROS_CPP_STD=1`** | **200** | **200** |

Two types became one — `std::is_same<rclcpp::Node, nros::Node>::value` is
asserted by a compile probe — and the acceptance criterion holds. The +8 is
`void* hosted_`, which is the entire cost of the hosted surface to a node that
never makes a hosted-shape call: the block is allocated lazily and a
freestanding image can never allocate it at all.

`~Node()` is byte-identical in every configuration. The hosted block's type is
hosted-only, so a `#if`-gated `delete` would have given two TUs of one image two
different inline destructors — the ODR half of the same defect the layout rule
exists to prevent. The block carries its own `destroy` function pointer instead.

## What each work item did

* **W1** — every hosted-only member behind one unconditional `void* hosted_`.
  `std::enable_shared_from_this` had to be **deleted** to get there: it is a
  hosted-only BASE with a `std::weak_ptr` member, i.e. 16 bytes of layout behind
  a capability probe. `shared_from_this()` survives as a method returning a
  pointer that aliases `this` with an empty owner — ledgered as a divergence,
  and the two example templates that call it still build.
* **W2** — `Node(const char*)` + `init()` + `ok()` freestanding; upstream's
  `std::string`/`NodeOptions` constructors hosted.
* **W3** — the out-ref and `shared_ptr` `create_*` families are overloads on one
  type with no ambiguity; `create_wall_timer` gains the member-binding template
  overload, which **retires `nros::bind_timer`** (all 22 in-tree call sites
  moved in the same commit, `NROS_BIND_TIMER` repoints to the member, the
  scaffold codegen and its test moved with them, and the free function stays one
  release as a deprecated forwarder whose diagnostic is gated).
* **W5** — `get_logger()` follows ROS 2. The merge FORCED this: two accessors
  with identical signatures and different behaviour (a real node logger vs the
  hardcoded `"nros.compat"` sentinel) cannot coexist on one type, and two
  overloads differing only in return type are ill-formed. `rclcpp::Logger` now
  carries the node's name **and** the opaque handle, with an implicit conversion
  to `nros_logger_t` so every native `NROS_LOG_INFO(node.get_logger(), …)` and
  the `logger == nullptr` check in `examples/native/cpp/logging` still compile.
* **W6 / W8** — the error-channel RULE is stated in `result.hpp` beside the
  types it governs, and `NROS_NODISCARD` is on both channels.
* **phase-430 W6** — the clock-taking `create_timer<C, &C::m>` member and the
  hosted free `rclcpp::create_timer(node, clock, period, cb)`, humble's only
  form, returning the same cell `create_wall_timer` does.

## The `TimerBase` ruling (phase-430 W7): DELETE

RFC-0089 decided the timer stays flat; the tree carried a `TimerBase` with a
virtual destructor and `detail::WallTimer` under it anyway, added by the same
campaign, with neither document noticing. The ruling is delete, argued from the
executor's dispatch:

1. **The vtable had no caller.** The only virtual member was `~TimerBase()`, and
   no virtual call through a `TimerBase*` existed anywhere — none can, because
   the executor's callback slot is `nros_cpp_timer_callback_t`, a raw
   `void(*)(void*)`, and dispatch goes through the STATIC
   `detail::WallTimer::trampoline`. Even the destructor's virtuality was dead:
   the cell is built with `std::make_shared<detail::WallTimer>()`, so the
   control block already records the concrete deleter.
2. **The name promises children we refuse to have.** `WallTimer` and
   `GenericTimer` stay absent by decision — the clock axis is a runtime field
   plus a second VERB, which is the shape phase-425 shipped.
3. **The flat shape is a better keep-alive.** `create_wall_timer` returns
   `std::shared_ptr<nros::Timer>` aliased onto the private cell, the shape
   `create_subscription` has always used.

**Cost, measured THREE times; the first two answers were wrong and both are
recorded in-branch rather than tidied away.**

1. *"Zero non-test call sites"* — a `grep --include` glob the shell expanded to
   nothing. Real: 3 source files + a book snippet. Acting on it, the name was
   deleted outright.
2. *"A one-release deprecated alias"* — better, still wrong: it treated the
   alias as a courtesy to in-tree consumers.
3. **`colcon-parity` then went red and named the real constraint:**

   ```
   rclcpp::Timer -> 'Timer' in namespace 'rclcpp' does not name a type;
                    did you mean 'Time'?          [REAL ROS 2, humble]
   ```

   That gate sources `/opt/ros/<distro>/setup.bash` and builds
   `examples/templates/local-msg-package` against **real rclcpp**. That template
   — and `cpp-port-minimal-publisher`, vendored UNMODIFIED — exists to
   demonstrate upstream source compiling here, and both declare a timer member.
   Upstream has no `rclcpp::Timer`. So **without the alias the intersection of
   "compiles under real ROS 2" and "compiles under nano-ros" is EMPTY for any
   ported node holding a timer.**

**Settled ruling:** the HIERARCHY is deleted (W7's question, and where the cost
was); the NAME is KEPT as the ported alias and is **not** deprecated, because a
deprecation would promise a removal that breaks the campaign's flagship property
on purpose. The two ours-authored consumers and the book snippet are reverted to
`std::shared_ptr<rclcpp::TimerBase>`, each with a comment saying why.

The expected-failure probe is replaced by a positive one,
`timer_base_is_a_name_not_a_hierarchy.cpp`, holding the half that can still
regress: `is_same<TimerBase, Timer>`, `!is_polymorphic`, both nested `SharedPtr`
spellings, both member declarations the templates contain, and a self-tested
detector for the continued absence of `WallTimer`.

**The general lesson, now in RFC-0089 and bigger than the timer:** for a name
upstream HAS and we lack, the ported alias is not a courtesy — it is the only
thing that keeps a file compilable both ways. "Do not take the name" is
available only where upstream does not have it either, and `colcon-parity` is
the one gate where that two-directional requirement is measured. RFC-0089
§"Timer, studied against RTOS semantics" is corrected accordingly; its earlier
§"Review of the invented parts" item 1, which proposed exactly this alias, is
vindicated.

`rclcpp::Timer` is UNCONDITIONAL, so a freestanding target gets a ROS 2 timer
name too.

## Three things RFC-0089 got wrong, corrected in the RFC

1. **The namespace direction is inverted**, and the blocker is
   `scripts/api-parity.py`'s own namespace roots: it reads the native C++
   surface as `{"nros"}` and gates only that bucket, so defining the class in
   `rclcpp` re-buckets ~60 members to `theirs-only` and `just check api-parity`
   goes red. The class is defined in `nros::` and aliased into `rclcpp::` — the
   direction every other type in this API already uses. Flipping it is a change
   to the extractor's roots, which moves the whole ported surface at once:
   phase-428's sweep.
2. **`Result<void>` with `Result` as the alias is not expressible in C++.** In
   one namespace `Result` is either a class-template name — in which case the
   bare `Result` every call site writes is ill-formed — or a non-template type
   name, in which case `Result<T>` is. Every re-export spelling collides. What
   was the actual finding, the rule for choosing a channel, landed instead.
3. **RFC-0047's "one component, several named nodes" does not exist.**
   `ComponentNode` owns exactly one node, created once in its constructor, and
   `entity_inventory.rs` states the invariant in so many words. What the subnode
   packages exercise is one node with several named CALLBACK GROUPS — which the
   merged node already carries. W4 has no several-named-nodes capability to
   preserve.

Plus one nothing had noticed: **a base class is a member for layout purposes**,
which is why `enable_shared_from_this` had to go and why it was invisible while
the whole shim class sat inside a capability `#if`.

## W7 is blocked, and the item does not mean what it says

"`nros::Node` deprecated with `NROS_DEPRECATED_MSG`" attaches to the ALIAS, and
after the merge the alias is `rclcpp::Node` — the name a user is supposed to
write. The macro ships and is USED, on the one ours-only name this phase did
retire (`nros::bind_timer`). The `Node` half waits on the namespace flip, and
whoever lands it must migrate the 218 in-tree `nros::Node` sites in the same
commit: a bare attribute warns at all of them at once, which the phase's own
"Not in scope" forbids.

## W4 is not started, and what it must decide first

Attempting it here would have shipped a half-merged `ComponentNode`, which is
worse than either half. The phase doc carries the full inventory; two decisions
are **not** in the work item as written:

* **The `create_publisher` RETURN SHAPE collides, and the merge is what creates
  the collision.** `ComponentNode::create_publisher<M>(const char*, const QoS&)`
  returns `Publisher<M>` by VALUE; the merged node's
  `create_publisher<M>(const std::string&, const QoS&)` returns a `shared_ptr`.
  On one type, `create_publisher<M>("chatter", 10)` — a string LITERAL — binds
  the `const char*` overload and returns by value. That is compile-and-differ
  manufactured by the merge itself.
* **The error latch is 24 unconditional bytes and the timer pool is 192**, on
  every node including freestanding ones, and neither can live in the hosted
  block (the boot-halt mechanism is exactly what a freestanding image needs).

Also recorded: no subnode package builds freestanding today (all three
`fixtures.toml` rows are `platform = "linux"`), and four gates are keyed on
`component_node.hpp` by path or by namespace.

## New probes, and the negative one is the item

* `one_node_type.cpp` (hosted) — one type, both create families, member binding,
  `rclcpp::create_timer`, the logger.
* `one_node_type_freestanding.cpp` (`-nostdinc++`, ThreadX minimal libcpp) — a
  TU that CONSTRUCTS a node and CREATES A PUBLISHER, type-checked through the
  template bodies. A header-parse lane cannot see this: `-include` instantiates
  nothing.
* `ported_create_publisher_freestanding_probe.cpp` — **expected failure**. A
  ported `create_publisher<M>("chatter", 10)` must not compile where there is no
  allocator, and the lane GREPS the diagnostic for the out-ref overload's
  signature, because "it failed" is also what a typo produces.
* `timer_binding_paths.cpp` + `bind_timer_deprecation_probe.cpp` — the
  `bind_timer` retirement, both halves, instantiated (every path that reached it
  is a template body, which no header-parse lane builds).
* `timer_base_is_a_name_not_a_hierarchy.cpp` — `TimerBase` and `Timer` are one
  flat type; no vtable; `WallTimer`/`GenericTimer` still absent (self-tested
  detector).

## Verified locally

`just check cpp` (green through every probe including the three new ones and the
`-nostdinc++` lane), `just check api-parity` (green — 5 new rows in
`node.json`), `check-cpp-capability-layout`, `check-cbindgen-headers`,
`check-ffi-struct-mirrors`, `check-api-parity-ledger`,
`check-cpp-freestanding-includes`, `check-cxx-standard-floor`,
`check-grep-q-error-conflation`, `check-book-identifiers`, `check-doc-refs`,
`cargo test -p cargo-nano-ros`.

`just colcon-parity` — the gate that caught the `TimerBase` deletion — is green
locally (3 packages, consumer ELF produced).

`just check fast` has 4 reds on this host, all pre-existing and environmental:
the zenoh-pico and XRCE vendored trees are not checked out
(`capability-conditionals`, `xrce-vendored-versions`), a
`docs/issues/README.md` commit citation, and a sandbox `PermissionError` inside
`check-xrce-source-manifest`'s own self-test. `just check cpp`'s final step
(`nros-cpp clippy`) and `check-compile-smoke` fail for the same unprovisioned
submodules.


jerry73204:
Note for whoever lands second: this overlaps **#724** (phase-426 W4, "delete both C++ parameter stores") on `packages/api/nros-cpp/include/nros/nros.hpp`, `component_node.hpp`, `just/check/lanes.just` and `docs/reference/api-parity-ledger/node.json`, so a queue rebase will conflict in at least `nros.hpp`.

The conflict is shallow and the resolution is mechanical in one direction: this PR moved the shim `rclcpp::Node`'s parameter facade onto the merged `nros::Node` with `params_` living in `detail::NodeHosted` (`node.hpp`), so #724's deletion applies to `detail::NodeHosted::params` and to the out-of-line `Node::declare_parameter` / `get_parameter` / `set_parameter` / `has_parameter` / `parameters` bodies in `nros.hpp`, plus their declarations in `node.hpp` under `NROS_CPP_NODE_HOSTED`. Nothing else in the merge depends on the store.

phase-427 W4 (deleting `ComponentNode`) is **not** in this PR and is not started; the phase doc records what it has to decide first.

jerry73204:
Sweep finding while landing this: `check-cpp-capability-layout` measures its rule on an authored three-name `TYPES` list, so every other public C++ type has been unmeasured since the gate was written. Asking the same question of the rest of the umbrella found **three types whose `sizeof` follows `NROS_CPP_STD`** — `nros::Timer` 24→32, `nros::GuardCondition` 32→40, `nros::ComponentNode` 55784→55848 (it holds `Timer timers_[8]`; 8×8 = 64 is exactly its delta). One member, twice: the `#ifdef NROS_CPP_STD` `std::unique_ptr<std::function<void()>> closure_`.

Same class as issues 0135/0460 and as the `rclcpp::Node::timers_` scar phase-417 paid for — and a supported state, since px4's bridge sets `-DNROS_CPP_STD` on one module of a larger image on purpose.

**Filed as issue 1225, not fixed here.** The mechanical repair is this PR's own `hosted_` trick and it costs a freestanding target +8 bytes per timer, +8 per guard condition and +64 per `ComponentNode`; the cheaper alternative (caller-owned closures, the shape `rclcpp::detail::WallTimer` already uses) is a design question phase-427 did not ask. `nros::Node` is 200/200, and `QoS`, `Executor`, `Clock`, `Time`, `Duration`, `CallbackGroup`, `NodeBuilder`, `LifecycleNode` and `rclcpp::NodeOptions` are clean.

jerry73204:
Correction on the `TimerBase` ruling — **`colcon-parity` caught what my measurement did not**, and the fix is in `a3c81ef`.

Deleting the name turned that gate red with `'Timer' in namespace 'rclcpp' does not name a type; did you mean 'Time'?` — from **real rclcpp**. `just colcon-parity` sources `/opt/ros/<distro>/setup.bash` and builds `examples/templates/local-msg-package` against upstream ROS 2; that template and `cpp-port-minimal-publisher` exist to prove upstream source compiles here, and both declare a timer member. Upstream has no `rclcpp::Timer`, so without the alias the intersection of "compiles under real ROS 2" and "compiles under nano-ros" is empty for any ported node holding a timer.

Settled: the **hierarchy** is deleted (W7's question, and where the vtable cost was); the **name** is kept as the ported alias and is *not* deprecated. The expected-failure probe is replaced by a positive one pinning `is_same<TimerBase, Timer>`, `!is_polymorphic`, and the continued absence of `WallTimer`/`GenericTimer`.

The general lesson is now in RFC-0089 and is bigger than the timer: **for a name upstream HAS and we lack, the ported alias is not a courtesy — it is the only thing that keeps a file compilable both ways.** "Do not take the name" is available only where upstream does not have it either, and `colcon-parity` is the one gate where that two-directional requirement is measured. Worth checking any future decision to withhold an upstream name against it before landing.

`just colcon-parity` is now green locally (3 packages, consumer ELF produced).

jerry73204:
Rebased onto `main`, and the W8 collision is resolved by **adopting main's design, not re-landing this branch's.**

## What collided

Both sides did phase-427 W8, reached the same language finding, and stopped in different places. RFC-0089 asked for one template named `Result` with the value-less case as `Result<void>`; C++ cannot express that — an identifier in a scope is a class, a class template, or an alias, never two.

* **This branch** concluded the settled shape was unreachable and kept two types: `Result` (class) and `Expected<T>` (template). `NROS_NODISCARD` deliberately empty on C++14.
* **`main`** ([landed 2026-09-07](https://github.com/NEWSLabNTU/nano-ros/commits/main/packages/api/nros-cpp/include/nros/result.hpp)) got the whole structure with only the template's own spelling changed: one template **`ResultOf<T>`**, `Result` an alias for its `void` specialization, `Expected<T>` kept one release as a `[[deprecated]]` **class** template (measured: `[[deprecated]]` on an alias template is silent on clang 14). Its `NROS_NODISCARD` is also stronger — the GNU spelling on clang below C++17, where clang honours it on types and gcc does not.

Main's is strictly the better answer to the same question, so this branch takes it.

## What that cost, concretely

The branch's first commit (`cdff5c55c`, W6/W8) touched **only** `result.hpp` and is now entirely superseded — **dropped**. Exactly one thing in it `main` lacks: `NROS_CPP_DEPRECATED_MSG`, which W7 needs to deprecate `nros::bind_timer`. It is re-landed as a small commit at the base of the branch, with the reason it lives in `result.hpp` and the note that it needs none of `NROS_NODISCARD`'s carve-outs (`deprecated` is honoured on classes and functions in both spellings by gcc and clang).

Every `Expected<T>` in the branch's headers auto-merged to main's `ResultOf<T>` — verified zero remain outside main's own deprecation probe.

Three documents asserted the superseded conclusion and are corrected rather than left:

* `docs/roadmap/phase-427-one-node-type.md` — the section headed "W8's settled shape is **not** expressible in C++" ended `Expected<T> keeps its name`, and claimed `NROS_NODISCARD` is empty on C++14. Both false now. Rewritten to say which half is inexpressible and point at main's own "W8 LANDED" section below it.
* `docs/design/0089-…` §3 — same claim, same fix. Main already corrected §"The error channel, settled" in place, so §3 keeps only what that section does **not** carry: the acceptance named the wrong probe (`rclcpp::init(argc, argv)` returns `void` here by the ported-channel rule, so a TU discarding it compiles clean and the gate would have been vacuous — `nros::init()` is the widening).
* `changelog.d/755.feat.md` — named the retired spelling as the current one.

## The gate that was armed for this

`check-cpp-capability-layout` listed `rclcpp::Node` hosted-only with the note *"phase-427 … takes this entry off the list, and the check below fails the moment the type becomes measurable so the removal cannot be forgotten."* It fired exactly as designed:

```
FAIL: rclcpp::Node is listed hosted-only but DOES measure freestanding (200)
  Remove it from hosted_only_reason(); its exemption is stale.
```

`hosted_only_reason()` is now empty — folded into the node-merge commit, since that is the change that made the type measurable. The function stays, so a future hosted-only type has a place to declare itself and both ratchet arms keep meaning.

## Verified

`just check cpp` — **all C++ checks passed**, including main's `phase-427 W8: the Expected<T> spelling is deprecated (expected failure)` probe and the capability-layout selftest (3 cases). `cargo +nightly fmt --check --all` clean, `clang-format` clean. `just check fast` 282/283 — the one red is `xrce-vendored-versions`, which refuses because a submodule is not checked out in this worktree and no C++ edit can reach it. `doc-refs`, `issue-ids`, `api-parity-ledger`, `doc-recipe-refs` all OK.

13 commits before, 14 after (the dropped W8 commit, plus two: the deprecation macro and the changelog correction).

PR #768, opened 2026-09-08
fix(phase-412 #7): one fact, three roads, three registries that each described a different subset

jerry73204:
A CMake-derived number reaches a Rust build script by one of three roads — the Zephyr resolver,
the cargo-leaf sidecar, the declared road — and each was checked by a registry naming a different
subset. Now every published `NROS_DERIVED_*` fact has a **stated disposition on every road**:
carried by these names, or not carried for a quoted reason, or recorded as an unexplained gap.

## What changed

**`check-knob-delivery.py`** — `DERIVED_PAIRS` values became tuples, so one fact can name several
resolved knobs, and it gained the two facts no registry named. `resolver_road()` harvests the
`_nros_resolve_derivable_knob` call sites out of `nros_cargo_build.cmake` with an if/endif stack
that keeps the enclosing `if(CONFIG_NROS_RMW_*)` guard, and the map must match that harvest in
both directions.

**`check-declared-fact-carriers.py`** — new `FACT_DISPOSITION`. `NotCarried` must **quote a comment
that already exists**, and the gate re-reads the cited file to confirm it is still there.
`ROAD_PAIRS` is now derived from it rather than authored beside it.

## Disagreements: resolved vs recorded open

| item | verdict |
|---|---|
| `LARGEST_TYPE` / `LARGEST_RX` / `LARGE_TYPES` carried by no road | **Resolved as provenance**, quoting `NanoRosMessageBounds.cmake:158-159`. Two reach the fragment only as *comment* lines; `LARGEST_RX` not at all. |
| `EXECUTOR_ACTION_CLIENTS` in no registry | Confirmed, now in both. |
| `DERIVED_PAIRS` 1:1 | Confirmed. 13 call sites over 11 facts — `MAX_SUBSCRIBERS` and `MAX_QUERYABLES` each feed a zenoh **and** an XRCE knob, so a value arriving at one and dropped at the other used to be green. |
| `EXECUTOR_MAX_NODES`, `SUBSCRIPTION_BUFFER_SIZE` resolver-only | **Recorded as OPEN** (issue 1233). No comment, issue or constant justifies them. Nothing invented. |

The four open roads are printed on every successful run rather than hidden, so the gate is green
*and* the gaps stay visible.

## Also found

- **`SUBSCRIPTION_BUFFER_SIZE` was in no registry either** — not just `EXECUTOR_ACTION_CLIENTS`.
  Note it is *not* `SUBSCRIBER_BUFFER_SIZE`, which is: two names one edit apart, two derivations.
- **A latent false positive, fixed:** the old rule demanded every mapped knob be present in the
  cache, and `NROS_SUBSCRIBER_BUFFER_SIZE` has sat inside `if(CONFIG_NROS_RMW_ZENOH)` since 1125 —
  so the gate would have reported a dropped value on any XRCE build dir. The guard harvest makes a
  backend-conditional absence clean and an *unconditional* one a failure.
- `ROAD_UNPAIRED`'s three `NROS_DECLARED_QOS_*` entries are inert: they are cmake global
  properties, never on the `corrosion_set_env_vars` road, so `produced()` never sees them.

## Verification

Five mutations, each restored to green — including renaming the fact at the XRCE call site (fails
in both directions) and deleting a cited provenance comment (`the comment it cites is GONE`).

I re-verified in a clean checkout rather than trusting the agent worktree's numbers:

```
just check fast                      281 gate(s) ran, 2 ledger skips, 0 failed
just check knob-delivery             DERIVED_PAIRS names every one of the 13 call sites over 11 facts
just check declared-fact-carriers    14 facts, each with a disposition on all 3 roads; 4 OPEN printed
```

Independent mutation by me: removing `NROS_EXECUTOR_ACTION_CLIENTS` from `DERIVED_ENV_KEYS` gives
`FACT_DISPOSITION says the sidecar road carries NROS_EXECUTOR_ACTION_CLIENTS and it does not.`

The worktree run reported three failures (`capability-conditionals`, `cpp-fmt`,
`xrce-vendored-versions`); all three are **absent in a clean tree** — they are uninitialised
submodule / unsynced artifacts, which is issue 1234.

One self-test bug worth repeating: the "quoted comment is gone" case first spelled the missing
sentence as a literal, which *put it in the file* and made the case pass for the wrong reason. It
concatenates at runtime now, with a comment saying so.


jerry73204:
Heads-up from the phase-426 landing session: this pull request is ejecting from the merge queue, and because a merge-queue batch entry contains every entry ahead of it, the entries behind it inherit the same red — `pr-755-…`, `pr-771-…` and `pr-548-…` all failed with the identical message. `just queue-triage` reads that as "the same check failed for several unrelated pull requests, so it is not yours"; here it is one entry at the head of the batch rather than shared infrastructure, so the heuristic points away from the cause.

The failure is this pull request's own gate, in `just check fast`:

```
check-declared-fact-carriers: 2 problem(s)
  NROS_DECLARED_MAX_QOS_DEPTH travels the declared road and FACT_DISPOSITION attributes it to
       no fact.
  NROS_DECLARED_NODES travels the declared road and FACT_DISPOSITION attributes it to
       no fact.
```

Neither fact is missing a road — both are produced, consumed and watched. What they lack is a `FACT_DISPOSITION` entry, and `FACT_DISPOSITION` is the authored registry this pull request introduces. Both facts landed on `main` after this branch was written:

- `NROS_DECLARED_NODES` — phase-426 W3, merged as `15b8daa4d` / `94bfdf1fc`. The parameter services are registered once per node, so the image's node count became a term in the zenoh queryable pool.
- `NROS_DECLARED_MAX_QOS_DEPTH` — phase-412 W3, #738, merged earlier in the same queue run.

On current `main` the gate is green — it reports twelve declared facts "produced, consumed and watched", `NROS_DECLARED_NODES` and `NROS_DECLARED_MAX_QOS_DEPTH` among them — so this is not a red on `main` that this branch inherits. It is the AUTHORED half of the registry being narrower than the tree it now describes, which is the same drift the change exists to prevent, arriving through the queue rather than through a commit.

The fix looks like a rebase onto current `main` plus two `FACT_DISPOSITION` rows. It is worth noting that an authored registry over an enumerable set has exactly this failure mode whenever a fact lands between the branch being written and the branch being queued, and this pull request's own reasoning ("three registries that each described a different subset") is the argument for the gate catching it — which it did.

Re-queuing before those two rows exist burns a batch slot and ejects everything behind it again.

PR #770, opened 2026-09-08
feat(phase-438 W1+W2, #1187): one detection site, then the std surface is REQUESTED — every embedded C++ FreeRTOS image compiles again

jerry73204:
Stacked on #763 (which is stacked on #746). **Retarget as those land.**

Two commits. W1 is behaviour-neutral and makes W2 small; W2 closes **issue 1187**.

## Acceptance — issue 1187's own, as it wrote it

```
$ NROS_CMAKE_EXTRA_DEFS=-DCMAKE_TOOLCHAIN_FILE=$PWD/cmake/toolchain/arm-freertos-armcm3.cmake \
  bash scripts/build/fixtures-build.sh freertos cpp zenoh
EXIT=0
requires_hosted hits: 0
```

Six ARM ELF binaries where none built before: `cpp_talker`, `cpp_listener`, `cpp_service_server`, `cpp_service_client`, `cpp_action_server`, `cpp_action_client`.

Per-header standalone parse on that toolchain (`arm-none-eabi 13.2`, `-std=c++14 -ffreestanding -fno-exceptions -fno-rtti`):

| | pass | fail |
|---|---|---|
| before | 26 | 20 |
| after | **47** | **0** |

## W1 — one detection site

Fourteen hand-copied instances of one preprocessor block, across ten headers, become `nros/std_detect.hpp`. Plus `<chrono>`'s block, which is **not** the same block and is moved rather than merged.

It cost three things, and only one was the obvious one:

1. **`nros.hpp` USED four capability macros and DEFINED none.** It depended on `publisher.hpp` / `options.hpp` / `subscription.hpp` being included first — the only header in the tree in that state, and a latent issue 0135 (two TUs of one image disagreeing about a capability, hence a layout). Consolidation fixes it as a side effect.
2. **The rationale drifted.** Eleven copies said the test is `__has_include` "rather than `__STDC_HOSTED__`, issue 0112" — and 0112 says no such thing. Rewritten, not moved.
3. **W2 was a fourteen-site sweep.** It is five lines in one file. The baseline is the count: 14 → 5 → 0, with no debt paid to get from 14 to 5.

**Behaviour-neutral, measured.** Hosted `pass=46 fail=0` → `47/0`; arm `26/20` → `26/21`. Both deltas are the new file appearing in the loop as its own header; the pre-existing failing set is byte-identical.

**`<chrono>` is not normalised.** No `#elif` arm, an `__STDC_HOSTED__` conjunct, and `<ratio>` as a *prerequisite* rather than a proxy — `duration_cast` is defined via `std::ratio_divide` and the safety island ships `<chrono>` with `<ratio>` absent. Its 79-line comment moves verbatim: the comment is the evidence, not decoration.

## W2 — delete discovery

| | `__STDC_HOSTED__` | `__has_include(<string>)` | `#include` |
|---|---|---|---|
| arm-none-eabi 13.2 `-ffreestanding` | 0 | **TRUE** | hard `#error` |
| Zephyr `-nostdinc++` minimal libcpp | 1 | FALSE | absent |
| hosted g++ 12.3 | 1 | TRUE | works |

libstdc++ 13 added `bits/requires_hosted.h`, so presence stopped implying usability. Each probe is right on exactly one embedded lane — **including issue 1187's own second fix option**, ANDing the two, which fails on whichever lane it is not measured on. Only the request is right on both.

### Where the flag went — five sites

- **`cmake/compat/NrosRclcppCompat.cmake`** — the shim's own apply function, which until now set **no compile definition at all**. This is the one that reaches real ported code: five `compile_check_fixture` rows plus the `diagnostic_updater` shim. Per-target, so it does not leak to Zephyr images using `nros::Node` directly.
- **The phase-417 ported-surface probes** — they exist to prove a ported rclcpp file compiles, which *is* the surface the flag names.
- **The four REFUSAL probes**, and the lane's own guard is why: without the flag `ros2_refuse_node_options_probe.cpp` still failed, but on `'NodeOptions' is not a member of 'rclcpp'` rather than on the refusal — the exact "broke for another reason, so this check proves nothing" case the grep exists to catch. It caught it.
- **`compile-check-fixtures.sh`'s snippets arm**, as a **rule not a list**: `platform_hdr_*` prove `<nros/platform.h>` type-checks *freestanding* and must not get the flag; everything else reaches `<rclcpp/rclcpp.hpp>` and does.
- **`check-cpp-capability-layout.sh`'s `HOSTED_FLAGS`** — not a new configuration, the same one asked for out loud.

### One predicted cost did not materialise

phase-438 listed `api-parity.py`'s `base` and `component` TUs as needing the flag. **They do not** — green with no ledger movement, so the native records were never coming from the discovered macros. Their flaglessness is now a stronger statement: they measure the native surface as a consumer who never asked for the porting flavour sees it.

## Unblocks

Issue 1146 can now measure the FreeRTOS app-task stack on the C++ path, so `freertos_app_config.c.in`'s 512 KiB — one number serving C and C++, measured only on the C half — can come down to the measured figure.

## Tier

`just check fast` (282 gates), `just check cpp`, `just check api-parity` — all green, plus the cross build above.

### DIFF
--- a/.config/cpp-freestanding-includes-baseline.txt
+++ b/.config/cpp-freestanding-includes-baseline.txt
@@ -1,32 +1,22 @@
-# Hosted STL includes reached with `NROS_CPP_STD` undefined, in the `#elif
-# defined(__has_include)` arm of nros-cpp's capability blocks.
+# Hosted STL includes reached with `NROS_CPP_STD` undefined, in nros-cpp's
+# capability blocks.
 #
 # A RATCHET, not an allowlist. It may only SHRINK: a line here that no longer
 # matches a violation is a FAILURE, so the debt cannot go stale, and a new
 # violation not listed here is a failure too, so it cannot grow.
 #
-# EMPTY as of issue 1240 (2026-09-09). All fourteen entries are paid.
+# EMPTY as of issue 1240 (2026-09-09). All fourteen entries are paid, and
+# phase-438 W1+W2 then removed the shape that produced them.
 #
-# WHAT PAID THEM, AND WHY IT IS NOT WHAT WAS PLANNED
+# WHAT PAID THEM, IN TWO STEPS THAT DISAGREED ABOUT THE INSTRUMENT
 #
-# The plan recorded here was phase-438 W2: "delete the `#elif` arm, so the std
-# surface is reachable only through `NROS_CPP_STD`." That would have been the
-# wrong fix, and `nros.hpp` already says why in its `<chrono>` note — NOTHING
-# THAT SHIPS DEFINES `NROS_CPP_STD`. It appears in no cmake module, no
-# toolchain file, no build.rs and no Kconfig, only in docs and in `check.just`'s
-# own probe TUs. Deleting the arm would not have tightened the gate on
-# freestanding targets; it would have removed `std::string`-taking `get_logger`,
-# `std::shared_ptr` aliases and the `<vector>` options surface from EVERY
-# configuration, hosted ones included. `nros.hpp` made exactly that mistake for
-# `<chrono>` on 2026-09-05 and corrected it two days later.
-#
-# What paid them instead is the corrected predicate from that same note, applied
-# to all fourteen sites at once (issue 1240):
+# Issue 1240 paid all fourteen by CORRECTING the probe in place, at all fourteen
+# sites at once:
 #
 #   #if defined(NROS_CPP_STD) || (defined(__STDC_HOSTED__) && __STDC_HOSTED__ \
 #       && __has_include(<string>))
 #
-# Neither probe alone is sufficient and both are necessary:
+# Neither probe alone is sufficient and both were necessary:
 #
 #   * `__has_include` alone answers "does the header FILE exist", which is TRUE
 #     under `-ffreestanding` for a libstdc++ whose `<string>` begins
@@ -37,15 +27,38 @@
 #     hosted compiler run `-nostdinc++` against Zephyr's minimal libcpp reads
 #     HOSTED and still has no `<string>`.
 #
-# The conjunction covers both lanes; `NROS_CPP_STD` stays as the explicit
-# override for a consumer who knows better than the probes.
+# phase-438 W1 then moved that one predicate to its single definition site,
+# `packages/api/nros-cpp/include/nros/std_detect.hpp`, and W2 DELETED the
+# discovery disjunct there. 1240's objection to deleting it — "nothing that
+# ships defines `NROS_CPP_STD`, so the arm is what gives the hosted build
+# `std::string`-taking `get_logger`, the `SharedPtr` aliases and the `<vector>`
+# options surface" — was correct about the tree as it stood, and W2 is what
+# makes it stop being true: five consumers now make the request out loud (the
+# rclcpp compat shim, the phase-417 ported-surface probes, the four refusal
+# probes, the `cpp_compat_snippets` compile-check arm, and
+# `check-cpp-capability-layout`'s hosted arm). A better probe was still a probe;
+# the macro answers which SURFACE a consumer asked for, which no probe can see.
 #
-# The walker was tightened in the same change so it cannot be satisfied by the
-# TOKEN alone: an `#if`/`#elif` naming NROS_CPP_STD that also reaches
-# `__has_include` without `__STDC_HOSTED__` is NOT a std region. Without that,
-# the broken shape and the corrected one score identically, which is the 0196
-# rule — a gate whose reach is narrower than the rule it enforces.
+# Measured on the toolchain that could not build at all before (arm-none-eabi
+# 13.2, `-std=c++14 -ffreestanding -fno-exceptions -fno-rtti`), per-header
+# standalone parse:
+#
+#     before W2   pass=26 fail=20
+#     after W2    pass=47 fail=0
+#
+# The walker was tightened by 1240 so it cannot be satisfied by the TOKEN alone:
+# an `#if`/`#elif` naming NROS_CPP_STD that also reaches `__has_include` without
+# `__STDC_HOSTED__` is NOT a std region. Without that, the broken shape and the
+# corrected one score identically, which is the 0196 rule — a gate whose reach
+# is narrower than the rule it enforces. After W2 no capability block reaches
+# `__has_include` at all, so the tightened walker has nothing left to reject
+# here; it stays because the rule outlives this file's emptiness.
 #
 # Keyed on FILE and HEADER, not on a line number, so an entry survives edits
 # above it and so a second `<memory>` appearing in a listed file is still
 # caught. Add a line here only with a written reason and a tracked issue id.
+#
+# The file stays rather than being deleted: the gate requires it, and the next
+# genuine hosted-include debt needs somewhere to be recorded with its reason
+# instead of being argued about in a review. An entry that stops offending
+# FAILS, so an empty file cannot quietly refill.
--- a/cmake/compat/NrosRclcppCompat.cmake
+++ b/cmake/compat/NrosRclcppCompat.cmake
@@ -114,6 +114,24 @@ function(_nros_compat_apply_force_includes target)
     # header) resolve to the nano-ros compat shims.
     target_include_directories(${target} PRIVATE
         "${_nros_compat_dir}/include")
+    # phase-438 W2 — ASK for the std surface. A file reaching
+    # `<rclcpp/rclcpp.hpp>` subclasses `rclcpp::Node`, whose signatures are
+    # spelled in `std::shared_ptr` / `std::string` / `std::vector`; that surface
+    # is now reachable only through `NROS_CPP_STD`, never discovered from the
+    # include path (issue 1187 — `__has_include(<string>)` is TRUE on
+    # arm-none-eabi `-ffreestanding`, where including it is a hard `#error`).
+    #
+    # This target is exactly the right place and nowhere else is: the shim is
+    # applied per-target and is deliberately NOT auto-applied on Zephyr (see
+    # below), so the opt-in follows the porting surface instead of leaking to
+    # images that use `nros::Node` directly.
+    #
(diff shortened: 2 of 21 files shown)


jerry73204:
### Retargeted to `main` — unstacked

This PR was based on `fix/1223-freestanding-includes-elif` (#763). A stacked PR cannot arm
auto-merge — auto-merge is a property of the BASE branch, and a feature branch has neither the
ruleset nor the queue — and it also never enters the queue, so it forfeits the speculative build
and serialises what stacking was reached for. Rebased onto `origin/main` and retargeted.

**Base:** `fix/1223-freestanding-includes-elif` -> `main`.

#### Commit accounting — 4 -> 3

| commit | origin | outcome |
| --- | --- | --- |
| `36831dea5` docs(phase-438, #1223, #1187) | inherited | **DROPPED** by the rebase — patch-identical to `1f9d6d54f`, already on `main` (`git log --cherry-mark` marks the pair `=`) |
| `4938c6750` fix(#1223, phase-438 W0) | inherited from #763 | **CARRIED** — see below |
| `9d7f54b8d` refactor(phase-438 W1) | this PR's own | replayed as `6542a31e1` |
| `d301bd7a7` feat(phase-438 W2, #1187) | this PR's own | replayed as `4cd9f515c` |

`origin/main..HEAD` was 4 commits, is now 3.

#### Why W0 is carried, and why that is safe

W0 creates `.config/cpp-freestanding-includes-baseline.txt`, which does not exist on `main`
(`git cat-file -e origin/main:.config/cpp-freestanding-includes-baseline.txt` fails). Both W1
and W2 *edit* that baseline, so without W0 they do not apply and `check-cpp-freestanding-includes`
has no baseline to score against. It is the minimum needed and nothing else was taken from #763.

The queue rebase-merges rather than squashes, so once #763 lands the carried commit drops by
patch-id on the next rebase with no intervention. Verified equal, not merely similar:

```
git show 4da00cd76 | git patch-id --stable  -> 92a3a33dd0d21d4b238fa43df731f8655035c83b
git show 30257e34c | git patch-id --stable  -> 92a3a33dd0d21d4b238fa43df731f8655035c83b   (#763 tip)
```

#### Conflicts

None. The rebase replayed all four commits with no textual conflict; the only non-replay was the
docs commit git itself dropped as already upstream. The prior CONFLICTING/DIRTY state was an
artifact of #763 having been rebased out from under this branch, not a content collision. No
manual conflict resolution was performed, so there is nothing non-mechanical to disclose.

#### Gates (run on the rebased tree)

- `just check fast` — 281 of 283 green. The two reds are **environmental**, both "submodule /
  vendored tree not checked out" in a fresh worktree, neither touched by this diff:
  - `capability-conditionals` — needs `zenoh-pico`. After
    `git submodule update --init packages/rmw/zenoh/zpico-sys/zenoh-pico` it reports
    `OK (5 platform manifest(s), 5 capability-gated row(s), IP_STACK = 'ip_stack')`.
  - `xrce-vendored-versions` — needs `nros setup --source micro-cdr --source
    micro-xrce-dds-client`; both trees are present in the primary checkout and this PR touches
    no XRCE path.
- `cargo +nightly fmt --check --all` — clean (this PR contains no Rust).
- `just check cpp` — `All C++ checks passed!`, including `check-cpp-capability-layout: OK — 5
  type(s) x 7 forced capability macro(s) hosted, plus a -nostdinc++ freestanding measurement
  against the ThreadX shim`.
- `just check c` — `All C checks passed!`.

Auto-merge is armed against `main`.

jerry73204:
Rebased onto `main`, and **this PR would have turned `check-cpp` red on `main` had it landed alone.** Reproduced here before fixing it.

## The problem

W2 makes the C++ std surface REQUESTED rather than discovered. Two compile probes — `one_node_type.cpp` and its sibling — were written after this branch was cut, by phase-427 / phase-430, and ask for the hosted surface without making that request. Against W2 they stop compiling:

```
19 errors, e.g.
  226 |     (void)node.declare_parameter<int64_t>(std::string("count"), 3);
      |                                         ^
```

`just check cpp` exits 1 on this branch rebased onto today's `main`, and exits 0 on `main` itself — so the red belongs to the pairing, not to either side. Neither branch's own PR lane sees it: the probes are new on `main`, and W2 is new here.

## The fix

Cherry-picked `fix(check): two phase-427/430 probes ask for the hosted surface W2 made a request` from #784, which found it. It touches only `just/check/lanes.just` and one compile TU — nothing in `nros.hpp` or `node.hpp` — so it is independent of #784's W3/W4 work and drops by patch-id when whichever of the two lands second.

After it: `just check cpp` → **All C++ checks passed!**, `just check c` → OK.

## Rebase

3 commits before, 3 after. One dropped: `fix(#1223, phase-438 W0)`, verified on `main` as part of #763.

The one conflict was positional — `nros.hpp`'s `#include "nros/std_detect.hpp"` against main's edits to the same include block; main's side was empty.

jerry73204:
Rebased onto `main` (was `e45594889`, now `2077ae65a`). 3 commits before, 3 after; no commit dropped.

## Conflicts and how each was resolved

**`main`'s `fix(#1240)` is the whole collision.** It landed the corrected two-probe predicate
`defined(NROS_CPP_STD) || (defined(__STDC_HOSTED__) && __STDC_HOSTED__ && __has_include(<x>))` at
all fourteen sites and emptied `.config/cpp-freestanding-includes-baseline.txt` — the same fourteen
sites W1 consolidates and W2 deletes discovery from.

- **10 headers** (`client` `fixed_string` `heap_string` `log` `options` `polling_subscription`
  `publisher` `service` `subscription` `timer`): took W1's side, i.e. block deleted, one
  `#include "nros/std_detect.hpp"`. Verified this loses nothing else from `main` — each resolved
  file is byte-identical to W1's original, so `#1240` touched only the blocks W1 removes.
- **`std_detect.hpp` at W1** now carries `#1240`'s predicate, not W1's original
  `#elif defined(__has_include)` shape. W1 claims BEHAVIOUR-NEUTRAL and on the old base it was;
  post-rebase, neutral means `main`'s predicate. It is also required rather than cosmetic:
  `#1240` tightened the walker so an `#if` naming `NROS_CPP_STD` whose live arm reaches
  `__has_include` without `__STDC_HOSTED__` is **not** a std region, which W1's original shape is.
- **`std_detect.hpp` at W2** is byte-identical to the original W2 in its code; only the rationale
  comment differs, rewritten to record `#1240` as the step before rather than as a competing fix.
- **`.config/cpp-freestanding-includes-baseline.txt`**: `main` emptied it at W1's position, so W1
  adds no ratchet entry. Its prose was rewritten rather than either side taken: `main`'s text says
  deleting the `__has_include` arm "would have been the wrong fix" *because nothing that ships
  defines `NROS_CPP_STD`* — true of the tree it was written against, and W2 is exactly what stops
  it being true (five consumers now make the request). Both steps are recorded, in order.
- **`scripts/check-cpp-capability-layout.sh` was DELETED on `main`** by `fix(#1225)`, which
  rewrote it as `scripts/check-cpp-capability-layout.py` deriving ~90 subjects instead of three.
  W2's edit to it — `HOSTED_FLAGS` gains `-DNROS_CPP_STD=1` — was ported to the `.py`, comment
  included, rather than resurrecting the deleted file. Not a new configuration: before W2 the
  hosted arm got the std surface because `__has_include` found libstdc++; the flag names the same
  one. Confirmed neutral by the gate running green with `main`'s 72-row baseline unmoved.

## The load-bearing third commit

`fix(check): two phase-427/430 probes ask for the hosted surface W2 made a request` survived the
rebase and is still what keeps the lane green — `just check cpp` prints **`All C++ checks passed!`**.
No new probes needing the flag have landed on `main` since it was written (its own sweep covered all
41 TUs under `packages/api/nros-cpp/tests/compile/`).

## Gates

`just check cpp` **OK** · `just check c` **OK** · `check doc-refs` **OK** · `check issue-ids` **OK**
(1211) · `check prose-issue-refs` **OK** · `check doc-recipe-refs` **OK** ·
`check-doc-commit-citations.py` **OK** (196 citations) · `cargo +nightly fmt --check --all` **OK**.

## One thing a reviewer should decide, not me

`#1240` almost certainly already fixes issue **1187** on `main` — under `-ffreestanding`
`__STDC_HOSTED__` is 0, so its conjunction is false and the poisoned headers are never included.
W2's commit message and the 1187 archival still read as if W2 is what fixes the FreeRTOS build.
W2's *own* argument is unaffected and is not about 1187: a corrected probe is still a probe, and a
capability macro answers which surface the consumer asked for. I left the message and the archival
as authored rather than rewriting the claim; say the word if you want 1187 re-opened and closed by
`#1240` instead.

jerry73204:
Superseded by #799, and by `c9c861360 fix(#1240)` on `main`.

This PR bundled phase-438 **W1** (consolidate the fifteen hand-copied capability blocks into one `std_detect.hpp`) with **W2** (delete the discovery arm, so the std surface is reachable only through `NROS_CPP_STD`). While it was open, a peer session hit the same `requires_hosted.h` wall on a GCC 16 host, diagnosed it independently as issue 1240, and fixed it with the `__STDC_HOSTED__ && __has_include` conjunction applied to all fourteen blocks. That landed first and it works.

**W1 is still worth having and is now #799**, rebuilt on top of 1240's predicate rather than replacing it — measured neutral against `main` on both toolchains.

**W2 is deferred as a design decision, not abandoned.** It remains written up in phase-438.

One correction to this PR's own reasoning, since it is quoted in the commit messages here: it claimed that issue 1187's second fix option — ANDing the two probes — "fails on whichever lane it is not measured on". That is **wrong**. It is true of each probe alone, and I carried it over to the conjunction without checking. The conjunction satisfies all three measured lanes, which is what 1240 demonstrates in practice.

Conversely, 1240's resolution argues that deleting the arm "would have removed `std::string`-taking `get_logger`, the `SharedPtr` aliases and the `<vector>` options surface from EVERY build including the hosted one". That is true of deleting the arm and adding nothing; this PR added `NROS_CPP_STD` to the five consumers that need it and measured 59 example TUs with 0 regressions. Both halves of the disagreement were overstated.

PR #773, opened 2026-09-08
phase-428: rclcpp:: is the home — fix the parity extractor, and three corrections the design owed

jerry73204:
The maintainer ruled that `rclcpp::` is the HOME of our C++ API. PR #755 found
that defining the merged node class there turns `just check api-parity` red and
kept the class in `nros::` with an alias; this branch changes the tool instead,
and lands three corrections the design owed.

Base is `phase-428-w10-qos-ssot` (#629 — it holds the RFC-0089 sections edited
here). No file of #755's is touched.

## 1. The extractor: our surface is BOTH vocabularies

The C++ parity lane rooted OUR native surface at namespace `nros` alone. A class
defined in `rclcpp::` was therefore extracted only by the compat translation
unit, landed `surface: ported`, and its `native_bucket` — the bucket `--check`
gates — read **`theirs-only`**. About sixty members of the merged node class
re-bucketed as names only ROS 2 has. The instrument was measuring the half of our
vocabulary we are deprecating.

Every C++ TU is now rooted at `OUR_CPP_ROOTS = {nros, rclcpp, rclcpp_action,
rclcpp_lifecycle}`.

**This widens what counts as OURS, not what counts as THEIRS.** `theirs` comes
from the recorded upstream surface and never from a namespace filter, so a name
rclcpp declares that our headers do not still correlates `theirs-only`.

### Bucket counts, before → after (C++)

| bucket | before (gated native surface) | after |
| --- | ---: | ---: |
| `same` | 88 | 135 |
| `arity-only` | 12 | 11 |
| `systematic` | 12 | 9 |
| `differs` | 20 | 21 |
| `ours-only` | 354 | 360 |
| `theirs-only` | **695** | **651** |

Forty-four names we ship stopped being reported as names only ROS 2 has. C and
Rust are unchanged (`101/0/21/8/372/354` and `89/7/2/5/1240/502`).

### The mutation proof

**Direction 2, on the real lane.** Four upstream-only names read `theirs-only`
before and after the change: `WallTimer`, `Waitable`, `Waitable::is_ready`,
`uninstall_signal_handlers`.

**Direction 1, end to end through clang.** Declaring
`rclcpp::uninstall_signal_handlers` — an upstream-only name — in one of our
headers and re-running:

| tool | ported summary | gated (native) summary |
| --- | --- | --- |
| **before** this branch | `same` 135 → 136, `theirs-only` 651 → 650 | **`theirs-only` 695 → 695, unmoved** |
| **after** | one surface: `same` 135 → 136, `theirs-only` 651 → 650 | same line, and it moved |

That is #755's symptom reproduced on demand: the old tool saw the declaration and
the gated surface did not. Declaration reverted; the tree is clean.

**Both directions are also `--self-test` cases** (`just check api-parity-ledger`),
mutation-tested: narrowing the roots back to `{nros}` fails *"our home namespace
counts as ours"*; adding `rclcpp::detail` fails *"the widening is a named list,
not a wildcard"*.

### Rows the widening forced us to ledger: **zero**

The gate got strictly stronger — it used to enforce the native surface and merely
REPORT the ported one, and now enforces the union — and `--check
--require-disposition` stayed green with no new rows, because the ledger had
already been written against the union. All 17 shards re-parsed with a
duplicate-key-raising `object_pairs_hook`: 0 duplicates.

### The `compat` TU is deleted, and that is measured

Its whole remaining content was the namespace it admitted; once that namespace is
ours, the `std` TU (same source, same flag, superset of roots) emits every record
it could, and its records de-duplicate away before the mark is applied. Measured:
`ours_cpp` returns 162 records, **0** carrying `surface: ported`. A TU that can
never contribute reads live and answers nothing, so it goes rather than standing
inert. The `surface` / `native_bucket` / `--check-ported` machinery stays — one
tuple row re-admits a genuinely separate ported header. `SURFACE_NOTE` goes too:
it had no reader anywhere in the tree.

**The flip is recorded as unblocked**, with what remains measured rather than
estimated: nine types are defined in `nros::` and aliased into upstream's
namespace (`Clock`, `Duration`, `Time`, `Publisher`, `Subscription`, `Client`,
`Service`, and the two action types under `rclcpp_action`), each with header and
line. `Node` is the tenth and is #755's file.

### (a) The ported tutorial is SIX edits, not two — 2 sites

The design and phase-427 W10 both claimed two. PR #753 measured it against
upstream's text, port held as two files and diffed at build time: **six**. Two
imports (any port has them), one error type — possible at all only because that
wave gave `NodeError` a `Display` impl, so the original claim was not merely
undercounted but unreachable — one `mut` binding rustc names, and the two
predicted `?`s.

A **seventh** difference is recorded rather than absorbed: the port writes
`spin_blocking` because `Executor::spin` is taken by `spin(Duration) -> !`. It
falls on a line that changes anyway, so it costs no edit — which is how a
divergence stops being visible. Moving upstream's `spin(SpinOptions)` onto that
name is a later wave, now an **open item** in both documents.

### (b) The RFC-0047 citation is wrong — 8 sites (6 design, 2 phase-427)

RFC-0047 is *"Unified sched-context binding via callback groups"*. It says nothing
about a second node identity; its one sentence about several nodes cites RFC-0046.

**The capability does not exist either.** `ComponentNode` owns exactly one
`Node node_` (`component_node.hpp:734`) — RFC-0044 Q1 chose WRAP over derive
precisely so it holds one. Both subnode packages open with *"ONE ComponentNode
with TWO callback groups"* and create `ctrl`/`telem` on one `sub_node`. They are
the RFC-0047 proof, just not of this — which is why the citation read as
plausible.

What is real is several named nodes per **IMAGE**, owned by the **executor**:
`Executor::create_node` / `node_builder(name)` (`executor.hpp:180`, `:189`),
under **RFC-0046**, whose summary says a multi-component launch yields
per-component graph nodes — one each.

Applied: the merge plan's **third constructor is dropped** (two, not three); no
ledger row is owed for a several-identities divergence; **W4's acceptance is
amended**, keeping its subnode-package clause, which was correct.

### (c) The alias rule, narrowed to what a gate proved — general rule + 5 sites

`just colcon-parity` builds the example templates against REAL rclcpp, so a
template must compile under upstream *and* here. Upstream has `rclcpp::TimerBase`
and **no** `rclcpp::Timer` (verified against the recorded surface: `TimerBase`,
`WallTimer`, `GenericTimer` present; `Timer` absent); ours is the reverse. With
the alias deleted, for any node holding a timer the intersection is **empty**.

> For a name upstream HAS and we lack, the ported alias is LOAD-BEARING and must
> NOT be deprecated — a deprecation promises a removal that empties the
> intersection on a schedule. "Do not take the name" is available only where
> upstream lacks it too.

Recorded that §"Review of the invented parts" item 1 — which proposed exactly this
alias — **was right**, and that the section which withdrew it keeps its whole study
of the HIERARCHY, still the reason `Timer` is flat. The collision gate's scope is
narrowed in prose to what it already meant: it fires on ours-only EXTENSIONS, so a
ported alias appearing in the upstream surface is correct, not a collision.

Part V's two surface sections are marked as history rather than rewritten, since
their measurements are the evidence for decisions that stand.

## Run

| lane | result |
| --- | --- |
| `just check api-parity` | **green** — "every divergence carries a ledger entry" |
| `just check api-parity-ledger` | green (self-test 0 failed; gen-api-comparison 0 failed) |
| `just check ledger-orphan-refs` | green — 17 shards, every cited path resolves |
| `just check doc-refs` | green |
| `just check gate-selftests` | green — 170/282 |
| ledger shards re-parsed, duplicate-key hook | 17 shards, **0** duplicates |
| `just check fast` | 274 of 278 — the four below |

Environmental, named not fixed, none touching a file this branch edits:

* `capability-conditionals` — wants the zenoh-pico submodule
* `xrce-vendored-versions`, `xrce-source-manifest` — want the XRCE submodule
* `doc-commit-citations` — one baseline entry (`d1d88f660`) is no longer needed
  because this clone resolves it; the citation is in
  `docs/issues/1127-live-peer-cells-skip-forever.md`, untouched here

No auto-merge.


jerry73204:
## Unstacked: base is now `main`

This PR was based on `phase-428-w10-qos-ssot` (#629). A stacked PR cannot arm auto-merge —
auto-merge is a property of the BASE branch, and a feature branch has neither the `main-rules`
ruleset nor the merge queue, so GitHub refuses with *"Protected branch rules not configured for
this branch"*. It also never enters the queue, so it forfeits the speculative build and
serialises exactly what stacking was reached for. Retargeted with `gh pr edit 773 --base main`.

### Own vs inherited

`origin/phase-428-w10-qos-ssot..HEAD` was **not** the split, because #629 had already been
rebased onto `main` — every one of its commits is a new object, so that range reported all 39.
The real boundary is #629's **pre-rebase** tip, `10765b1f2`:

| | count |
| --- | --- |
| #773's own commits (`10765b1f2..HEAD`) | **6** |
| inherited from #629's old tip | **33** |
| #629 after its own rebase onto main | 31 |

#773's own six touch exactly three files: `docs/design/0089-ros2-api-adoption-and-the-compile-or-conform-rule.md`,
`docs/roadmap/phase-427-one-node-type.md`, and `scripts/api-parity.py`.

### What was carried from #629, and why

**All 31 commits of the rebased #629**, by rebasing onto `origin/phase-428-w10-qos-ssot` and
then onto `origin/main`. This is not padding — it is a hard textual dependency:

* `scripts/api-parity.py` — #629 rewrites 547 lines of it; #773's headline commit
  (`fix(api-parity): root our C++ surface at both vocabularies`) edits the very block #629
  introduces. Detaching them is not a rebase, it is a rewrite.
* `docs/roadmap/phase-427-one-node-type.md` — #629 adds 28 lines, #773 adds 56 more to the same doc.
* The ledger shards and `check-ledger-orphan-refs` that #773's extractor change is *measured
  against* come from #629. Without them `just check api-parity-ledger` has nothing to compare.
* `docs/issues/1099-…md` is added by #629's `3b8c4d7f0`, and #773's baseline fix below depends on it.

Because the queue rebase-merges rather than squashes, these 31 drop by patch-id on the next
rebase once #629 lands. Expected and safe.

### Commit counts

**BEFORE 39 → AFTER 38** (37 rebased + 1 new).

Two inherited commits dropped in #629's rebase, and both are genuinely superseded on `main`:

1. **`ec92b5645 feat(phase-428 W6): [[nodiscard]] on Result`** — superseded by
   `a17c1beea feat(phase-427 W8): one error-channel template, NROS_NODISCARD`. Verified by
   reading `origin/main:packages/api/nros-cpp/include/nros/result.hpp`: `NROS_NODISCARD` is
   already on `ResultOf<void>`, on `ResultOf<T>`, and on the deprecated `Expected<T>`. Main's
   phase-427 W8 shape (one `ResultOf<T>` template, `Result` an alias for `ResultOf<void>`,
   `Expected<T>` kept one release as `[[deprecated]]`) wins, and it already carries the
   loudness this commit was adding to the older split shape.
2. **`3fe45b800 docs(phase-428): cite #557 by PR number`** — superseded by
   `5753d8d3c docs: drop three commit hashes that no clone can resolve`. That commit solved the
   same problem the other way: `main`'s line 516 now reads ``unsupported until
   `ci/land-node-std-gate` merges`` with no hash at all. Re-applying this commit would have
   *re-introduced* a bare `00e8f3087` in backticks, which `check-doc-commit-citations` on main
   would then have to carry.

### Non-mechanical resolutions

**One real conflict**, in `scripts/api-parity.py` at `CPP_TRANSLATION_UNITS`:

* #629 (HEAD side) added a 16-line comment on the `component` TU recording the 2026-09-07
  measurement that it contributes zero records and is kept as a *guard* for issue 0818 rather
  than as a supplier.
* #773 (incoming side) changed the namespace roots on all three TUs from `{"nros"}` to
  `OUR_CPP_ROOTS`.

These are independent edits to the same lines — a comment addition and a value substitution.
Resolved by keeping #629's comment block verbatim and applying #773's root substitution to all
three entries. Verified with `ast.parse`, then by `just check api-parity` (green inside
`check fast`) and `just check api-parity-ledger` (green, self-test 0 failures). No ledger row
changed as a result, so the widening did not silently re-bucket anything.

**One new commit**, `fix(check): the 1099 rows are discharged`. `check-prose-issue-refs` is a
ratchet and it went red on the rebase: two rows —
`docs/roadmap/phase-428-api-porting-principle-sweep.md:1099` and `just/check/lanes.just:1099` —
were classified `IN FLIGHT` on 2026-09-08 with the note that the issue file *"exists on
`origin/design/context-init-shape`"* and *"the file is coming"*. It has come: `3b8c4d7f0`,
carried here from #629, adds `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md`,
so both citations resolve on this branch and the rows describe nothing. Edited **by hand**, not
with `--write-baseline`, because that flag regenerates rows only and destroys the `CLASSIFIED`
block that is the sole record of why the remaining seven are legitimate.

> **Note for #629:** the same red is latent there. `3b8c4d7f0` is on #629 too and #629 does not
> touch `.config/prose-issue-ref-baseline.txt` (`git diff origin/main..origin/phase-428-w10-qos-ssot
> -- .config/prose-issue-ref-baseline.txt` is empty), so if #629 reaches the queue first it will
> eject on `check-prose-issue-refs` for exactly this reason. Cherry-picking `4428274e4` onto it
> fixes it, and the duplicate drops by patch-id here.

No ledger key duplication occurred — the ledger shards took no conflict on this rebase, and
`just check api-parity-ledger` (which is what catches git's silent duplicate-key merge) is green.

### Gates

Run on `4428274e4`:

| gate | result |
| --- | --- |
| `just check fast` | **1 of 285 failed** — `xrce-vendored-versions` only |
| `cargo +nightly fmt --check --all` | green |
| `just check api-parity-ledger` | green (self-test 0 failed; gen-api-comparison self-test 0 failed) |
| `just check ledger-orphan-refs` | green (self-test 10 cases OK; 17 shards, every path and bare filename resolves) |
| `just check doc-refs` | green |
| `just check issue-ids` | green (1190 issues, unique and self-consistent) |
| `just check prose-issue-refs` | green (7 known in-flight, no new dangling reference) |
| `just check cpp` | green — *All C++ checks passed*, including the three phase-427 W8 cases (checked result compiles, discarded result fails, `Expected<T>` deprecated) |

`xrce-vendored-versions` is **environmental**, not a branch failure: the Micro-XRCE trees are
`nros setup --source` provisioned rather than submodules, and are absent from a fresh worktree —
the gate says so itself (*"no vendored tree checked out — nothing verified"*).
`capability-conditionals` failed for the same class on the first pass and went green after
`git submodule update --init packages/rmw/zenoh/zpico-sys/zenoh-pico`; `check cpp` needed the
same submodule and is green with it.

No submodule pin moved: `git diff origin/main..HEAD --submodule=short` contains no
`Subproject commit` line.

### Final state

`base: main` · `mergeable: MERGEABLE` (was `CONFLICTING`/`DIRTY`) · `autoMergeRequest` non-null,
so auto-merge is armed · `mergeStateStatus: BLOCKED`, which is the normal pre-queue state while
required checks run.

jerry73204:
## Rebased onto `main` (`632bdbeba`)

`d6e2a4396` → `dfbae47eb`. **45 commits before, 46 after** — nothing dropped, one added.

**No-loss verification:**

```
comm -23 <(git log --format=%s origin/main..d6e2a4396 | sort -u) \
         <(git log --format=%s origin/main..HEAD       | sort -u)
```
→ **empty**. All 45 replayed; no patch-id drops (this branch carries #704's family, and #704 has not landed yet, so nothing on `main` could absorb them).

Note the local ref was **behind** `origin/phase-428-rclcpp-is-the-home` (`3170a4359` vs `d6e2a4396`) — the rebase was done from `origin/`, not from the stale local branch, so nothing was silently reverted.

### Conflicts, and what was measured

**1. `packages/api/nros-cpp/include/nros/nros.hpp`** (`224a2c653`) — the phase-427 node merge on `main` deleted the shim `class Node`; `rclcpp::Node` is now an unconditional `using Node = ::nros::Node` and the seven `create_*` verbs are out-of-line `nros::Node::create_*` definitions. `main` wins on structure; every shim member the branch side re-declared was verified present on `main` by name before dropping. The commit's actual payload — the seven verbs going through `detail::require_created` instead of `(void)`, so a failed create ABORTS rather than returning a valid-looking pointer to a dead entity — was **ported onto `main`'s shape**, along with the `shutdown()` ADOPT→ADOPT-BOUNDED doc revision. The resolution is **byte-identical to PR #704's** (`diff` against `d9b0168e8:…/nros.hpp` is empty), so the two branches cannot diverge in the queue.

**2. An EIGHTH discarded create.** `main`'s phase-430 W6 landed `rclcpp::create_timer(node, clock, period, callback)` as a free function *after* the branch's sweep ran, with the same discarded `Result` (`nros.hpp:865`). Fixed as part of the class, not left as a survivor — and it is the quietest member, since a timer handle is stored and never dereferenced.

**3. `docs/reference/api-parity-ledger/timer.json` — `cpp:create_timer` gained a `disposition`.** A gate failure the rebase created, not a conflict. This branch widens `--require-disposition` to `same_shaped_divergences`; `main` supplied it a subject. Confirmed the rule is this branch's: `git show origin/main:scripts/api-parity.py | grep -c 'SHAPE corresponds'` → `0`. Value measured from `main`'s own comment above the function — *"the same ADOPT-BOUNDED trade `get_clock()` itself makes"* (`nros.hpp:815`) — so `adopt-bounded`, `why` recording the four bounds (`nros::Clock*` vs `Clock::SharedPtr`, millisecond truncation, no `group` parameter, `shared_ptr<nros::Timer>` vs `TimerBase::SharedPtr`). **No re-verdict.**

**4. `docs/roadmap/phase-427-one-node-type.md`, W4 bullet** (`13d8bd324`) — **`main` was internally inconsistent and the branch fixes it.** `main`'s W4 bullet says RFC-0047's several-named-nodes "turns out not to exist (it is several named CALLBACK GROUPS)" *and*, three lines later, that it "survives as a documented ours-only capability on it". `main`'s own §"W4's RFC-0047 acceptance criterion is deleted, not weakened" (line 730) already rules the first way. Resolution keeps `main`'s newer NOT STARTED framing and the four blockers it lists, and takes the branch's strike of the survives-clause with its measurement (`component_node.hpp:734` owns exactly one `Node`; both subnode packages open "ONE ComponentNode with TWO callback groups"); a cross-reference to `main`'s independent ruling was added so the two read as one.

**5. `phase-427-one-node-type.md` timer row + `RFC-0089` timer paragraph** (`f8f1756f5`) — the interesting one. Both sides reach the same conclusion by different routes, and **the tree decides which wording is current**:

* `main`'s roadmap row: *"the NAME survives one release as a DEPRECATED alias"*.
* `main`'s own later commit `de0993456` ("the TimerBase NAME is load-bearing — colcon-parity said so"): *"the NAME is KEPT as the ported alias, and is NOT deprecated."*
* `packages/api/nros-cpp/include/nros/timer.hpp:274-310`: *"KEPT, and NOT deprecated, and the reason is MEASURED rather than stylistic."*

So `main`'s roadmap row and RFC paragraph are **stale against `main`'s own code** — the branch's wording is the correct one. Resolution takes the branch's substance (kept, not deprecated, load-bearing because `colcon-parity` builds our templates against real `/opt/ros` headers where `rclcpp::Timer` does not exist) and preserves `main`'s landing record (LANDED 2026-09-08, phase-430 W7; the hierarchy stays deleted) plus its pointer into §"AMENDED 2026-09-08". Both wrong answers this row previously carried — "withdrawn", then "deprecated for one release" — are named in the merged text rather than tidied away.

**6. `RFC-0089`, the spin-family section** (`1a4dd5fd7`) — **not a collision.** `main` inserted §"AMENDED 2026-09-08" and the branch inserted §"The `spin` family" at the same offset. Both kept, `main`'s first (it is the earlier date). Verified no duplicate heading afterwards.

**7. `phase-427-one-node-type.md`, W4 acceptance + W5 heading** (`549b49c79`) — two hunks, both additive-vs-status. Kept `main`'s W5 status (`DONE (compile half; runtime half owed)`) and the branch's W4 acceptance SPLIT (the "builds for a freestanding target" clause is a new PORT, not a check, and moves to W11) plus the whole `(a)`/`(b)`/`(c)` amendment block.

**8. `.config/prose-issue-ref-baseline.txt`** (`4428274e4`) — resolved BY HAND. `main`'s copy of this file is **sorted**, its CLASSIFIED header comments interleaved among the rows: someone ran `--write-baseline` on `main` and it destroyed the block's ordering. That damage is pre-existing on `main` and was deliberately **not** repaired here. The resolution keeps `main`'s file otherwise byte-for-byte — including its new `docs/issues/1231-…:1164` row — removes only the two `1099` rows plus the two stale note lines, and leaves the branch's discharge note in place of the `IN FLIGHT` line.

Discharge verified rather than assumed: `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md` **is** in the rebased tree and both citations still exist. `just check prose-issue-refs` reports `8 known in-flight` (the branch's own message says 7 down from 9, which was true against its old base; `main` has since added a row, so the measured figure is 8 down from 10).

### Gates

| gate | result |
|---|---|
| `just check fast` | **284 / 285** |
| `just check cpp` | OK — including `failed_create_aborts: OK` |
| `just check api-parity-ledger` | OK (self-test 0 failures) |
| `just check ledger-orphan-refs` | OK — 17 shards, every cited path and bare filename resolves |
| `scripts/check-doc-commit-citations.py` | OK — 206 citations across 899 files |
| ledger duplicate-key scan (`object_pairs_hook` across all 17 shards) | **none** |

The one `check fast` failure is `xrce-vendored-versions`, **environmental and reproduced on a plain `origin/main` worktree**: `check-xrce-vendored-versions: no vendored tree checked out — nothing verified`. `capability-conditionals` passed here.

State: **MERGEABLE**, auto-merge armed.

jerry73204:

`dfbae47eb` → `66c6d89ee`. **46 commits before, 47 after** (one new commit records the numbering reconciliation). No-loss verified:

```
comm -23 <(git log --format=%s origin/main..dfbae47eb|sort -u) \
# (empty)
```

Nothing dropped, nothing skipped by patch-id. Only two of the 46 conflicted.

**1. `docs/roadmap/phase-427-one-node-type.md` — a work-item NUMBER collided, and the design did not.**

This is the one worth reading. main landed the `spin` family rename as **phase-427 W11** (`cf4b6aada`, `ea96be9fb`, `161249a5c`, with a `## W11 [rust] … LANDED 2026-09-09` section at the end of the file). This branch's `090637b71` had queued the same work as **W12**, and given **W11** to the freestanding subnode port it split out of W4's acceptance. After the rebase the document had two W11s and a "planned" bullet for work that is done.

Checked before touching it: the two are the **same design**, not two designs. main's landed section and this branch's plan carry the same four verbs, the same signatures and the same dispositions (`spin` adopt-bounded, `spin_once` the tick primitive, `spin_some` adopt-bounded, `spin_forever` extension), both citing RFC-0089 §"The `spin` family: upstream's names get upstream's contracts", and main's section explicitly credits PR #753's port measurement as the reason. So this is a numbering collision, and main's number wins because it landed.

Applied:

| before (this branch) | after |
|---|---|
| W11 = freestanding subnode port | **W12** — with both cross-references in W4's acceptance moved with it |
| W12 = the `spin` family rename (planned) | **W11**, marked LANDED, pointing at main's own section rather than restating the plan |
| "OWNED 2026-09-09: that later wave is W12" | "that later wave is W11, and it LANDED the same day" — CLOSED |

The rewritten W11 bullet names the two things this branch's plan did not have, because they are the parts a reader would otherwise take from the stale plan: `SpinOptions::poll_interval` has a **hand-written** `Default` (the derive is `Duration::ZERO`, which would have made every `SpinOptions::default()` a busy-poll that still passes every functional test), and `stop_on_first_error` exists at all.

`docs/roadmap/phase-430-ros-time-and-clock-driven-timers.md` asked, in its own words, that the seventh porting difference's closure "have an answer in both documents rather than only in the one that raised it" — so its two `phase-427 W12` citations moved too.

**One claim corrected while moving it, because both documents had it wrong in the same direction.** Closing the RENAME does not drop the rclrs port's EDIT COUNT. main's own W11 section says so and this branch's plan did not: the ported line still differs, because ours is `?` where upstream is `.first_error()?` — the no-allocator divergence, not a naming one. The seventh *difference* is gone; the six *edits* are not five.

**2. `.config/prose-issue-ref-baseline.txt` — reapplied by hand onto main's repaired header.**

main's `fix(#1241)` repaired the scrambled CLASSIFIED header and added `check_baseline_shape`, which refuses a baseline whose comments follow the first data row. `d67fd03af` was written against the scrambled version, so git's line merge produced garbage. Resolution took main's repaired file (verified byte-identical to `origin/main`'s before editing) and removed only this commit's payload: the two `…:1099` rows and the 1099 IN-FLIGHT note, since this branch ships `docs/issues/1099-cpp-trigger-transition-crosses-lifecycle-ids.md`.

The branch's version left a parenthetical "(1099 was here … and is DISCHARGED)" in place of the note. That was dropped in favour of a plain deletion, deliberately: #629, #704 and #753 all carry a version of the same commit, and making all four produce a **byte-identical** file means whichever lands first, the others drop it by patch-id instead of conflicting on this one line. Verified identical against #629's tree.

#### Stack consistency

This branch sits on top of #629 / #704 / #753, so the shared commits were made to resolve identically rather than plausibly. Verified by `diff` against the rebased #704 tree:

- `rclcpp::create_timer`'s function body — identical;
- the `cpp:create_timer` ledger row's `why` — identical (one stale phrase, `"phase-428 rebase"`, was corrected to `"phase-427 W9 rebase"` to match, via a fixup into its own commit rather than a new one);
- `.config/prose-issue-ref-baseline.txt` — identical.

#### Ledger

All 17 shards scanned for duplicate keys with a `json.load(object_pairs_hook=…)` counter, since git's line-based merge creates them silently and JSON keeps only the last: **0 duplicates**.

#### Gates

| gate | result |
|---|---|
| `python3 scripts/check-doc-commit-citations.py` | OK (205 citations resolve) |
| `just check cpp` | OK |

The one `check fast` failure is `xrce-vendored-versions`, environmental rather than pre-existing on this branch: reproduced identically on a plain `origin/main` worktree, because a fresh worktree has no `nros setup --source micro-cdr --source micro-xrce-dds-client`.

jerry73204:
## Rebased onto `main` (`b5097a460`)

`bdf2fa07a` → `8dc07afec`. **47 commits before, 47 after.** No-loss check (`comm -23` on subjects, old tip vs new HEAD, both against `origin/main`) is **empty** — nothing dropped.

### Conflicts and resolutions (4 files, 3 commits)

**1. `docs/reference/api-parity-ledger/log.json`** (in `docs(phase-428 W5): 82 ledger rows`)
Resolved ROW-WISE then FIELD-WISE from the three conflict stages rather than by line, because git's line merge silently creates duplicate keys on this family and JSON keeps only the last. `ours=73 theirs=73 out=74` — the branch's W5 `why` appends and its `disposition: adopt` on `c:node_get_logger` merged in beside main's rows, plus one new row (`c:log_severity_t`). **Zero field-level collisions.** All 17 shards re-scanned with an `object_pairs_hook` counter afterwards: **0 duplicate top-level keys, 0 conflict markers.**

**2. `.config/doc-commit-citations-baseline.txt`** (in `docs(phase-428): restore the citation fix the rebase dropped`)
Composed, not picked. The branch DELETES the `RE-ADDED 2026-09-08` block (its own commit restores the citation fix, so `839f69b5c` / `fe807a999` resolve again and `d1d88f660` is already carried by the 1127 block above). Main independently ADDED the `e3786749a` / issue-1227 DTCM block. Kept both: theirs' deletion + main's new block, comments still above every data row.
Verified: `check-doc-commit-citations` OK — 206 citations across 913 live files resolve, 12 declared foreign. It reports `d97c9c606` and `e3786749a` as *possibly* no longer needed (not a failure); both are main's lines and are left alone.

**3. `scripts/api-parity.py`** (in `fix(api-parity): root our C++ surface at both vocabularies`)
Genuine collision on the `compat` TU: main (phase-438 W2) refined its comment and made the compat CMake path set `NROS_CPP_STD`; this branch DELETES the TU, because with `rclcpp` in `OUR_CPP_ROOTS` the `std` TU is a superset and `compat` contributes 0 records. Took the branch's deletion — that is the commit's whole point, and main's edit describes a construct the branch removes — but carried forward the one fact from main's block that outlives its TU (phase-438 W2 having `_nros_compat_apply_force_includes` set `NROS_CPP_STD`), **verified against `cmake/compat/NrosRclcppCompat.cmake:134` before writing it down.**

**4. `docs/roadmap/phase-427-one-node-type.md`** (in `docs(phase-427 W4): the _in resolution, the byte split…`), two hunks, both COMPOSED:
- W5 status line: base and branch both read `DONE (compile half; runtime half owed)`; main changed it to `DONE.` The branch never touched that line — the conflict was pure adjacency with its new "Amended 2026-09-09" block. Kept **main's `DONE.`** plus the branch's block.
- End of document: kept the branch's W11/W12 numbering paragraph AND main's new `## W1 is REVERSED by RFC-0096 / phase-442` section. Taking either alone loses a true fact.

### Gates

| gate | result |
| --- | --- |
| `cargo +nightly fmt --check --all` | 0 |
| `just check api-parity` | 0 |
| `just check api-parity-ledger` | 0 (self-test 0 failed) |
| `just check doc-refs` | 0 |
| `just check issue-ids` | 0 (1218 issues) |
| `just check prose-issue-refs` | 0 |
| `scripts/check-doc-commit-citations.py` | 0 |
| `just check cpp` | 0 — all cases |
| `cargo check -p nros-node --all-targets --features std` | 0 |

`xrce-vendored-versions` is ENVIRONMENTAL, and that was measured rather than assumed: a control worktree on plain `origin/main` (`b5097a460`, nothing of this branch) fails it identically — `no vendored tree checked out`, the XRCE sources are a `nros setup --source` provisioning step no worktree has.

Submodule pins unchanged from `origin/main` (zenoh-pico initialised at its recorded pin `dd071b8d3`, not moved).

jerry73204:
## Rebased onto `main` after #629 landed — 47 commits → 14

`8dc07afec` → `765521076`, base still `main`. As on #704, most of what fell out is this
branch's own work re-authored on `main` by #629. What remains is the RFC-0089 /
phase-427 documentation this PR is actually for, plus `scripts/api-parity.py`.

### No-loss verification

`comm -23` over commit subjects left **34** dropped subjects; **33 resolve on `main`**
by subject. The one that does not:

* `fix(check): the 1099 rows are discharged — this PR ships the file they cite` —
  on `main` as `c1db3a752` under a different subject (`fix(#1099): the ratchet's two
  IN-FLIGHT rows are discharged by this branch`). Verified as a **byte-identical patch**:
  same two data rows and the same IN-FLIGHT comment paragraph removed from
  `.config/prose-issue-ref-baseline.txt`, with the CLASSIFIED header block above the
  first data row left intact in both.

### Non-mechanical resolutions

**1. `docs(phase-428 W5): 82 ledger rows` — skipped as subsumed by `86792f9ea`.**
Patch-diffed against `main`'s: the only difference in the whole 13-file commit is the
**field order** inside one row, `c:parameter_declare_*` in `param.json`
(`verdict, disposition, bucket, why` here vs `verdict, bucket, why, disposition` on
`main`). Compared as dicts: `d == b` is `True`.

**2. `fix(phase-428): eager-created C entities resolve their names` — skipped; its
prose delta re-landed as its own commit.** `main` carries the code as `8b231b484` +
`eb3339dae` (9 `require_created` occurrences: definition + 8 call sites). What `main`
does **not** carry is three lines of `detail::abort_failed_create`'s doc block, which
still describes the pre-phase-427 shim. Rather than assume "subsumed", those three lines
come back as `docs(nros.hpp): the create-refusal comment still described the retired
shim`, measured against the header as it stands:

  * `(void)node_.create_…(…)` names a member that no longer exists — the seven verbs are
    out-of-line `nros::Node::create_*` definitions writing `this->create_…`
    (`nros.hpp:524`, `:576`, `:618`, `:725`, `:736`, `:752`, `:763`).
  * "on `rclcpp::Node`" holds only through the alias (`using Node = ::nros::Node`), which
    is what makes the following paragraph about out-ref `nros::Result` verbs read.

  Comment only. `just check cpp` passes with `failed_create_aborts: OK`, the
  expected-failure probe that instantiates exactly this path.

**3. `docs(phase-428 Q1): every live refusal site has a row citing it` — skipped as
subsumed by `678ac9d71`.** Compared row-wise across all 7 shards: **20 rows changed, 18
byte-identical to `main`**. The two that differ, `cpp:SystemDefaultsQoS` and
`cpp:SystemDefaultsQoS::SystemDefaultsQoS`, differ because `main` is **later and more
correct**: this branch's copy cites `rclcpp_compat.hpp`, which phase-417 stage 6 step A
deleted, while `main` cites `packages/api/nros-cpp/include/nros/qos.hpp:767` with the
retirement noted parenthetically. Taking this branch's side would have re-introduced a
citation of a file that does not exist — which is precisely what this PR's own
`check-ledger-orphan-refs` refuses.

**4. `ledger(phase-428 Q1 leftovers): measure the `env` surface` — skipped as subsumed
by `9e3dc52f2`.** Both non-JSON files (`phase-428-api-porting-principle-sweep.md`,
`scripts/api_parity/extract_rust.py`) have identical patches. Of 64 changed ledger rows,
**56 are byte-identical** to `main`; the 8 that differ all differ because `main` evolved
them **further**, verified individually:

  * `rust:Context::default_from_env` and `rust:init` — re-verdicted `gap`/`extension` →
    `rename` on `main` by `e2c7f823d`, with a `rename` block naming the decision.
  * `rust:Context` — `main` adds a sentence pointing at that `rename`.
  * `cpp:Node::{declare,get,has}_parameter` — `main` replaced brittle line numbers
    (`nros.hpp:834` / `:851` / `:875`) with a description of the out-of-line definitions.
    Word-level diff shows that is the **only** change.
  * `rust:Context::new`, `rust:Context::from_env` — this branch's `why` is a literal
    substring of `main`'s.

**5. `docs(phase-428): restore the citation fix the rebase dropped` — dropped
deliberately.** Its entire purpose is on `main` as `cc428cf8d`; after the rebase the only
residue was the removal of a **trailing blank line** in
`.config/doc-commit-citations-baseline.txt`. Keeping a 30-line commit message describing
landed work in front of a whitespace change is worse than dropping it, and the ratchet's
shape rule (comments above the first data row) is unaffected either way — the file was
edited by hand, never with `--write-baseline`.

### Gates

Worktree detached from `origin/<branch>`, outside the repo tree.

* `just check fast` — **289/290**. Only `xrce-vendored-versions`, which fails
  identically on a plain `origin/main` control worktree. Pre-existing and environmental.
* `just check api-parity`, `api-parity-ledger`, `ledger-orphan-refs`, `doc-refs`,
  `issue-ids`, `prose-issue-refs` — all OK.
* `python3 scripts/check-doc-commit-citations.py` — OK, 207 citations across 914 live
  files resolve. The two "may no longer be needed" entries (`d97c9c606`, `e3786749a`) are
  reported identically on the `origin/main` control, so they are not this branch's.
* `just check cpp` — all C++ checks passed, `failed_create_aborts: OK`.
* `cargo +nightly fmt --check --all` — clean.
* `cargo check -p nros-node --all-targets --features std` — clean.
* All 17 ledger shards re-parsed with an `object_pairs_hook` counter: **zero** duplicate
  top-level keys.

jerry73204:
Correction to the comment above, resolution 4: the row counts are **62 changed, 54 byte-identical** to `main`, not 64/56 — re-counted from the per-shard tally (action 9, exec 3, init 12, log 3, node 5, other 3, param 6, pubsub 3, qos 11, service 1, timer 5, types 1). The 8 divergent rows and the evidence for each are unchanged.

PR #779, opened 2026-09-08
fix(#1233): four road-gaps, closed by delivery and not by an excuse

jerry73204:
Stacked content note: carries 768's commit (`fix(phase-412 #7)`) because this
work builds on its registry. Based on `main`, not on 768's branch — the queue
drops the carried commit by patch-id once 768 merges.

phase-412 #7 recorded four road-gaps with no reason anywhere in the tree:
`NROS_DERIVED_EXECUTOR_MAX_NODES` and `NROS_DERIVED_SUBSCRIPTION_BUFFER_SIZE`
took the resolver road only. Both now travel all three. Neither gap turned
out to have a reason worth writing down instead.

**The node table.** phase-412 W1 withheld it because under-counting HALTS the
board. It does — as `NodeError::NodeTableFull`, a named failure at
registration — and that is a property of the FAILURE, not of the road the
number arrived on; the same count was admitted to the resolver road on
exactly that ground. Delivered unfloored (the floor belongs to the consumer
that names the knob, issues 1015/1033).

Found while wiring it: `nros-rmw-zenoh/build.rs` is a **second reader** of
the knob, and its own comment already said "keep them in sync" with
`nros-node` with nothing enforcing it. `nros-node` consults `$DOTCONFIG`;
that crate has no Kconfig rung for the name, so on a Zephyr Rust image
`CONFIG_NROS_EXECUTOR_MAX_NODES` reached one crate and not the other. Both
take the declared rung now; the zenoh side keeps its own `min(1)` because it
indexes the pool unconditionally (0827) and `nros-node` does not.

**The take buffer, and why 1122 left it behind.** Its guard differs.
`_nros_payload_facts_env` needs `PAYLOAD_STATUS derived` AND
`BASIS subscribed`; this knob publishes under `MESSAGE_BOUNDS_STATUS derived`
over the CLOSURE, because the producer says so at the site — "BASIS
`closure`, always. Narrowing this one is the under-derivation." `RX_BUF` is
one global size for every entity and `DEFAULT_TX_BUF` aliases it, so a type
the image only publishes still has to fit. A row in the existing carrier
would have refused every image whose join answered on the closure basis,
which for this fact is all of them. Hence its own carrier
(`_nros_take_buffer_env`), and on the sidecar road a third key list
(`DERIVED_CLOSURE_ENV_KEYS`) fed by a new `nros_cli_core::leaf_take_buffer`
that derives over the leaf's `generated/` tree and refuses on any unbounded
member.

Measured, host build:

| env | `DEFAULT_RX_BUF_SIZE` | `MAX_NODES` |
| --- | --- | --- |
| neither | 1024 | 4 |
| `NROS_DECLARED_*` = 880 / 2 | 880 | 2 |
| declared + named 333 / 6 | 333 | 6 |

`env > Kconfig/board > derived > crate default` holds on both.

`check-declared-fact-carriers`: 12 declared facts produced, consumed and
watched; 14 published facts with a stated disposition on all three roads;
**zero OPEN**. New watch lines are spelled literally, since
`env_usize_declared`'s formatted line is invisible to grep.

Ran: `just ci gate` (green — 1225 unit tests, 2 capability skips for a host
port), `just check fast`, `just check declared-fact-carriers`, `just check
knob-delivery`, `just check config-knob-census`.

Issue: docs/issues/archived/1233-derived-fact-road-gaps.md (resolved)


jerry73204:
Taken out of the merge queue on purpose: this branch carries all 3 commits of #768, which is also queued, and the duplicate made both entries UNMERGEABLE. Once #768 merges, this gets rebased onto main (the 3 shared commits drop by patch-id) and re-armed. #820 stays held behind this PR, since it cites issue 1233.

PR #778, opened 2026-09-08
docs(phase-438 W5, #1187): NROS_CPP_STD is a PORTING surface — say which one a consumer is on

jerry73204:
## What

phase-438 W5 — documents `NROS_CPP_STD` as what W2 made it: an opt-in **porting**
surface, not a capability the toolchain hands you.

Before this phase, the six `NROS_CPP_HAS_*` capability macros were DISCOVERED
from the include path with `#elif defined(__has_include)`, so any hosted
compiler gave a consumer the std-flavoured C++ API whether they asked for it or
not. W2 deleted that arm. Nothing in the book or the reference docs told a
consumer which surface they were on, and after W2 that question has an answer
they need to be able to reach without opening a header.

## The framing, since it is not the obvious one

**The split is *ported rclcpp code* versus *code written for nano-ros*. It is
NOT host versus embedded.** `NROS_CPP_STD` exists so an upstream ROS 2 source
file compiles unmodified — that is the whole reason to ask for it — and every
real nano-ros program in this tree, native/host builds included, is on the
freestanding surface already. Measured: **1** in-tree user of the hosted
`shared_ptr`-returning `create_*` (a compile test) against **27** call sites of
the freestanding out-ref form in `examples/` alone.

The one shipping build path that sets the flag is the px4 uORB bridge module,
plus `cmake/compat/NrosRclcppCompat.cmake`, which W2 made set it for every
target using the rclcpp compat shim — so a ported project usually needs nothing.

## Files

| File | What it gains |
| --- | --- |
| `book/src/reference/cpp-api.md` | A `Two surfaces` section: what each surface is, a **"do I need this flag?" table** with one row per consumer shape, how to turn it on, that the compat CMake already does, and the three-lane probe table as the reason nothing detects it. |
| `book/src/getting-started/porting-a-cpp-node.md` | The same at the point a porter meets it — the one compile definition, and that they will not type it. |
| `docs/reference/c-api-cmake.md` | Reference depth, in that document's voice, beside the other CMake knobs; names both in-tree paths that set the definition and the measured cost of the discovery it replaced. |
| `docs/guides/cpp-api.md` | "Optional std Mode" re-cut. Its opening — *"for any toolchain with a C++ standard library"* — was exactly the host-versus-embedded framing this phase disproves. |
| `changelog.d/1187.breaking.md` | New. See below. |
| `docs/design/0018-cpp-api-design.md` | See below. |
[...]

PR #784, opened 2026-09-08
feat(phase-438 W4+W3): `rclcpp::Node` is ONE class on every target — the std surface is additive methods, not a second shape

jerry73204:
**Base: `main`.** Retargeted and rebased since it was opened, so this branch now also carries phase-438 W1 and W2 (previously PR #770) — the W4 work sits directly on top of them and cannot be separated from them, since `nros.hpp`'s guard structure is what W2 changed. Sibling of the W5 docs PR (#778), not stacked on it.

phase-438 **W4** and **W3**, in that order, because W3's outcome is decided by W4. W4 is also **phase-427 W1, relocated**, so this unblocks phase-427 W2–W7.

## W4 — the hosted half becomes additive

`rclcpp::Node` was two different SHAPES of one class. With the capability macros on it had a `std::enable_shared_from_this<Node>` base, a `std::vector<std::shared_ptr<void>>` member and a full set of `shared_ptr`-returning factories; with them off it did not exist at all, because the whole class sat inside `#if defined(NROS_CPP_HAS_SHARED_PTR) && defined(NROS_CPP_HAS_STD_STRING) && defined(NROS_CPP_HAS_STD_VECTOR) && defined(NROS_CPP_HAS_STD_FUNCTION)`.

That was tolerable while those macros were DISCOVERED from the include path, because "hosted" and "wants the porting surface" moved together. W2 made the surface a consumer REQUEST, so one image can now legitimately hold a TU that asked and a TU that did not — which `examples/px4/cpp/bridge/.../CMakeLists.txt:123` does on purpose. Two shapes of one class in one image is issue 0135 / 0460: they link, and each writes the object through its own layout.

What changed:

- **The layout is unconditional.** The base class and `owned_entities_` moved into a `detail::NodeHosted` box reached through one unconditional `void* hosted_`, allocated lazily — a freestanding node, and a hosted node that never touches the std surface, allocates nothing.
- **`rclcpp::NodeOptions` became unconditional too**, which it had to be: the node holds one by value, so a class present in one configuration and absent in the other would itself have been a conditional member. It has no data members, so only `arguments()` stays gated and `sizeof(NodeOptions)` is 1 in both arms.
- **The factories became overloads.** The out-ref forms mirroring `nros::Node` (`create_publisher(out, "topic", qos)` and siblings) are declared on every target; the `shared_ptr`-returning and `std::string`-keyed forms sit beside them under the porting surface.
- **The box's deleter is carried in the object**, as a second unconditional `void (*)(void*)`. An `#ifdef` inside `~Node()` would have satisfied the `sizeof` rule and still given one inline symbol two bodies across TUs that are permitted to disagree.

### `shared_from_this` costs one source line, as RFC-0089 said it would

The base is gone; `shared_from_this()` survives as a hosted-only method over a `std::weak_ptr<Node>` in the box, populated by an explicit `bind_shared`:

```cpp
auto node = std::make_shared<MyNode>();
node->bind_shared(node);      // <-- new
node->post_init();
```

The two in-tree templates that use upstream's two-phase idiom gained one line each. A file that forgets it still COMPILES and aborts at the call with the migration text rather than returning an empty pointer to dereference — weaker than the "mechanical edit" standard, admissible only because it is loud. `changelog.d/1204.breaking.md` carries this for users.

### Acceptance — measured, both configurations

```
sizeof(rclcpp::Node)
  hosted       -std=c++17 -DNROS_CPP_STD=1                       3728
  freestanding -std=c++14 -ffreestanding -nostdinc++ (ThreadX)   3728
```

It was 3752 hosted and ABSENT freestanding: −16 for the base, −24 for the vector, +16 for the two pointers.

**The arms have teeth on this type, mutation-tested rather than assumed.** An `#ifdef NROS_CPP_HAS_SHARED_PTR`-gated `double` injected next to `hosted_` reports `3728 without, 3736 with` on the new arm and `3736 vs 3728` on the freestanding one. That is the exact mutation issue 1204 measured as *uncatchable* — and the reason it was uncatchable, a guard naming the very macro the `#ifdef` tests, is what this PR removes. **Issue 1204 is resolved and archived.**

## W3 — off the hosted-only list, and the honest reason it was there

`hosted_only_reason()` is now EMPTY, and the two-way ratchet is what keeps it that way. The removed reason string was wrong twice, and both corrections are recorded in its place:

- **Name the CONSTRUCT, never a line.** It said "at nros.hpp:447". 447 was the preceding `} // namespace rclcpp`; the `#if` was at 454, and the line moved twice more inside this phase alone.
- **Say what is MISSING, not what is guarded.** The real reason the class could not be measured freestanding was the OWNERSHIP SHAPE — every factory returned `std::shared_ptr` and this tree has no freestanding equivalent — so the guard was a consequence. A symptom-shaped reason invites the wrong fix (delete the guard), which is what this phase's first draft of W3 proposed before a compile measurement refuted it.

The gate also gained the **third arm** its own `HOSTED_FLAGS` comment had asked for: hosted `-std=c++17` **without** `-DNROS_CPP_STD`. It isolates the flag from the toolchain, which the freestanding arm cannot do alone because that arm varies `-std`, `-nostdinc++` and the shim at once.

## What this does NOT do

The freestanding `rclcpp::Node` is layout-identical, constructible, derivable, and its out-ref factories work. It is **not** a full freestanding port. `Node::SharedPtr`, `rclcpp::spin(node)` / `spin_some(node)` / `spin_until_future_complete(node, …)`, `TimerBase` and the `shared_ptr` factories stay on the porting surface, because their signatures are spelled in an ownership type this tree does not have freestanding — giving `shared_ptr` a freestanding equivalent is explicitly out of scope for phase-438. A freestanding program drives the same executor through the unconditional `nros::spin()` / `nros::spin_once()`.

### The additive half needed its own check

`check-cpp-capability-layout` measures the LAYOUT half. Nothing measured the METHOD half, and nothing could have: the per-header `-fsyntax-only` loop only PARSES the templates, and every other ported-surface probe compiles with `-DNROS_CPP_STD=1` — the configuration in which the new out-ref overloads are the least interesting. So they were declared, never compiled, and green either way.

`packages/api/nros-cpp/tests/compile/rclcpp_node_freestanding_surface.cpp` constructs a `rclcpp::Node` by value, instantiates every unconditional factory, walks the `const char*`-keyed parameter forwarders, derives from the class and static-asserts that neither the base nor the deriver is polymorphic (RFC-0089 correction 1, which had no in-tree home). It names nothing from the porting surface, so it fails exactly when something moves OFF the unconditional half. The lane compiles it hosted-without-the-flag and `-nostdinc++`.

Mutation-tested both directions — moving the out-ref `create_publisher` back behind `#ifdef NROS_CPP_HAS_SHARED_PTR`:

```
probe:       error: 'class rclcpp::Node' has no member named 'create_publisher'
layout gate: exit 0 — still GREEN
```

The second line is the point: a deleted method moves no layout, so the two checks are complementary and neither alone states W4's contract.

## Two things found on the way

**Issue 1239 (filed, fixed, archived).** The two rclcpp compat entry points are supposed to be equivalent and were not. `cmake/compat/stubs/Findrclcpp.cmake` — the one serving a stock `find_package(rclcpp)` + `ament_target_dependencies` — still force-included `nros/rclcpp_compat.hpp`, deleted in phase-417 stage 6 step A, so every C++ TU compiled against it died with `fatal error: nros/rclcpp_compat.hpp: No such file or directory` before reading a line of its own source. It also never set `NROS_CPP_STD`, which W2 added to `_nros_compat_apply_force_includes` and only there. Measured at the base commit with this diff stashed, on the `shadowing` fixture: `expected class-name before '{' token`, `'TimerBase' is not a member of 'rclcpp'`, `'spin' is not a member of 'rclcpp'`. Fixing it is what let the ported templates be verified at all.

**A red fast-line gate on `main`** (second commit). `AGENTS.md:535` cited `bd4b087c1` — one of #629's own commits — inside the paragraph explaining that rebasing a stacked branch strands it. #629 was rebased twice, the hash stopped resolving, `check-doc-commit-citations` went red, and the `pre-push` hook then blocks every push in the repository. Cited by description instead; not baselined, because the sentence reads correctly without the hash.

## Verification

| what | result |
| --- | --- |
| `just check cpp` | green — includes the layout gate + selftest and BOTH `-nostdinc++` header-parse arms (ThreadX shim and Zephyr minimal libcpp) |
| `just check fast` | green — 281 gates ran, 1 skipped (`ivc-fsp`, unprovisioned), 0 failed |
| per-header standalone parse, `arm-none-eabi-g++ 13.2-nros1`, `-std=c++14 -ffreestanding -fno-exceptions -fno-rtti` | **pass=47 fail=0**, unchanged |
| FreeRTOS C++ cross build (issue 1187's own acceptance) | exit 0, six ARM EABI5 ELF binaries |
| porting compile tests under `packages/api/nros-cpp/tests/compile/` | compile (via `just check cpp`), including the new `rclcpp_node_freestanding_surface.cpp` in both arms |
| the five ported `cmake-fixture` rows | all build: `cpp_port_minimal_publisher`, `cpp_port_rclcpp_compat_smoke`, `cpp_port_topic_state_monitor`, `local_msg_pkg`, `shadowing` |

Tier run: `just check cpp` + `just check fast` + the two cross/fixture builds above. Not a full `just ci`.

### DIFF
--- a/.config/cpp-capability-layout-baseline.txt
+++ b/.config/cpp-capability-layout-baseline.txt
@@ -17,6 +17,13 @@
 #                          was supposed to exist there" is a real defect wearing
 #                          the same clothes; issue 1204 is the version of this
 #                          gate that inferred it and passed on absence.
+#   std-only <subject>     the subject does not EXIST hosted with
+#                          `-DNROS_CPP_STD` withheld (phase-438 W4's arm). Same
+#                          argument as `hosted-only`, on a different axis, and
+#                          a SEPARATE kind on purpose: one line must not excuse
+#                          two arms, or a type that starts vanishing on the
+#                          second one is absorbed by an exemption written for
+#                          the first.
 #
 # Subject spellings are the DERIVED ones -- `scripts/check/
 # cpp_capability_subjects.py` prints the list. A line naming a subject the
@@ -62,6 +69,26 @@
 # correct outcome for a type whose whole surface is hosted, and none of them is
 # a layout that two TUs can disagree about.
 
+# --- the `std-only` entries -------------------------------------------------
+#
+# The same three types on the other axis, and for a DIFFERENT reason -- which
+# is why the kinds are separate rather than one line excusing both arms.
+# `hosted-only` is about a toolchain that cannot reach the standard library;
+# `std-only` is about a consumer that did not ASK for the porting surface.
+# Since phase-438 W2 the `NROS_CPP_HAS_*` macros are defined by `NROS_CPP_STD`
+# and by nothing else, so a hosted TU that withholds the flag has none of them
+# and the same three guards close for it.
+#
+# `Rate` / `WallRate` were NOT on this arm when W4 first ran it, and the reason
+# is worth recording. `NROS_CPP_HAS_STD_CHRONO` had a SECOND definition in
+# `node.hpp`, still carrying the `__STDC_HOSTED__ && __has_include(<chrono>)`
+# discovery that W2 deleted from the shared site -- so a hosted TU kept the
+# macro from the copy W2 never reached, and the two `<chrono>`-guarded types
+# went on existing without the flag. phase-438 W1 as it landed on `main`
+# (`cf4b0048a`) folds that copy into `std_detect.hpp`, which is what lets this
+# arm measure the flag rather than the include path. Two definitions of one
+# capability macro is issue 0135's shape; one site is the fix.
+
 diverges ::nros::ComponentNode
 diverges ::nros::GuardCondition
 diverges ::nros::Timer
@@ -70,3 +97,6 @@ diverges ::rclcpp::TimerBase
 hosted-only ::rclcpp::NodeOptions
 hosted-only ::rclcpp::Rate
 hosted-only ::rclcpp::WallRate
+std-only ::rclcpp::NodeOptions
+std-only ::rclcpp::Rate
+std-only ::rclcpp::WallRate
--- a/.config/cpp-freestanding-includes-baseline.txt
+++ b/.config/cpp-freestanding-includes-baseline.txt
@@ -1,32 +1,22 @@
-# Hosted STL includes reached with `NROS_CPP_STD` undefined, in the `#elif
-# defined(__has_include)` arm of nros-cpp's capability blocks.
+# Hosted STL includes reached with `NROS_CPP_STD` undefined, in nros-cpp's
+# capability blocks.
 #
 # A RATCHET, not an allowlist. It may only SHRINK: a line here that no longer
 # matches a violation is a FAILURE, so the debt cannot go stale, and a new
 # violation not listed here is a failure too, so it cannot grow.
 #
-# EMPTY as of issue 1240 (2026-09-09). All fourteen entries are paid.
+# EMPTY as of issue 1240 (2026-09-09). All fourteen entries are paid, and
+# phase-438 W1+W2 then removed the shape that produced them.
 #
-# WHAT PAID THEM, AND WHY IT IS NOT WHAT WAS PLANNED
+# WHAT PAID THEM, IN TWO STEPS THAT DISAGREED ABOUT THE INSTRUMENT
 #
-# The plan recorded here was phase-438 W2: "delete the `#elif` arm, so the std
-# surface is reachable only through `NROS_CPP_STD`." That would have been the
-# wrong fix, and `nros.hpp` already says why in its `<chrono>` note — NOTHING
-# THAT SHIPS DEFINES `NROS_CPP_STD`. It appears in no cmake module, no
-# toolchain file, no build.rs and no Kconfig, only in docs and in `check.just`'s
-# own probe TUs. Deleting the arm would not have tightened the gate on
-# freestanding targets; it would have removed `std::string`-taking `get_logger`,
-# `std::shared_ptr` aliases and the `<vector>` options surface from EVERY
-# configuration, hosted ones included. `nros.hpp` made exactly that mistake for
-# `<chrono>` on 2026-09-05 and corrected it two days later.
-#
-# What paid them instead is the corrected predicate from that same note, applied
-# to all fourteen sites at once (issue 1240):
+# Issue 1240 paid all fourteen by CORRECTING the probe in place, at all fourteen
+# sites at once:
 #
 #   #if defined(NROS_CPP_STD) || (defined(__STDC_HOSTED__) && __STDC_HOSTED__ \
 #       && __has_include(<string>))
 #
-# Neither probe alone is sufficient and both are necessary:
+# Neither probe alone is sufficient and both were necessary:
 #
 #   * `__has_include` alone answers "does the header FILE exist", which is TRUE
 #     under `-ffreestanding` for a libstdc++ whose `<string>` begins
@@ -37,15 +27,38 @@
 #     hosted compiler run `-nostdinc++` against Zephyr's minimal libcpp reads
 #     HOSTED and still has no `<string>`.
 #
-# The conjunction covers both lanes; `NROS_CPP_STD` stays as the explicit
-# override for a consumer who knows better than the probes.
+# phase-438 W1 then moved that one predicate to its single definition site,
+# `packages/api/nros-cpp/include/nros/std_detect.hpp`, and W2 DELETED the
+# discovery disjunct there. 1240's objection to deleting it — "nothing that
+# ships defines `NROS_CPP_STD`, so the arm is what gives the hosted build
+# `std::string`-taking `get_logger`, the `SharedPtr` aliases and the `<vector>`
+# options surface" — was correct about the tree as it stood, and W2 is what
+# makes it stop being true: five consumers now make the request out loud (the
+# rclcpp compat shim, the phase-417 ported-surface probes, the four refusal
+# probes, the `cpp_compat_snippets` compile-check arm, and
+# `check-cpp-capability-layout`'s hosted arm). A better probe was still a probe;
+# the macro answers which SURFACE a consumer asked for, which no probe can see.
 #
-# The walker was tightened in the same change so it cannot be satisfied by the
-# TOKEN alone: an `#if`/`#elif` naming NROS_CPP_STD that also reaches
-# `__has_include` without `__STDC_HOSTED__` is NOT a std region. Without that,
-# the broken shape and the corrected one score identically, which is the 0196
-# rule — a gate whose reach is narrower than the rule it enforces.
+# Measured on the toolchain that could not build at all before (arm-none-eabi
+# 13.2, `-std=c++14 -ffreestanding -fno-exceptions -fno-rtti`), per-header
+# standalone parse:
+#
+#     before W2   pass=26 fail=20
+#     after W2    pass=47 fail=0
+#
+# The walker was tightened by 1240 so it cannot be satisfied by the TOKEN alone:
+# an `#if`/`#elif` naming NROS_CPP_STD that also reaches `__has_include` without
+# `__STDC_HOSTED__` is NOT a std region. Without that, the broken shape and the
+# corrected one score identically, which is the 0196 rule — a gate whose reach
+# is narrower than the rule it enforces. After W2 no capability block reaches
+# `__has_include` at all, so the tightened walker has nothing left to reject
+# here; it stays because the rule outlives this file's emptiness.
 #
 # Keyed on FILE and HEADER, not on a line number, so an entry survives edits
 # above it and so a second `<memory>` appearing in a listed file is still
 # caught. Add a line here only with a written reason and a tracked issue id.
+#
+# The file stays rather than being deleted: the gate requires it, and the next
+# genuine hosted-include debt needs somewhere to be recorded with its reason
+# instead of being argued about in a review. An entry that stops offending
+# FAILS, so an empty file cannot quietly refill.
--- a/changelog.d/1204.fix.md
+++ b/changelog.d/1204.fix.md
@@ -1,0 +1,30 @@
+`check-cpp-capability-layout` can now measure the type it was filed over.
+Issue 1204 recorded that the gate forced capability macros ON against a
+baseline where they were already on, so it compared the baseline against itself
+seven times, and that `rclcpp::Node` could not be measured any other way
+because the whole class sat inside `#if defined(NROS_CPP_HAS_SHARED_PTR) &&
+...` — an `#ifdef` on that macro *inside* that block is tautological.
+
+phase-427 W1-W3/W5 removed the guard: `rclcpp::Node` is `::nros::Node`, one
+class declared unconditionally, with the hosted state behind a single lazily
+allocated `detail::NodeHostedBase*` and the `shared_ptr`-returning factories as
+additive overloads beside unconditional out-ref forms. phase-438 W3/W4 is what
+MEASURES that outcome: the gate gains a third arm — hosted `-std=c++17`
+**without** `-DNROS_CPP_STD`, which isolates the opt-in flag from the toolchain,
+because the freestanding arm varies `-std`, `-nostdinc++` and the shim all at
+once — and `rclcpp::Node`'s exemption is gone. The arm covers all 90 subjects
+issue 1225's derivation produces, and the derivation asks for the flag itself,
+so a subject that only exists in the porting surface cannot slip out of the
+gate; the one that does, `rclcpp::NodeOptions`, is declared `std-only` in
+`.config/cpp-capability-layout-baseline.txt`.
+
+Measured: `sizeof(rclcpp::Node)` is **200** in all three arms — hosted with the
+flag, hosted without it, and `-std=c++14 -ffreestanding -nostdinc++` against the
+ThreadX shim.
+
+A second check states the half `sizeof` cannot: the `cpp` lane now compiles
+`packages/api/nros-cpp/tests/compile/rclcpp_node_freestanding_surface.cpp`,
+which instantiates every unconditional factory and derives from the class, in
+both no-flag arms. Deleting a method moves no layout, so the capability gate
+stays green on that mutation while the probe fails — the two are complementary,
+and neither alone states the contract.
--- a/changelog.d/1239.fix.md
+++ b/changelog.d/1239.fix.md
@@ -1,0 +1,11 @@
+The `find_package(rclcpp)` compat entry point works again, and now matches the
+`ament_auto_*` one. `cmake/compat/stubs/Findrclcpp.cmake` still force-included
+`nros/rclcpp_compat.hpp`, a header deleted in phase-417 when the `rclcpp::`
+spellings moved into `<nros/nros.hpp>` itself, so every C++ TU compiled against
+the stub died with `fatal error: nros/rclcpp_compat.hpp: No such file or
+directory` before reading a line of its own source. It also never set
+`NROS_CPP_STD`, which phase-438 W2 made the way to ask for the std-flavoured
+`rclcpp::` surface — `ament_target_dependencies` links the stub's INTERFACE
+target without going through the per-target helper that does set it. Both are
+fixed; a stock `find_package(rclcpp)` + `ament_target_dependencies` package
+compiles again.
--- a/cmake/compat/NrosRclcppCompat.cmake
+++ b/cmake/compat/NrosRclcppCompat.cmake
@@ -114,6 +114,24 @@ function(_nros_compat_apply_force_includes target)
     # header) resolve to the nano-ros compat shims.
     target_include_directories(${target} PRIVATE
         "${_nros_compat_dir}/include")
+    # phase-438 W2 — ASK for the std surface. A file reaching
+    # `<rclcpp/rclcpp.hpp>` subclasses `rclcpp::Node`, whose signatures are
+    # spelled in `std::shared_ptr` / `std::string` / `std::vector`; that surface
+    # is now reachable only through `NROS_CPP_STD`, never discovered from the
+    # include path (issue 1187 — `__has_include(<string>)` is TRUE on
+    # arm-none-eabi `-ffreestanding`, where including it is a hard `#error`).
+    #
+    # This target is exactly the right place and nowhere else is: the shim is
+    # applied per-target and is deliberately NOT auto-applied on Zephyr (see
+    # below), so the opt-in follows the porting surface instead of leaking to
+    # images that use `nros::Node` directly.
+    #
+    # Until now this function set NO compile definition at all, and the whole
+    # ported-code path worked only because a hosted compiler happened to find
+    # libstdc++. `scripts/api-parity.py` even documented the opposite —
+    # "that is the flavour the compat CMake path builds under" — of a path that
+    # defined nothing. It is true from here.
+    target_compile_definitions(${target} PRIVATE NROS_CPP_STD=1)
 endfunction()
 
 # Phase 210.E.3.c — Zephyr context: do NOT auto-apply force-include of
--- a/cmake/compat/stubs/Findrclcpp.cmake
+++ b/cmake/compat/stubs/Findrclcpp.cmake
@@ -20,9 +20,30 @@ if(NOT TARGET rclcpp::rclcpp)
     # NrosRclcppCompat lives at `../include/` relative to this stub dir.
     get_filename_component(_nros_compat_inc_dir "${CMAKE_CURRENT_LIST_DIR}/../include" ABSOLUTE)
     target_include_directories(rclcpp::rclcpp INTERFACE "${_nros_compat_inc_dir}")
+    # Issue 1239 — `nros/rclcpp_compat.hpp` was force-included here too, and
+    # `f35f0b878` (phase-417 stage 6 step A) DELETED that header when the
+    # `rclcpp::` spellings moved into `<nros/nros.hpp>` itself. A force-include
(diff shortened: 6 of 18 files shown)


jerry73204:
Retargeted from `feat/phase-438-w1-w2-std-detect` to `main`.

**Why:** it stacked on #770, which was itself unstacked and rebased onto `main` earlier, so this branch pointed at history that no longer existed. A stacked PR also cannot arm auto-merge at all — auto-merge is a property of the BASE branch, and a feature branch has neither the ruleset nor the queue (`Protected branch rules not configured for this branch`).

**5 → 4 commits.** One dropped: `fix(#1223, phase-438 W0)`, verified present on `main` as `e5adb2d94` — it landed with #763.

**The two carried W1/W2 commits are provably the same patch as #770's**, so they drop by patch-id when #770 lands rather than conflicting. That is the claim the whole unstacking argument rests on, so it is measured rather than asserted:

```
1895d4c24 == 4cd9f515c  patch-id 6998f0dda7a3cf598b673071586af33e6a6825cd
d77d870cd == 6542a31e1  patch-id d44be81413a28bf181b0308e47d28077f47ed7f9
```

**No-loss check:** every commit subject on the pre-push tip is present on the new one, except the one drop above.

**Verified:** `just check cpp` — `All C++ checks passed!`; `doc-refs`, `doc-recipe-refs`, `issue-ids`, `prose-issue-refs` OK; `check-doc-commit-citations` OK (193 citations across 891 live files resolve).

jerry73204:
## Situation change: phase-427 W1 landed on `main` while this PR was in review, and it supersedes W4

`f4c5ea765 feat(#427 W1-W3,W5): one node type, named rclcpp::Node — and it fits freestanding` is now on `main`. That is the same work item this PR carries as phase-438 W4 (the phase doc says so explicitly: *"This IS phase-427 W1"*). It reaches the same conclusion by the same route — one merged `rclcpp::Node`, an unconditional `void* hosted_`, hosted state in a `detail::NodeHosted` box, a destructor that goes through a stored function pointer — and it also empties `hosted_only_reason()`, which is this PR's W3.

It differs in one design decision, and `main`'s is the better one for consumers: `shared_from_this()` there returns a `shared_ptr` ALIASING `this` with an empty owner, so no `bind_shared` call and no source edit in the two ported templates. This PR's version needs one added line per ported file. `main`'s weakens lifetime instead (the returned pointer does not extend the node's life) and documents that; mine keeps ownership and costs an edit. Not worth relitigating — `main`'s has landed and the templates are untouched there.

**So the W4 and W3 commits here should not be rebased onto `main`.** Rebasing `99d713101` produces conflicts in `nros.hpp` and `check-cpp-capability-layout.sh` whose only honest resolution is "take `main`'s", because `main` already did the work.

### What in this branch is NOT superseded

1. **phase-438 W1 + W2** (`d77d870cd`, `1895d4c24`) — `std_detect.hpp` does not exist on `main`, so the `#elif __has_include` arm is still live there and issue 1187 is still open in effect. These rebase cleanly apart from one trivial include-placement conflict.
2. **The gate's third arm** — hosted `-std=c++17` **without** `-DNROS_CPP_STD`. `main` has `HOSTED_FLAGS=(-std=c++17)` and no such arm, which is correct only while the macros are discovered from the include path. The moment W2 lands, that baseline stops measuring the std surface at all, and the arm that isolates the flag from the toolchain becomes the thing that keeps `sizeof(rclcpp::Node)` honest across the px4 split.
3. **Issue 1239** — `cmake/compat/stubs/Findrclcpp.cmake` still force-includes `nros/rclcpp_compat.hpp` on `main` today (deleted in phase-417), so every stock `find_package(rclcpp)` + `ament_target_dependencies` consumer fails at its first TU. Independent of everything above. Its second half (setting `NROS_CPP_STD` on the stub's INTERFACE target) is needed once W2 lands.
4. **The dead commit citation in `AGENTS.md`** — still present on `main`, so `check-doc-commit-citations` is red there, and that gate is on the `pre-push` hook, so it blocks every push in the repository by every contributor.
5. **`rclcpp_node_freestanding_surface.cpp`** — instantiates the unconditional node surface with no `NROS_CPP_STD`, in two arms, mutation-tested. It was written against this branch's node; whether it compiles against `main`'s merged node is unverified and needs re-measuring before it is worth keeping.

### Suggested disposition

Retarget this PR's content to items 1–4 (dropping the W4/W3 commits), or close it in favour of #770 plus a small follow-up carrying 2–4. I have not force-rebased anything: the branch is still at `fc86bf349`, which is what CI went green on. Handing the decision back rather than guessing.

jerry73204:
## Rebased onto `main` — part by part, with the measurement behind each call

`main` landed phase-427 W1-W3/W5 (#755, "one node type, named `rclcpp::Node`") after this
branch was cut. That work and this PR's headline commit set out to do **the same thing**, so
the rebase was a supersession audit rather than a conflict resolution. `fc86bf349` → `64632639d`,
5 commits → 6.

### Superseded, and the evidence for each

**The whole `nros.hpp` half of `99d713101` (393 insertions).** It rebuilds a shim
`class Node` inside `nros.hpp` — every one of the 6 conflicts' branch sides is a member of it
(`node_.get_name()` delegation, `hosted_state_()`, `bind_shared`, the `std::string` parameter
and `create_client` overloads). `main` has **no `class Node` in `nros.hpp` at all**: the merge
moved it to `node.hpp` as one unconditional `class Node` with `using Node = ::nros::Node`, and
`nros.hpp` now holds only its hosted method *bodies* under `NROS_CPP_NODE_HOSTED`. Every member
the branch side re-declared exists on `main` (checked by name: the `std::string` `declare_parameter`
/ `get_parameter` / `set_parameter` / `has_parameter` at `node.hpp:658-671`, `create_service` /
`create_client` at `node.hpp:591-630`). The branch side is also *behind* `main` in one place —
it returns `std::shared_ptr<TimerBase>`, the `TimerBase::SharedPtr` that phase-430 W7 deleted —
so re-applying it would have regressed that.

**`bind_shared` + `NROS_RCLCPP_REFUSE_UNBOUND_SHARED_FROM_THIS` (log.hpp) + the two template
edits.** Both designs drop the `std::enable_shared_from_this<Node>` base for the same stated
reason (a base changes layout in configurations that may legitimately disagree). They differ in
the replacement: this branch requires an explicit `node->bind_shared(node)` per ported file and
aborts loudly when it is omitted; `main` returns a pointer that **aliases `this` with an empty
owner** (`node.hpp:541-547`), needing no source edit. `main` has no `bind_shared` member, so
carrying the template edits would not have compiled. `main`'s design is landed and shipped, and
adopting the branch's would be both a regression and a new breaking change — so the
`1204.breaking.md` fragment is gone and `1204.fix.md` replaces it. Nothing breaking survives.

**`options.hpp` making `rclcpp::NodeOptions` unconditional.** Its stated justification is
"the node holds a `NodeOptions` BY VALUE". That is false on `main`: the object lives in
`detail::NodeHosted` (`node.hpp:272`), which is itself inside `#ifdef NROS_CPP_NODE_HOSTED` and
lazily allocated, so no conditional member exists. `NodeOptions` is also **not on
`check-cpp-capability-layout`'s `TYPES` list**, so the new arm does not require it either.
Reverted to `main`'s whole-class gate.

**`hosted_only_reason()`.** Both sides empty it; `main`'s comment is fuller and records that
phase-427's ratchet is what forced the entry off. Took `main`'s.

### Carried, because `main` does not have it

- **`check-cpp-capability-layout.sh` ARM 2** — hosted `-std=c++17` with `-DNROS_CPP_STD`
  *withheld*. `main`'s gate has two arms; this is the third, and it is the one that isolates the
  flag from the toolchain. Not measurable before W2. **Re-measured against `main`'s class, not
  copied**: `sizeof(rclcpp::Node)` is **200** in all three arms (the branch's `3728` was its own
  implementation). Mutation-tested on `main`'s shape — an `#ifdef NROS_CPP_HAS_SHARED_PTR`-gated
  `double` beside `clock_` in a throwaway copy of the include tree gives `200 without, 208 with`
  on the new arm and `208 vs 200` freestanding.
- **`Findrclcpp.cmake` (issue 1239)**, both halves. Verified live on `main`:
  `nros/rclcpp_compat.hpp` is **absent from `origin/main`'s tree** (only
  `rclcpp_components_compat.hpp` exists), so the force-include kills every TU compiled against
  the stub. The `NROS_CPP_STD=1` half pairs it with W2.
- **The W4 compile probe**, adapted rather than dropped. It did **not** compile against `main`:
  `initialized()`, `get_node_options()`, `parameters()` and the *whole* parameter facade — the
  `const char*`-keyed forms included — are hosted-only on `main`, because the store and options
  live in the hosted box. It now uses `ok()` and names no parameter forwarder, and the comment
  records where the line actually falls. Both arms compile; mutation-tested (out-ref
  `create_publisher` moved back behind a capability macro → `no matching function for call to
  'nros::Node::create_publisher'`, while the layout gate stays green — which is why both checks exist).
- **W1 and W2** (#770's commits) — not superseded. `main`'s `nros.hpp` still **uses four
  `NROS_CPP_HAS_*` macros and defines zero**, which is exactly the include-order defect W1 fixes.
- **The AGENTS.md citation fix** — independent, and `check-doc-commit-citations` is green.

### One new commit: `fix(check): two phase-427/430 probes ask for the hosted surface W2 made a request`

`just check cpp` was red once W1+W2 sit under the node merge, and neither side is wrong.
`one_node_type.cpp` (phase-427) and `timer_base_is_a_name_not_a_hierarchy.cpp` (phase-430 W7)
landed on `main` **after #770 was cut**, so W2 never saw them and they compile with no flag.
Measured: 18 → 0 and 13 → 0 errors with `-DNROS_CPP_STD=1`; both are **0 without the flag on
plain `origin/main`**, where the macros are still discovered from the include path. So this is W2
working. Swept rather than patched at the two reported sites — all 41 TUs under
`tests/compile/` were compiled both ways; six need the flag and W2 had already flagged four.

> **⚠️ #770 needs this commit too.** It is currently `CONFLICTING`/`DIRTY` and **not armed**
> (`autoMergeRequest` is null), and its W1 hits the same `nros.hpp` conflict. If it lands
> without this fix, `check-cpp` goes red on `main`.

### Verification

- **No loss.** The only subject present before and absent after is the W4 commit, which was
  *reworded*, not dropped. File-level: only `1204.breaking.md` (→ `1204.fix.md`) and the two
  template edits leave; `one_node_type.cpp` joins.
- **Patch-ids vs #770.** W2 matches exactly (`6998f0dda…` both sides). W1 differs, because its
  `nros.hpp` conflict had to be resolved against `main`'s changed context — and #770's own W1
  hits the identical conflict. Simulated it: cherry-picking #770 onto `main` with the same
  resolution and rebasing this branch on the result prints `skipped previously applied commit`
  for **both** W1 and W2, leaving exactly the 4 commits that are genuinely this PR's.
- **Gates.** `just check cpp` → *All C++ checks passed!*; `just check c` → *All C checks passed!*;
  `cargo +nightly fmt --check --all`, `doc-refs`, `issue-ids`, `prose-issue-refs`,
  `check-doc-commit-citations` all green. `just check fast`: **282 of 283** — the one failure is
  `xrce-vendored-versions`, reproduced identically on a plain `origin/main` worktree (unchecked-out
  vendored tree). `capability-conditionals` also fails on plain `origin/main` for the same class of
  reason and passes here once zenoh-pico is initialised.

jerry73204:
Rebased. Base is `main`; the W1/W2 commits are now **#770's own objects** (`4ae269dff`, `574fe259a`,
`2077ae65a` — identical SHAs, hence identical patch-ids), so they drop cleanly when #770 lands.
6 commits before, 5 after; one dropped **deliberately**, verified present on `main`:

- `fix(check): AGENTS.md cited a commit that the rebase it describes had already deleted` —
  `main`'s `7d7d04025 docs(AGENTS.md): the stacked-branch paragraph stranded its own citation, and
  it is red on main` fixes the same line, and names the stranded commit by SUBJECT rather than
  leaving it anonymous. Took `main`'s text whole.

No other commit lost (`comm` against the pre-push tip is otherwise empty).

## The one real conflict: `fix(#1225)` deleted the gate this PR extends

`scripts/check-cpp-capability-layout.sh` no longer exists — `main` rewrote it as
`scripts/check-cpp-capability-layout.py`, deriving ~90 subjects from clang's AST instead of three
authored names. W4's arm was **ported into the rewrite**, not reapplied against the old shape:

- `HOSTED_NO_STD_FLAGS = ["-std=c++17"]`, a tenth arm in `read_all`, and the comparison in
  `compare()` — which is now a function taking `nostd` alongside `forced`/`freestanding`, so the
  synthetic ratchet controls drive it too.
- **Selftest case 7 added** (7 cases, was 6): a layout that MOVES when the porting surface is
  requested must fail; a `hosted-only` entry must NOT excuse a hosted-no-std absence; a stale
  `std-only` entry must fail. Both ratchet directions, same shape as cases 4/5.
- `--report` gains the `no-std` column; the OK line names the arm.

## Two calls that are mine, with the measurement behind each

**1. The subject DERIVATION now asks for `-DNROS_CPP_STD` too.** Measured: the umbrella exports
**90** subjects with the flag and **89** without — `::rclcpp::NodeOptions` is the difference, since
W2 made the macros a request. Deriving without it silently narrows the subject list, which is
exactly what `#1225` exists to prevent, and it orphaned that subject's existing baseline line (the
first symptom: `NodeOptions is in the baseline as hosted-only but is no longer a derived subject`).
The flag is appended inside `cpp_capability_subjects.derive()` so the mutated-tree control cannot
drift from the real run.

**2. A third baseline kind, `std-only`.** W4's rule as authored is "a type that fails to compile
without the flag is NOT excusable" — written against five hand-picked types, none of which was
std-only. Over 90 derived subjects there is exactly one that is: `rclcpp::NodeOptions`, whose whole
surface is `std::string` + `std::vector<Parameter>` (`options.hpp`, `#if defined(NROS_CPP_HAS_STD_
STRING) && defined(NROS_CPP_HAS_STD_VECTOR)`), so there is no smaller shape of it to keep. It is a
**separate kind from `hosted-only`** on purpose: one line must not excuse two arms, or a type that
starts vanishing on the second one gets absorbed by an exemption written for the first. W4's teeth
are intact — an unlisted subject that vanishes without the flag still fails, and `hosted-only` is
explicitly rejected as a cover (selftest case 7 proves it).

## W4's acceptance, re-measured on the wider gate

    sizeof         baseline   no-std   freestanding
    ::nros::Node      200       200        200
    ::rclcpp::Node    200       200        200

Unchanged from what W4 reported — but now over 90 derived subjects rather than the 5 the arm was
authored against, and `rclcpp::Node` carries no exemption of any kind.

Docs corrected where they described the deleted script: `hosted_only_reason()` and "the exemption
list is EMPTY" are no longer true (`#1225` moved the registry to
`.config/cpp-capability-layout-baseline.txt`, which is not empty — 5 `diverges` + 3 `hosted-only`
+ 1 `std-only`, none of them `rclcpp::Node`). Updated in the roadmap, `changelog.d/1204.fix.md`
and the archived issue 1204.

## Gates

`just check cpp` **OK** — includes W4's own probe line
(`phase-438 W4: rclcpp::Node's unconditional surface instantiates (no NROS_CPP_STD; hosted +
-nostdinc++)`) · `just check c` **OK** · `just check fast` **284 gates, 1 failed** —
`check-xrce-vendored-versions`, which fails identically on a plain `origin/main` worktree
(vendored XRCE trees not checked out), reproduced · `check-cpp-capability-layout` **OK, 90 subjects,
selftest 7 cases** · doc-refs / issue-ids / prose-issue-refs / doc-recipe-refs /
check-doc-commit-citations **OK** · `cargo +nightly fmt --check --all` **OK**.

jerry73204:
## Rebased onto `main` — 5 commits in, 4 out

`98f7c562b` → `20509ba71` (rebased twice; `main` moved under the first push with the phase-437 batch, then the second rebase was clean).

No-loss check, both times:

```
comm -23 <(git log --format=%s origin/main..<before>|sort -u) \
```

empty except for the one commit deliberately dropped, below.

### Dropped: this branch's own W1

`4ae269dff refactor(phase-438 W1): one detection site for the six capability macros — and `nros.hpp` stops depending on include order`

Superseded by `cf4b0048a` on `main` (`fifteen copies of one capability predicate become one std_detect.hpp`). Not a patch-id drop — different subjects, different diffs — so it was verified by content instead:

* Both create `packages/api/nros-cpp/include/nros/std_detect.hpp` and both define all six `NROS_CPP_HAS_*` macros there, once.
* `main`'s version is a **superset**: it also folds `node.hpp`'s `<chrono>` block (−87 lines there), which this branch's W1 left in place.
* The one thing this branch's W1 had that `main`'s does not is the explicit `#include "nros/std_detect.hpp"` in `nros.hpp`. That is covered: `main`'s `nros.hpp` includes `nros/log.hpp` at line 20, unconditionally, before the first macro use at line 384, and `log.hpp` includes `std_detect.hpp`. Verified with `git grep -l std_detect.hpp origin/main -- packages/api/nros-cpp/` — 11 headers plus the file itself.

### Conflicts and resolutions

**1. `just/check/lanes.just`** — mechanical. `main` had rewritten the c++17 probe's comment (phase-426 W4 deleted `ComponentNode::adopt_launch_seed_`); W2 was adding `-DNROS_CPP_STD=1` to the same line. Kept `main`'s comment, kept W2's flag.

**2. `packages/api/nros-cpp/include/nros/std_detect.hpp`** — three prose hunks, all resolved toward W2 (the substance of the commit: discovery removed, `#if defined(NROS_CPP_STD)` and nothing else). Verified after resolution: six `#define NROS_CPP_HAS_*`, all in `std_detect.hpp`, every predicate `#if defined(NROS_CPP_STD)`, no `__has_include` outside comments.

Plus one **non-mechanical** call. `main`'s W1 carried a section `--- Why the predicate is a DISJUNCTION over a CONJUNCTION ---` (moved verbatim from `publisher.hpp`), which git merged cleanly because W2 never saw it. It is a live prescription — *"Gate on the declared std flavour, else ASK THE COMPILER — with BOTH probes"*, and *"do not gate on `NROS_CPP_STD` ALONE"* — sitting directly above code that now does exactly that. Left in, the file would contradict itself at the one site whose whole purpose is that the rationale can only drift once. **Deleted (40 lines).** Nothing unique is lost: the W2 block immediately below records the same history — issue 0112's `-fno-freestanding -nostdinc++`, issue 1240's `__has_include` / freestanding `#error`, GCC 16, issue 1187's arm-none-eabi 13.2 — and states the conclusion that now holds.

**3. `scripts/check-cpp-capability-layout.py`** — no conflict. `fix(#1225)`'s `.sh`→`.py` rewrite is in this branch's merge base (`git merge-base --is-ancestor d93cc31e6 7ff68c777` ⇒ true), so W2/W3/W4 were already written against the Python. Nothing to port.

### The rebase surfaced a real defect in the branch's own W1

`check-cpp-capability-layout` went red after the rebase, on the third arm this PR adds:

```
- ::rclcpp::Rate does not compile hosted WITHOUT -DNROS_CPP_STD.
- ::rclcpp::WallRate does not compile hosted WITHOUT -DNROS_CPP_STD.
```

Not a bad resolution — a measurement that could not be taken before. This branch's W1 left a **second** definition of `NROS_CPP_HAS_STD_CHRONO` in `node.hpp` (`git show 98f7c562b:.../node.hpp` line 142), still carrying the `__STDC_HOSTED__ && __has_include(<chrono>) && __has_include(<ratio>)` discovery that W2 deleted from the shared site. So on this branch a hosted TU without the flag kept the macro from the copy W2 never reached, and the two `<chrono>`-guarded types went on existing. `main`'s W1 folds that copy in; the arm then measures the **flag** instead of the include path.

Two definitions of one capability macro is issue 0135's shape — the thing W1 exists to remove. The branch was passing its own new gate because its W1 was incomplete.

Fix, folded into the W3+W4 commit (arm and baseline land together): `std-only ::rclcpp::Rate` and `std-only ::rclcpp::WallRate` beside the existing `std-only ::rclcpp::NodeOptions`, with a `--- the std-only entries ---` header section recording the above. `hosted-only ::rclcpp::NodeOptions` was **not** touched — it is on `origin/main` (line 70), it is a different axis, and the gate's own selftest (`a stale std-only entry fails`, `hosted-only does not excuse it`) confirms neither kind absorbs the other. The gate now reports:

```
check-cpp-capability-layout: OK -- 90 DERIVED subject(s) x 8 forced capability
macro(s) hosted, plus hosted-without-NROS_CPP_STD and a -nostdinc++
freestanding measurement against the ThreadX shim
```

### Gates

| gate | result |
| --- | --- |
| `just check fast` | 285/286 — only `xrce-vendored-versions` |
| `just check c` | OK |
| `scripts/check-doc-commit-citations.py` | OK — 209 citations resolve |
| `just check cpp` | **red, pre-existing on `main`** |

Both reds reproduced on a control worktree of plain `origin/main`, so neither is this branch:

* `xrce-vendored-versions` — environmental: `packages/rmw/xrce/xrce-sys/{micro-cdr,micro-xrce-dds-client}` are not checked out in a fresh worktree; the gate reports two SKIPs and exits 1.
* `just check cpp` — `-D clippy::missing-safety-doc` on `nros_cpp_declare_remap`, `packages/api/nros-cpp/src/lib.rs:3262`. Same file, same line on `origin/main`; this branch's diff against `origin/main` for that file is **empty**. This is a red on `main` that arrived with `feat(names): one node-FQN join…` and it will block anyone running `just check cpp` until it is fixed. Every C++ arm of the gate (the probes, the freestanding sweep, `heap_sequence_lifetime`, `cpp-capability-layout`) passes before it.

Base `main`, MERGEABLE, auto-merge armed.

PR #786, opened 2026-09-08
docs(#1227): four measured verdicts, one retraction, and why the bisect stopped

jerry73204:
Updates #1227 with what four board builds established, and stops the bisect
rather than continue with a method that had begun producing wrong answers.

## Measured

`mr_canhubk3/s32k344`, wiped build directory each time:

| commit | verdict |
| --- | --- |
| `1976727d8` | GOOD -- DTCM 93344 B, 71.22% |
| `e3786749a` | GOOD -- DTCM 93344 B, 71.22% |
| `d7e7bb477` | GOOD -- DTCM 93336 B, 71.21% |
| `fb94d1896` | BAD -- overflowed by 45040 bytes |

`d7e7bb477` is the newest commit known to build. Not environmental --
`e3786749a` reproduces 71.22% a day after the same tree first produced it, and
the three GOOD commits agree to within 8 bytes. Every BAD is the identical byte
count, which is itself evidence of one cause.

## Retraction

`01cb87d8f` was reported BAD earlier in this investigation. **Withdrawn.** It
built only because it ran before the probe directory was touched, so `sync`
skipped metadata regeneration and it consumed metadata emitted by another
commit's codegen. Its number is not attributable to that commit, and the
11-commit range that rested on it is withdrawn with it.

The first suspect, `dc2416f6e`, was also picked by reading subject lines. It is
bad -- but so is `3e5dc6e4f` below it, a CI cache fix that cannot move RAM.

## Why it stopped

Most commits in the interval cannot build the current ASI tree at all. The
CLI's `sync` step regenerates message metadata whenever the CLI changes, and at
eight measured commits that regeneration fails:

```
error: static assertion failed:
NROS_UNBOUNDED__nav_msgs_msg_odometry__field_header_frame_id
```

ASI's caps are not applied by those commits' codegen. Bisecting a library
against a fixed consumer assumes the consumer builds at every commit; here it
does not, for most of the interval. Finishing this wants the ASI side moved in
lockstep, or that codegen incompatibility understood -- not more builds of the
same shape.

Issue #1235 (PR #774) is the other thing in the way: `sync` refuses a correctly
paired resolver on these commits, and the sanctioned rebuild commands cannot
repair what their own errors name.

## Note on the baseline

`.config/prose-issue-ref-baseline.txt` gains ONE row, because 1227 now cites
1235 and 1235's file is in #774. Added by hand rather than with
`--write-baseline`, which rewrites the file and would have dropped 22 existing
rows -- the ratchet-shrink that put `main` red three times this week.

### DIFF
--- a/docs/issues/1227-island-image-dtcm-overflows-on-main.md
+++ b/docs/issues/1227-island-image-dtcm-overflows-on-main.md
@@ -103,6 +103,62 @@ Issue 1235: the `sync` step refuses a CORRECTLY paired resolver on these commits
 `setup-cli` / `setup-launch-resolve` cannot repair what their own errors name.
 Every bisect step needs this working around before it can even reach a compiler.
 
+## ROOT CAUSE FOUND -- `0368d4040`, and it is a correct fix exposing a real gap
+
+Bisected on the HOST, seconds per step instead of twelve minutes. `nros-node`'s
+`build.rs` derives the arena from the knobs and prints `cargo:arena_size`, so
+the island's number is reproducible with a plain `cargo build -p nros-node` and
+the island's knobs in the environment -- no Zephyr, no board, and every commit
+in the interval builds, which is what made it searchable at all after the board
+route stalled.
+
+| commit | derived arena |
+| --- | --- |
+| `316915dc3` (parent) | 61,936 |
+| **`0368d4040`** | **207,096** |
+| `6e7779b05` (main today) | 207,096 |
+
+`git bisect run` over `d7e7bb477..fb94d1896` names `0368d4040` --
+`fix(#1190): the arena priced a subscription's QoS history at three slots`. The
+arena grows by 145,160 B in one commit and has stayed there since.
+
+### It is not a mistake to revert
+
+That commit CORRECTED an under-count. The old model charged `3 * rx_buf + 512`
+per subscription -- a `TripleBuffer`, the `depth <= 1` case -- while the runtime
+registers `KEEP_LAST(10)`: eleven slots plus a length array. Issue 1190 is the
+symptom of the old number, a zenoh listener taking `BufferTooSmall` on register.
+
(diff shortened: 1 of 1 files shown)


jerry73204:
Rebased onto `main` (`7ff68c777`). 2 commits before, 2 after; no-loss check (`comm -23` of subjects against the pre-push sha `497cd157b`) empty.

**One conflict: `.config/prose-issue-ref-baseline.txt`.**

This branch's copy was the pre-repair scrambled one (`--write-baseline` output, comments after the first data row). `main` has since repaired that header (`fix(#1241)`), and the new `check_baseline_shape` refuses a baseline whose comments follow a data row — so the branch side could not be kept.

Resolved by taking `main`'s repaired file verbatim, after measuring that neither row this branch added is still needed:

- `docs/issues/1227-…md:1235` — `docs/issues/1235-cli-resolver-pairing-blocks-building-older-commits.md` is **on `main`** now (`git ls-tree origin/main docs/issues/`), so the 1235 citation resolves and the row would be a stale ratchet entry.
- `docs/issues/1231-…md:1164` — `docs/issues/1231-*.md` does not exist on `main`; that row belonged to a different branch's citing file, not to this one.

Net effect: this PR's diff against `main` is now the single file `docs/issues/1227-island-image-dtcm-overflows-on-main.md`.

Gates (docs-only diff): `check-doc-refs` OK · `check-issue-ids` OK (1211 issues) · `check-prose-issue-refs` OK (9 known in-flight, no new dangling) · `check-doc-recipe-refs` OK (self-test + 47 known dead, no new) · `check-doc-commit-citations.py` OK (208 citations across 907 files resolve — including this PR's `0368d4040`) · `cargo +nightly fmt --check --all` clean.

`MERGEABLE` / `CLEAN`, and now `QUEUED` in the merge queue (position 8).

PR #787, opened 2026-09-08
docs(phase-441): the one "on-target" live-peer cell runs on host sockets

jerry73204:
Phase-433 closed at 20/20 Runtime interop cells with a live verdict, and named
its own remainder — W7, *"an RTOS image against a stock peer, on a platform that
is not Zephyr"*, explicitly flagged as possibly its own phase. This is that
phase.

Writing it required looking at what the existing on-target cell exercises, and
the answer is less than its name says.

## Measured

`interop::CELLS` has exactly **one** non-Linux runnable row —
`zephyr-qos-rust-zenoh` on `ZephyrNativeSim` — plus one carved sibling. Every
other row is `PlatformId::Linux`. So live-peer coverage of the RTOS ports is one
platform, one language, one RMW, one workload.

`FreertosMps2`, `FreertosPosix`, `NuttxArm`, `NuttxRiscv`, `ThreadxLinux`,
`ThreadxRiscv64`, `Esp32Qemu`, `QemuBaremetal`, `Fvp`, `Px4` and
`ZephyrQemuCortexM` have **no live-peer cell at all**.

## What changes the argument

`ZephyrNativeSim` is `native_sim/native/64`, where sockets are **offloaded to
the host** and the pointer width is the host's. The tree already says so, in
`ZephyrQemuCortexM`'s own doc comment — which also records that bringing that
second witness up cost **five real defects, every one invisible to native_sim**.

So our one on-target interop cell never puts an RTOS network stack in the path.
It proves our zenoh-pico code talks to `rmw_zenoh_cpp`; it does not prove it
does so from a device. The cell is not worthless — it caught issue #141 — but
the phrase "on-target" is doing work the artifact does not support.

## The constraint that decides the ordering

The existing cell already connects by TCP to a baked locator
(`CONFIG_NROS_ZENOH_LOCATOR`) rather than by multicast, and
`scripts/qemu/setup-network.sh` records the peer-side half
(`scouting/multicast/enabled=false`).

QEMU offers slirp (unicast only, no root) and TAP (multicast, needs
`ip tuntap add`). **CLAUDE.md forbids `sudo`, so TAP cannot be the default
path.** Everything proposed must work over slirp — unicast to a known
endpoint — which is what zenoh-pico already does. Cyclone's SPDP is multicast by
default, so whether it crosses slirp at all is a **finding to record rather than
assume**: that is W2, a measurement, not a cell.

Also planned around rather than discovered: issue 0432 means the first target
cannot be the Rust image (`zephyr-lang-rust` cannot compile for a board whose
devicetree has gpio nodes), so W1 is a C or C++ cell.

## Work items

| | |
| --- | --- |
| **W1** | the same workload, one board over — native_sim → QEMU Cortex-M |
| **W2** | is Cyclone reachable from a guest at all? a measurement |
| **W3** | a second *kernel*, not a second Zephyr board |
| **W4** | the lane needs QEMU and the ledger cannot say so |
| **W5** | retire the phrase "on-target", or earn it |

W1 states its expected finding in advance so it is falsifiable: phase-337 W2
found five defects crossing this same boundary with a non-interop workload. If
W1 finds none, *that* is worth writing down.

## What it does not promise

Not real hardware — QEMU is a target-shaped emulator, and calling it on-target
is the same overreach this phase exists to fix one level down. Not every
platform. Not the QoS workload's completeness. Nothing about `Px4` or `Fvp`,
neither of which has a QEMU networking story here.

## Note on the second commit

`check-prose-issue-refs` refused the doc for citing issue 1164, whose file is on
PR #641. That is the in-flight case the baseline exists for, and issue 1231's
citation of 1164 already has a row there. Added **by hand**: the gate's failure
message recommends `--write-baseline`, which regenerates the file and drops the
authored per-row classification block — running it replaced four paragraphs of
rationale with the generic header. Following the advice destroys the record it
asks you to add to. Not fixed here, but noted in the commit so the next person
hand-edits too. Both 1164 rows leave when #641 lands.

Phase id reserved with `just phase-new` (`refs/phase-ids/441`). Docs only.

https://claude.ai/code/session_01TzPvAy8k8yiuoGmKMzKyiJ

### DIFF
--- a/.config/prose-issue-ref-baseline.txt
+++ b/.config/prose-issue-ref-baseline.txt
@@ -32,12 +32,18 @@
 #         whoever claimed the id. Not deleted: a reference removed is a
 #         record lost, and the id is not free either.
 #
+#   1164  IN FLIGHT — `docs/issues/1164-*.md` is on PR #641 (the phase-433 W6
+#         events cell). Cited by phase-441, whose W1 verdict table names the
+#         three issues the on-target cell demonstrated. The citation is
+#         correct and the file is coming; the row leaves when #641 lands.
+#
 docs/issues/1092-rmw-shape-gate-licenses-deviation-without-pinning-it.md:9999
 docs/issues/1116-rustdoc-diagnostics-outside-the-published-crate-set.md:1110
 docs/issues/1163-compile-smoke-checks-nros-c-under-default-features-only.md:1080
 docs/issues/1186-executor-backing-latch-test-races-siblings.md:0417
 docs/roadmap/phase-428-api-porting-principle-sweep.md:1099
 docs/roadmap/phase-432-codegen-one-producer-many-packs.md:1080
+docs/roadmap/phase-441-rmw-on-target-verification.md:1164
 just/check/docs.just:1110
 just/check/lanes.just:1099
 scripts/build/rustdoc-set.sh:1110
--- a/docs/roadmap/phase-441-rmw-on-target-verification.md
+++ b/docs/roadmap/phase-441-rmw-on-target-verification.md
@@ -1,0 +1,186 @@
+# Phase 441 — the one "on-target" live-peer cell runs on host sockets
+
+**Status (2026-09-09). Not started. Analysis below; work items W1–W5 proposed.**
+
+Phase-433 closed with twenty of twenty Runtime interop cells carrying a live
+verdict, and named its own remainder:
+
+> ### W7 — one on-target cell that is not QoS
+>
+> An RTOS image against a stock peer, on a platform that is not Zephyr. This is
+> the largest job here and the one with the weakest current argument, so it goes
+> last and may reasonably become its own phase.
+
+This is that phase. The argument is stronger than phase-433 could make it,
+because writing it down required looking at what the existing on-target cell
+actually exercises — and the answer is less than its name suggests.
+
+## What we have, measured
+
+`interop::CELLS` has exactly one non-Linux runnable row:
+
+```rust
+ic("zephyr-qos-rust-zenoh",
+   c(ZephyrNativeSim, Rust, Zenoh, Qos, Interop, Runtime), …)
+```
+
+plus one carved sibling (`zephyr-qos-cpp-cyclone-CARVED`). Every other row is
+`PlatformId::Linux`. So the live-peer coverage of the RTOS ports is: **one
+platform, one language, one RMW, one workload.**
+
+`FreertosMps2`, `FreertosPosix`, `NuttxArm`, `NuttxRiscv`, `ThreadxLinux`,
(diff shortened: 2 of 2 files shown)


jerry73204:
Rebased onto `main` (`7ff68c777`). 3 commits before, 2 after — one deliberate drop, below. No-loss check (`comm -23` of subjects against pre-push sha `ede732023`) shows exactly that one subject and nothing else.

**Conflict: `.config/prose-issue-ref-baseline.txt`, twice (commits 2/3 and 3/3).**

This branch carried the pre-repair scrambled baseline (comments after the first data row). `main` has since repaired that header (`fix(#1241)`), and the new `check_baseline_shape` refuses a baseline whose comments follow a data row — so the branch side could not survive as written.

Resolved by rebuilding on top of `main`'s repaired file: the `1164` classification note goes in the CLASSIFIED block above all data rows, and the row goes in sorted position after `phase-432`.

Measured, not assumed:
- `docs/issues/1164-*.md` is still absent from `main` and from this tree, and `docs/roadmap/phase-441-rmw-on-target-verification.md:135` cites it (`1137, 1164 and 1231 each demonstrated.`) — so the row is still needed and the citation is still correct.
- The branch's other row, `docs/issues/1231-…md:1164`, was dropped: `docs/issues/1231-*.md` does not exist on `main`, so that row baselines a citation from a file that is not here. It belongs to whichever branch carries 1231, not to this one.

**Deliberately dropped commit:** `docs(phase-441): re-anchor the baseline row away from PR #788's deletion`. Its entire purpose was to move this row off the line adjacent to `docs/issues/1231-…md:1164`, which PR #788 deletes — and that neighbouring row does not exist on `main`. The row now sits between `phase-432` and `just/check/docs.just`, adjacent to no other branch's pending delete, so the adjacency the commit removed is already absent. Its reworded note is carried into the note this rebase writes.

Baseline delta against `main` is now +1 row and +1 comment block, nothing else.

Gates (docs-only diff): `check-doc-refs` OK · `check-issue-ids` OK (1211 issues) · `check-prose-issue-refs` OK (10 known in-flight, no new dangling) · `check-doc-recipe-refs` OK (self-test + 47 known dead, no new) · `check-doc-commit-citations.py` OK (196 citations across 908 files resolve) · `cargo +nightly fmt --check --all` clean.

jerry73204:
Superseded by #801 — same phase doc, rebased onto main with the baseline commits dropped.

This branch carried two extra commits adding a `prose-issue-ref-baseline` row for phase-441's citation of issue 1164, whose file was in flight on PR #641. #641 has since merged, so the citation resolves on its own and the row would be a dead entry in a ratchet that may only shrink. Rebasing was the right fix and the ruleset declines a force-push here, so #801 is the same single commit on current main. Net diff there is the phase doc alone.

PR #792, opened 2026-09-08
feat(phase-436 W7 + #1242): a C-ABI for the wake-source seam, and Zephyr parks through it

jerry73204:
> **Stacked on #648.** Rebase to `main` once it lands.

## #1242 — the park granularity comes from the primitive

`park_granularity_us()` was a hardcoded `1_000`, justified by "every `nros_platform_wake_*` port takes `uint32_t timeout_ms`". The ABI signature is not the limit in either direction:

| | constant claimed | reality |
|---|---|---|
| **ThreadX** | 1 ms | **10 ms** tick (`TX_TIMER_TICKS_PER_SECOND=100`, both board configs) |
| FreeRTOS | 1 ms | 1 ms — correct by coincidence |
| **POSIX/NuttX** | 1 ms | **timespec-native**, could do µs |

And it never consulted the installed primitive, so a port supplying a microsecond `ParkUntilFn` still couldn't produce a sub-ms park. W3's jitter report reads this to say what its number is worth — so on ThreadX it claimed precision against a 10 ms tick. **That is the failure W3 exists to prevent, one layer below W3, in its own dependency.** Both are mine.

`set_park_primitive` now takes the granularity the port declares.

## W7.a — the C-ABI surface

The seam existed only as a Rust method. `nros_cpp_executor_register_wake_source`, `..._set_park_primitive`, `..._wake_handle`.

## The trap that made `wake_handle` necessary

Wiring Zephyr, I checked what happens to the async wake once a park is installed:

```rust
let timeout_ms = if parked { 0 } else { ... };   // → wake.wait_ms(0)
```

**A park primitive silently demotes the async wake to a poll.** A park waiting on its own fresh semaphore can't be broken by the backend's listener — the listener signals the *executor's* object — so it sleeps its whole deadline with data already waiting. Worse than not parking. It passes every test and is invisible in a quiet run.

`wake_handle` hands the port the right object; the constraint is written into the platform header, the shim and the board entry, because the next port gets it wrong by default.

## Zephyr

`nros_platform_wake_park_until_us` — `k_sem_take` with **`K_USEC`**, the same wait without the millisecond floor its own signature imposes. `..._park_granularity_us` derives the real floor from `CONFIG_SYS_CLOCK_TICKS_PER_SEC`, since the kernel rounds `K_USEC` up to whole ticks. Both **optional** in `platform.h`.

## Verification

Instrumented at both ends, because installing and being called are different failures a quiet run can't separate:

```
park install tier=`control` wake=0xaad9d0 gran=1000us rc=0
park CALLED first deadline=5000us
```

Non-null wake = the trap avoided. 5000 µs = ASI's declared `spin_period_us`, not the 10 ms the tier loops pass `spin_once` — W1's min and #648's nominal in one number.

**Limit stated plainly:** this board's tick is 1 kHz, so the derived granularity is 1000 µs — the same number the old constant produced. The run proves the value *flows from the port*; the unit tests carry the behavioural difference (10 ms coarser, 1 µs finer).

Also fixes a latent gate of mine: `nros_cpp_executor_set_spin_nominal_us` called `cpp_ctx_checked` without the `rmw-cffi` gate that function carries.

**Scope:** Zephyr's C arm only. ThreadX, bare metal and Zephyr's Rust arm remain unwired, and **bare metal has no one-shot timer to wire** — no board in the tree can arm a deadline compare today.

378 passed, 0 failed. Image builds, boots, installs on every tier, no violations.

https://claude.ai/code/session_01XPKwuBESsmHcbgkCN3odw4

### DIFF
--- a/packages/api/nros-cpp/include/nros/nros_cpp_ffi.h
+++ b/packages/api/nros-cpp/include/nros/nros_cpp_ffi.h
@@ -1488,6 +1488,62 @@ nros_cpp_ret_t nros_cpp_executor_set_active_groups(void *executor,
  * `tiers` must be a valid pointer to `n_tiers` [`NativeTierSpecC`] entries,
  * valid for the duration of the call. `session_name` is NULL or a valid
  * null-terminated string.
+ * phase-436 W7.a — register a platform deadline source on this executor.
+ *
+ * The seam existed only as a Rust method, so the board entries that go through
+ * this ABI — FreeRTOS, NuttX and the C arm of Zephyr, which is the arm the
+ * Zephyr FVP lane takes — could not reach it at all. They hold an opaque
+ * handle, not a typed `Executor`.
+ *
+ * `next` returns microseconds FROM NOW until this source next needs the
+ * executor; `u64::MAX` means nothing pending, which is how an asynchronous
+ * source spells itself while breaking the park by signalling instead.
+ *
+ * # Safety
+ * `handle` must be a live executor handle from this ABI, or NULL. `next` and
+ * `ctx` must outlive the executor.
+ */
+nros_cpp_ret_t nros_cpp_executor_register_wake_source(void *handle,
+                                                      uint64_t (*next)(void*),
+                                                      void *ctx);
+
+/**
+ * phase-436 W7.a — the wake object a port's park primitive must wait on.
+ *
+ * Returns NULL when this build has no wake primitive linked.
+ *
+ * A `ParkUntilFn` that waits on its own fresh object cannot be broken by the
+ * backend's listener, because the listener signals THIS one — and installing
+ * such a park makes latency WORSE, since `spin_once` drops the transport
+ * drain to non-blocking once a port has parked. Pass this as the park's
+ * `ctx`.
+ *
+ * # Safety
+ * `handle` must be a live executor handle from this ABI, or NULL.
+ */
+void *nros_cpp_executor_wake_handle(void *handle);
+
+/**
+ * phase-436 W7.a — install THE primitive this executor blocks on.
+ *
+ * Singular by nature: only one thing can actually wait. `granularity_us` is
+ * the finest park the primitive can express, and the port states it because
+ * only the port knows — an RTOS tick is a coarser limit than the ABI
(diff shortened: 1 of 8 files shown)


jerry73204:
Added the two Rust arms — **ThreadX** and **Zephyr's Rust entry**. Both hold a typed `Executor`, so unlike the C arms they needed no new FFI, only the primitive and the wiring.

`port_park::install_port_park` puts the non-obvious part in **one** place rather than once per board: a `ParkUntilFn` must wait on the executor's own wake object, or it cannot be broken by data arriving and sleeps its full deadline with a message waiting. Stating that per-board is how it drifts.

**ThreadX gains the primitive and buys no resolution** — that's the point. `tx_semaphore_get` counts ticks, `TX_TIMER_TICKS_PER_SECOND` is 100 on both shipped boards, so the real floor is **10 ms**. Installing it makes the executor stop *claiming* the 1 ms it never had there (issue 1242). Conversion rounds up: a short wait is a missed deadline, a long one merely late.

Four compile problems on the way, three of them the same shape — a gate landing on the wrong item:

- `::nros::executor` doesn't exist; the facade never re-exported it → now `::nros::port_park`
- my Zephyr insertion split a `#[cfg(feature = "zephyr-edf")]` pair of `apply_tier_deadline` definitions — same class as the W7.a insertion that stole `nros_board_native_run_tiers`'s gate
- `port_park`'s import was ungated while its function is gated → `-D warnings` rejected it (issue 0487's class)
- ThreadX's install belonged in the shared `apply_tier`, not duplicated at the boot site

**And one green result that wasn't evidence:** `cargo check` on the Zephyr board passed *while never compiling `entry_tiers.rs`* — it sits behind `#[cfg(feature = "tiers")]`. Real errors only appeared under `--features tiers`. ThreadX can't be checked standalone at all (pre-existing cyclic dep with `nros-board-threadx-linux`, verified on an unmodified tree), so it's checked through that dependent.

378 passed, 0 failed. Zephyr board clean with and without `tiers`; ThreadX clean via its dependent.

Bare metal stays unwired — no board can arm a one-shot deadline compare, which is a hardware-code gap, not a wiring one. Happy to take that separately.

PR #797, opened 2026-09-08
phase-427 W4: ComponentNode deleted — its five directories move to the one node type

jerry73204:
Implements **phase-427 W4**. `nros::ComponentNode` is deleted; its members are on
`nros::Node`, its timer pool is the opt-in `nros::NodeWithTimers<N>`, and its
macros are in `component.hpp`.

**Based on `main`, not on `phase-427-cpp-one-node-type`** — PR #755 merged on
2026-09-08 and its branch was deleted, so the base the task named no longer
exists. All 13 of its commits are in `main` (verified by subject; one was
reworded in the rebase).

---

## Blocker 1 — the silent collision, resolved by RENAME

The ours-only family takes `_in` names: `create_publisher_in`,
`create_subscription_in`, `create_wall_timer_in`, `create_timer_in`. Third
application of the rule in this campaign, after the reordered C node initialiser
and the clock-taking C timer verb.

**The collision is real. The trigger is not the spelling the work item named.**
Measured against the real headers, gcc 13 `-std=c++17`:

| ported call | binds | how loud |
| --- | --- | --- |
| `create_publisher<M>("chatter", 10)` | UPSTREAM (`shared_ptr`) | correct, silent |
| `create_publisher<M>("chatter", rclcpp::QoS(10))` | **OURS (by value)** | accepted, note only |

The item predicted the integer form would bind ours, reasoning that decay is a
standard conversion where `std::string` needs a user-defined one. It does not:
`int -> size_t` is also standard and rescues upstream's second argument, so ours
is not better in *every* argument. What collides is the **explicit-QoS** form —
`nros::QoS(int)` is `explicit`, `rclcpp::QoS(size_t)` is not — and gcc takes it
as an extension, emitting nothing but *"ISO C++ says that these are ambiguous"*.
That is the spelling a ported rclcpp file is most likely to carry, so this is a
stronger argument for the rename, not a weaker one.

**Proof + negative control:** `tests/compile/one_node_type_ours_only_names.cpp`,
wired into `just check cpp`. It asserts with `decltype` on the call expression —
the only way to ask which overload the compiler *picked*:

- **positive** — on the real merged type, every ported spelling reaches
  upstream's overload, and `create_publisher_in` is still reachable and still
  returns by value;
- **negative control** — reconstructs the pre-rename shape and asserts *that*
  shape binds the wrong overload. Without it the file would pass equally well if
  the collision had never existed.
- It also pins the corrected prediction, so the reasoning cannot rot into folklore.

## Blocker 2 — the unconditional bytes, resolved by SPLITTING

`sizeof(ComponentNode)` measured **55,784 bytes**. The 24-byte latch and
192-byte pool are rounding error next to its inline
`ParameterServer<256, 8, 4096>` (~55.5 KB) — which **no consumer uses** (zero
`declare_parameter` call sites across all three consuming packages), so the
parameter facade did not move.

- **Error latch: stays unconditional**, as ruled. It is the channel a
  `-fno-exceptions` target has instead of a throwing constructor; behind the
  hosted block, the firmware case — the one that cannot throw — would have none.
- **Timer pool: a template parameter defaulting to zero**, as ruled, realized as
  `template <::size_t MaxTimers> class NodeWithTimers : public Node`.

`Node` itself could not become the template, measured rather than preferred:
`rclcpp::Node` is an alias asserted by `std::is_same`; 218 sites spell `Node`
with no argument list; and every `Node&` parameter would otherwise accept one
depth only, so a component with a pool would not be a node the executor could
take. Deriving keeps the IS-A relation `ComponentNode` never had.

### `sizeof(nros::Node)`

| | hosted `g++ -std=c++17` | freestanding `-nostdinc++` (ThreadX shim) |
| --- | --- | --- |
| before | 200 | 200 |
| **after** | **224** | **224** |

+24 — exactly the latch — and identical in both configurations, so
`check-cpp-capability-layout` still holds: a probe may gate a METHOD, never
`sizeof`. `NodeWithTimers<N>` adds its pool on top, opt-in. The POC's `Listener`
owns no timer and derives `Node` directly; under `ComponentNode` it paid for
eight slots.

## The gates: 14, not 4

**9 would have hard-errored** — api-parity's TU list (`extract_cxx` raises on
any clang error), **`nros.hpp`'s own include** (widens that from one TU to all
four *and* breaks `just check cpp`'s header glob — the biggest miss in the
original report), 4 compile TUs, `entry.cpp.jinja`, the example/e2e surfaces.

**5 would have passed vacuously** — the guard-probe's `PROBE_CONTEXT` key and
its upstream half, two `emit_cpp.rs` assertions on *emitted text*, and 18 ledger
rows (api-parity has no orphan check).

**1 was misreported**: `check-cxx-standard-floor.py` reads its floor from
`CMakeLists.txt` via regex and never names the header — prose only, unaffected,
so nothing was changed in it.

Each retarget carries a negative control. The guard-probe gained a **stale-key
check** — an entry naming a nonexistent file is now a failure — verified to flag
the old key and to find zero stale keys today. `emit_cpp.rs` now pins both
directions.

## Ledger

18 rows: **14 deleted** (subject is now literally an existing `cpp:Node::*`
row), **4 renamed**, **3 added**. Three surviving rows had their `why` rewritten
because the merge changed their meaning. `api-parity --check
--require-disposition` ends *"every divergence carries a ledger entry"*.

## Split out

**Issue #1247** — "one of them builds for a freestanding target" is a **port,
not a check**: no subnode package has ever had a freestanding fixture row (all
three consumers are `platform = "linux"`). Filed rather than attempted, per the
task's split.

Also recorded, not acted on: deleting `adopt_launch_seed_` removed the tree's
**last `if constexpr`**, so the `cxx_std_17` floor has no live justification.
Issue 1118 raised it deliberately; lowering it is its own decision.

## Verification

| | |
| --- | --- |
| `just check cpp` | **88 probes green** (clippy step red: zenoh-pico submodule absent in worktree) |
| `just colcon-parity` | **rc=0** — the dual-compile gate, templates built against real rclcpp |
| `api-parity` / `api-parity-ledger` | green, with the required closing line |
| `cpp-capability-layout` | OK — 3 types x 7 macros, no layout follows a probe |
| `ffi-struct-mirrors`, `cbindgen-headers`, `cpp-fmt` | green |
| `cargo test -p nros-cli-core` | 989 passed, 0 failed |
| `just check fast` | 273/277 |

**Environmental reds, named not fixed** (identical before and after this
branch): `capability-conditionals` + `xrce-vendored-versions` +
`xrce-source-manifest` (submodules not checked out; the last is a sandbox
`PermissionError` in its own self-test), and `doc-commit-citations` (cites
`22e511442`, which this clone lacks, from a file this PR does not touch).


jerry73204:
## Rebased onto `main` (`cf4b0048a`) — CONFLICTING -> MERGEABLE

`f96fed115` -> `31cbb88ef`. Branch point moved `a9a9febcd` -> `cf4b0048a`, 49 commits of `main` picked up. **7 commits before, 8 after** (one added, see below); every original subject verified present:

```
comm -23 <(git log --format=%s cf4b0048a..f96fed115|sort -u) \
         <(git log --format=%s cf4b0048a..HEAD|sort -u)   # empty
```

### Conflicts and how they were resolved

**1. `nros/component_node.hpp` — modify/delete.** `main`'s phase-426 W4 (`3bf30405f`) rewrote this file's parameter facade; this branch deletes the file. Resolved as DELETE. Nothing was lost: W4's replacement — `detail::node_param_declare/get/has` forwarding to the executor's one store — is on `nros::Node` in `node.hpp`/`nros.hpp` already, and it is `Node`'s copy that survives the merge. The only ComponentNode-local addition with no twin is `declare_error_what_`; see the note on the loud arm below.

**2. `nros.hpp`, `detail::adopt_executor_param_seed` — content.** This branch DEDUPLICATED the helper (deleting `ComponentNode::adopt_launch_seed_`, its C++17 twin); `main` DELETED it outright, because with one store there is nothing to reconcile and the adoption is the FFI's own `ALREADY_EXISTS` path. **Took `main`** — it is the more advanced answer to the same problem, and it subsumes the branch's.

**3. `nros.hpp`, the `-- parameters --` block — content.** Same shape: branch documented the launch-seed adoption it was preserving, `main` documented the single-store forwarders that replace it. **Took `main`.**

**4. `docs/reference/api-parity-ledger/param.json` — content.** `main` extended the `why` of the three `cpp:ComponentNode::*_parameter` rows with W4 notes; this branch deletes those rows because their subject is gone. **Took the deletion** — a `why` on a row for a deleted type is not worth preserving, and `api-parity.py` has no orphan-row check, so they would have rotted silently (that is what `e7690e365` is about). Shard re-parsed with an `object_pairs_hook` duplicate-key counter: **no duplicate keys**; the other 16 shards were untouched by the rebase.

### Two edits beyond conflict resolution

Both are prose on `main` that this branch makes FALSE, in files git merged cleanly because neither side touched the same lines:

- `nros.hpp:52` said the parameter facade is "the ONE parameter facade **both node types** wear... so `rclcpp::Node` and `nros::ComponentNode` cannot disagree". After W4 there is one type. Reworded; the guarantee is stronger, not weaker.
- `nros.hpp:588` said "the loud half is `nros::ComponentNode`, whose facade makes it boot-fatal" — the type this PR deletes. Now states the fact: no C++ path is boot-fatal on a missing parameter store any more, `nros::Node` still carries the `set_error`/`ok()` flag, and routing the parameter path back onto it is a separate decision because it changes what an upstream-shaped `declare_parameter` does. **Flagging this for review** — it is the one behavioural loss in the merge, and it is the branch's own consequence (deleting the class deletes its facade), not something the rebase introduced.

### The added commit — `31cbb88ef`

Two gates landed on `main` AFTER this branch was cut, so neither was in W4's `grep -rn 'component_node'` sweep, and both name `nros::ComponentNode` as a live subject:

- **`.config/cpp-capability-layout-baseline.txt`** — issue 1225 (`d93cc31e6`) replaced the three-authored-name gate with a ~90-subject derivation, and its ratchet carries `diverges ::nros::ComponentNode`. The gate is explicit that a baselined subject the derivation no longer produces is a **failure**, not an inert line (the issue-0743 class), so this would have gone red on the merge. The row **follows its subject**: the cause is `nros::Timer`'s one `#ifdef NROS_CPP_STD` member, and whatever holds a `Timer` array pays 8 bytes per slot — that is now `nros::NodeWithTimers<4>` (328 -> 360, measured). One line either way, the ratchet does not grow, and `Node` itself holds no timer array and does not diverge, which is what W4 set out to make true.
- **`packages/api/nros-cpp/tests/compile/param_hosted_overloads.cpp`** — the phase-426 W4 probe (`22c991bf6`) that finally compiles the `std::vector<T>` parameter overloads. Its second half subclassed `::nros::ComponentNode`. Retargeted to `::nros::Node`, whose `Node(NodeHandle, name)` ctor is where that shape lives after the merge. Its framing changes with it: its two halves were two facades that could disagree, and are now two call SHAPES over one.

phase-438 W1 (`std_detect.hpp`) needed nothing — this branch re-adds no inline `NROS_CPP_HAS_*` predicate block; `node.hpp` includes `nros/std_detect.hpp` as `main` left it.

### Gates

| gate | result |
|---|---|
| `just check fast` | **284 / 285** — only `xrce-vendored-versions`, which fails identically on a plain `origin/main` worktree (no vendored tree checked out) |
| `just check cpp` | all **88** compile probes green; clippy step red on `nros_cpp_declare_remap` missing `# Safety` — **pre-existing on `main`**: `git diff cf4b0048a HEAD -- packages/api/nros-cpp/src/` is EMPTY, and `main`'s own copy of the function has no `# Safety` either |
| `just check c` | OK |
| `just check api-parity` | "every divergence carries a ledger entry" |
| `just check api-parity-ledger` | self-test 0 failures |
| `scripts/check-doc-commit-citations.py` | OK — 209 citations resolve |
| `scripts/check-cpp-capability-layout.py` | OK — 90 derived subjects x 8 forced macros + freestanding arm |
| `cargo +nightly fmt --check --all` (root + `packages/cli`) | clean |
| `clang-format --dry-run -Werror` on the edited C++ | clean |
| `cargo check -p nros-cli-core --all-targets` | clean |

Base `main`, MERGEABLE, auto-merge armed.

jerry73204:
## Rebased onto `main` (`b5097a460`)

`31cbb88ef` → `38c5cf274`. **8 commits before, 9 after** — the extra one is a rebase fixup for a way `main` moved under this branch (below). No-loss check (`comm -23` on subjects, old tip vs new HEAD) is **empty**.

### Conflicts and resolutions (2 files)

**1. `docs/roadmap/phase-427-one-node-type.md`** — status line, COMPOSED. Main added "(W5's runtime half and W6's book page closed 2026-09-09)"; this branch adds W4 to the landed list and notes its freestanding acceptance split out as issue 1247. Both facts kept:

> **Status (2026-09-09). W1, W2, W3, W4, W5, W6 and W8 LANDED (W5's runtime half and W6's book page closed 2026-09-09; W4's freestanding acceptance split out as issue 1247); W7 blocked and re-scoped**

**2. `.config/cpp-capability-layout-baseline.txt`** — main added a `std-only` COMMENT section (phase-438 W4's arm) in the same region where this branch DELETES the data row `diverges ::nros::ComponentNode`. Kept main's comment block, dropped only the row. `diverges ::nros::NodeWithTimers<4>` — the subject the row follows — is present exactly once. Comments still sit above the first data row (line 99), so the baseline's shape rule holds, and `check-cpp-capability-layout` is green over its 90 derived subjects.

### The one new commit: `fix(rebase): the _in-names lane needs the flag that now grants the hosted half`

Not cosmetic — the lane this branch adds was **red** after the rebase.

phase-438 W2 landed on `main` after this branch was cut and deleted the `__has_include` discovery arm from `std_detect.hpp`, so `NROS_CPP_HAS_SHARED_PTR` / `_STD_STRING` / `_STD_VECTOR` / `_STD_FUNCTION` — and therefore `NROS_CPP_NODE_HOSTED` — are now granted by `NROS_CPP_STD` **and by nothing else** (verified: `origin/main:packages/api/nros-cpp/include/nros/std_detect.hpp` defines all four under `#if defined(NROS_CPP_STD)` alone).

`one_node_type_ours_only_names.cpp` asks the compiler which overload it PICKS between our `create_publisher(const char*, const QoS&)` and upstream's `create_publisher(const std::string&, …)`. Upstream's half lives behind `NROS_CPP_NODE_HOSTED` (`node.hpp:625`, `:629`), so without the flag it is not merely unselected — it is **absent**, and all four `decltype` call expressions failed with `no matching function` instead of measuring the collision the lane exists to pin. `-DNROS_CPP_STD` added, matching the three sibling `one node type:` lanes that already carry it, with the reason recorded in the recipe so it is not re-litigated.

Verified: `just check cpp` green, all cases, including this lane's own negative control still binding the WRONG overload (the two ISO-ambiguity warnings on the reconstructed pre-rename shape are the control working).

### Gates

| gate | result |
| --- | --- |
| `just check fast` | 286/287 — only `xrce-vendored-versions` |
| `cargo +nightly fmt --check --all` | 0 |
| `just check api-parity` | 0 |
| `just check api-parity-ledger` | 0 |
| `just check doc-refs` | 0 |
| `just check issue-ids` | 0 |
| `just check prose-issue-refs` | 0 |
| `scripts/check-doc-commit-citations.py` | 0 |
| `just check cpp` | 0 — all cases |
| `cargo check -p nros-node --all-targets --features std` | 0 |

All 17 `api-parity-ledger` shards re-scanned with an `object_pairs_hook` counter after the rebase: **0 duplicate top-level keys, 0 conflict markers.**

`xrce-vendored-versions` is environmental — a control worktree on plain `origin/main` (`b5097a460`) fails it identically (`no vendored tree checked out`).

jerry73204:
Rebased onto `main` (`e802a796f`). BEFORE `38c5cf274` (9 commits) -> AFTER `5ab9665d8` (10 commits, one added below).

**No loss.** `comm -23 <(git log --format=%s origin/main..38c5cf274|sort -u) <(git log --format=%s origin/main..HEAD|sort -u)` is EMPTY — every one of the 9 original commits replayed, nothing dropped.

## Conflicts and how they were resolved

Exactly one commit conflicted, `docs(#427 W4): the 18 ComponentNode ledger rows follow their subject or go`, in three ledger shards. All three are the row-wise/field-wise shape, resolved by merging rather than by taking a side.

**`node.json` — UNION of rows.** `main` had added five `cpp:NodeOptions::*` refuse-loud rows (phase-428 Q1) and rewritten `cpp:NodeOptions::get_rcl_node_options`'s `why`; this branch adds five `cpp:Node::{error_code,error_what,set_error,check_declared_depth}` + `cpp:NodeWithTimers` rows. Disjoint key sets, so both survive; `get_rcl_node_options` takes main's newer `why` (the branch never edited it — it was carried along by the line-based diff). Verified disjoint before merging: each of the five branch keys occurs 0 times in `git show origin/main:docs/reference/api-parity-ledger/node.json`.

**`timer.json` — UNION of rows.** `main` added `c:timer_fini`, `cpp:Timer::cancel`, `cpp:Timer::reset`; the branch adds `cpp:NodeWithTimers::create_timer_in` and `::create_wall_timer_in`. Again 0 occurrences of either branch key on main.

**`pubsub.json` — FIELD-wise compose on one row.** `cpp:Node::create_publisher_in`'s `why` was extended by BOTH sides from the same first paragraph: main appended the phase-428 W5 RANKED-17 "THE GROUP IS DROPPED" finding, the branch appended the phase-427 W4 overload-collision measurement. Taking either side alone loses a true fact, so the merged `why` is `base + W4 paragraph + W5 paragraph` — the shared base sentence asserted identical on both sides before splicing, and the two additions are about different things (what the verb does with the group vs. why the name exists at all).

**Duplicate-key scan.** Git's line-based merge can silently produce duplicate JSON keys (last one wins) even in files it auto-merged without conflict. All 17 shards were re-parsed with an `object_pairs_hook` counter after the rebase: **no duplicates**, all 17 parse.

## One new commit

`fix(rebase): a third gate main added reaches the header W4 deletes`

`check-ledger-orphan-refs` (phase-428 W6) landed on `main` after this branch was cut. W4 deletes `packages/api/nros-cpp/include/nros/component_node.hpp`, which orphaned **16 citations in `param.json` rows that main owns** — 15 bare-basename and 1 repo-relative path. The gate is sentence-scoped (a bare name is excused only by `deleted`/`removed`/`retired` in the SAME sentence; a path has no such escape), so the fix goes inside each citing sentence: the shared RFC-0044 boilerplate (6 rows) and the COLLAPSE-CHECK sentence (3 rows) now record the deletion, and `cpp:Node::set_parameter`'s path citation moves to `packages/api/nros-cpp/include/nros/node.hpp:709-721`, where those declarations actually are on this branch. No verdict, disposition or argument changed.

## Gates

| gate | result |
| --- | --- |
| `just check fast` (290 gates) | **1 FAIL: `xrce-vendored-versions`** — environmental, see below |
| `just check issue-ids` | OK (1220 issues) |
| `python3 scripts/check-doc-commit-citations.py` | OK — 207 citations resolve |
| `python3 scripts/check-ledger-orphan-refs.py` | OK — 17 shards, every citation resolves |

`xrce-vendored-versions` is NOT this branch: it needs `packages/rmw/xrce/xrce-sys/{micro-xrce-dds-client,micro-cdr}` checked out via `nros setup --source`, which a fresh worktree does not have. Reproduced identically on a control worktree at plain `origin/main` (`e802a796f`) — same three lines, same exit 1.

jerry73204:
Superseded by #806 (merged 2026-09-10 20:20 UTC), which carried all 11 of this branch's commits. `git cherry origin/main origin/phase-427-w4-delete-component-node` marks every one of the 11 as `-` (patch already on main), so nothing here is missing from main.

PR #807, opened 2026-09-09
phase-440 W1: one Zephyr workspace resolver, store arm added below, behaviour unchanged

jerry73204:
Implements **phase-440 W1** (RFC-0095 D4). Source-only: shell, `just`, Python and one roadmap doc. Nothing compiled changes.

## What

`scripts/lib/zephyr-workspace.sh` is now the one Zephyr-workspace resolver. `just/zephyr.just`, `scripts/build/west-fixtures.sh` and `scripts/check-tier-preconditions.sh` read it. A `just` expression and a shell function cannot literally share code, so `just` **calls** the helper as a command rather than restating the ladder:

```just
ZEPHYR_WORKSPACE := shell("scripts/lib/zephyr-workspace.sh --version $1 resolve-or-default", NROS_ZEPHYR_VERSION)
```

The `$NROS_STORE/workspaces/zephyr/<version>` arm is added **below** the existing ones, so with no store populated nothing moves; W4 inverts the order. The store root resolves through `${NROS_STORE:-${NROS_HOME:-$HOME/.nros}}` — never an absolute literal, and the gate asserts no path (RFC-0095 D2). The checkout-relative arm stays **last** (RFC-0095 D4 / phase-440 non-goals): a contributor patching a Zephyr module still points `$NROS_ZEPHYR_WORKSPACE` at their own tree.

## The three did not agree, so the fold had to pick an answer

Measured before and after over a 9-shape × 2-line matrix of synthetic hosts, with each pre-change ladder extracted **verbatim from `git show HEAD:`** rather than retyped, and the store deliberately empty:

```
cells: 54   unchanged: 39   changed: 15
OK: no cell on which the three resolvers agreed changed its answer.
```

That claim is checked mechanically, not by eye: every changed cell is one where the three resolvers already gave *different* answers. On this host (nothing provisioned) the resolved path is byte-identical before and after — `just` → `zephyr-workspace`, west-fixtures → `ZEPHYR_BASE` unset, tier-preconditions → empty.

The disagreements were real, and all three are gone:

| | before | |
| --- | --- | --- |
| 4.4 line | `just zephyr` → the 4.4 sibling | `west-fixtures.sh` was version-**blind** → the 3.7 in-tree tree; `check-tier-preconditions.sh` had no 4.4 arm at all, so it reported "no Zephyr workspace" against a tree `just zephyr` was building in |
| a set-but-not-yet-created `$NROS_ZEPHYR_WORKSPACE` | honoured by `just` and tier-preconditions | **fell through** to a different tree in west-fixtures |
| the predicate | `just` tested the *directory* exists | the other two tested `<cand>/zephyr`, so a half-finished `zephyr-workspace/` beat a real sibling |

The reconciliation is stated in the helper's header rather than discovered: the override wins unconditionally (it is the install target too, so it must resolve before the tree exists), a candidate is a workspace when it holds `zephyr/`, and the ladder is version-aware one line at a time. `--absolute` joins `<root>/` with no normalisation, so `ZEPHYR_BASE` keeps the exact string it had.

## The gate

`check-zephyr-workspace-resolvers` (fast lane, 0.37 s) refuses a fourth spelling, in the `check-cli-source-dirs` shape — it names the **direction** of drift both ways: a file that *gained* the ladder is a fourth spelling; a listed file that *lost* it is a stale licence to delete. It keys on the legacy sibling literal `nano-ros-workspace` (not preceded by `-`, which drops the unrelated `nros metadata --nano-ros-workspace` flag) outside comment-only lines; `zephyr-workspace` is a legitimate path constant all over the tree and would be mostly noise.

`.config/zephyr-workspace-resolvers.txt` is the ratchet: 29 entries, one reason each, sorted, and it only shrinks. It records the **26 files W1 did not fold** and why — the CLI's own Rust resolver, the nros-tests harness, four more shell ladders, three Python ones, the `scripts/zephyr/*` legacy-tree rung — because most want W4/W5 first. The task named three call sites; the sweep found the class is 33 files.

Mutation-tested, 15 assertions, all as expected:

| | | |
| --- | --- | --- |
| M1 | a fourth spelling in an unlisted tracked file | RED |
| M2 | **the W1 fold reverted** — the old ladder back in `check-tier-preconditions.sh` | RED |
| M3 | a ratchet line deleted while its file still spells it | RED |
| M4 | a listed file stops spelling it (the shrink direction) | RED |
| M5 | a ratchet entry with no `# reason` | RED |
| M6 | the ratchet unsorted | RED |
| M7 | a comment-only mention | GREEN, by design |
| | every restore | GREEN |

The gate runs its own 18-assertion selftest on the normal path (`check-gate-selftests`), and that selftest is itself a negative control: widening `LADDER` to catch the CLI flag fails 1 assertion, disabling the comment filter fails 4.

## Verification

* `just check fast` — **273 ran, 12 skipped, 0 failed**. Skips are all environmental on this worktree (no in-tree CLI built, no Zephyr workspace, no NVIDIA FSP tree, uninitialised submodules).
* `just ci gate` stops at `check::build`, **9 of 21**, and every one of the nine (`rmw-uorb`, `sched-dim-arms`, `rmw-xrce`, `staticlib-symbols`, `cpp`, `borrowed-e2e`, `test-targets`, `cli-tests`, `workspace-features`) is listed in `.config/ungated-gates.txt`, which states that each **fails in a pristine worktree with no build artifacts**. Causes read as such: a cmake cache generated from the parent checkout's path, and `third-party/dds/cyclonedds` not checked out. None of the nine mentions any file in this diff.
* Tier 1 only — this changes no compiled code.

## Not done here, deliberately

The remaining 26 spellings are recorded in the ratchet rather than folded. Two are worth naming: `packages/cli/nros-cli-core/src/cmd/build.rs` is the resolver a **user** meets (RFC-0095 D2) and has a `--zephyr-workspace` rung the shell chain does not, so it wants its own fold rather than a shell call; `scripts/check-zephyr-workspace-checkout.sh` must *sweep* every provisioned tree rather than resolve one, which is exactly what W5 generalises away.

https://claude.ai/code/session_01Nq9x24szjuBMqdtwhfgxiM

### DIFF
--- a/.config/zephyr-workspace-resolvers.txt
+++ b/.config/zephyr-workspace-resolvers.txt
@@ -1,0 +1,61 @@
+# Files that spell the Zephyr workspace resolution chain themselves, instead of
+# calling `scripts/lib/zephyr-workspace.sh`.
+#
+# phase-440 W1, RFC-0095 D4. The chain is
+#
+#     $NROS_ZEPHYR_WORKSPACE
+#       -> the checkout-relative trees for the selected line
+#            3.7:  zephyr-workspace, ../nano-ros-workspace
+#            4.4:  ../nano-ros-workspace-4.4
+#       -> $NROS_STORE/workspaces/zephyr/<version>
+#
+# and until W1 it had THREE spellings that did not agree: on the 4.4 line
+# `just zephyr` named the 4.4 sibling, `scripts/build/west-fixtures.sh` resolved
+# the 3.7 in-tree workspace, and `scripts/check-tier-preconditions.sh` reported
+# no workspace at all. Those three now read the helper. This file is what stops
+# a FOURTH from landing quietly.
+#
+# It is a RATCHET, like `.config/ungated-gates.txt` and
+# `.config/gate-lane-exempt.txt`: it should only ever SHRINK. An entry is not a
+# blessing — most of these are the same fold W1 did, not yet done for a caller
+# in another language or on another lane. Folding them is phase-440 W4/W5 work.
+#
+# FORMAT: one path per line, sorted, `# reason` REQUIRED on every line.
+# `check-zephyr-workspace-resolvers` enforces the format and BOTH directions of
+# drift: a file that gains the ladder and is not listed here fails; a listed
+# file that no longer spells it fails too, so a line cannot outlive its subject.
+#
+# The gate keys on the LEGACY SIBLING literal (`nano-ros-workspace`, not
+# preceded by `-`), outside comment-only lines: nobody writes that path except
+# to state a fallback rung, whereas `zephyr-workspace` is a legitimate path
+# constant all over the tree.
+
+just/zephyr-dev.just  # the two-line sweep has its OWN override pair (NROS_ZEPHYR_WORKSPACE_37/_44) so it can drive BOTH lines in one run — the helper answers for one line at a time
+justfile  # the nextest deselect guard mirrors west-fixtures.sh's ladder inline; it runs before any `source`, inside a just recipe body. Fold with W4
+packages/cli/nros-cli-core/src/cmd/build.rs  # the CLI's own resolver, in Rust, with a `--zephyr-workspace` flag rung the shell chain does not have. RFC-0095 D2 makes this the resolver a USER meets; it needs its own fold, not a shell call
+packages/testing/nros-tests/src/fixtures/binaries/mod.rs  # Rust; picks the west BUILD root, which is the workspace only when that tree is writable — a different question with a shared first half
+packages/testing/nros-tests/src/zephyr.rs  # Rust; keys on `ZEPHYR_NANO_ROS` and accepts a `zephyr-workspace` SYMLINK, so it is a third rung set again
+packages/testing/nros-tests/tests/fvp_runtime_ws.rs  # Rust test-local copy; folds with zephyr.rs
+packages/testing/nros-tests/tests/fvp_smoke.rs  # Rust test-local copy; folds with zephyr.rs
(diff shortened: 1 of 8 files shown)


jerry73204:
Rebased onto `main`. One conflict, and both sides were right about different things.

`docs/roadmap/phase-440-nros-store-is-the-root.md`, the status line:

- **main:** `W2 CLOSED by deletion (issue 1248); W1 and W3-W8 open.`
- **this branch:** `W1 LANDED; the rest open; W2 was ATTEMPTED and reverted.`

Each side updated the wave it owns. `main` closed W2 — the box-sync gate's lane question, answered by deleting the gate, since `ros2-box-sync.sh` is retired and there is no mirror left to be unfaithful to. This branch **is** W1, so from its own vantage W1 lands.

Not a case of picking a side: the two statements compose, and taking either alone loses a true fact. Merged to

```
**Status (2026-09-09). W1 LANDED; W2 CLOSED by deletion (issue 1248); W3-W8 open.**
```

`main`'s account of W2 is kept verbatim, including the issue reference — it is newer and more specific than "ATTEMPTED and reverted", and §W2 further down already carries the full reasoning.

Verified `W1` is genuinely not on `main` (`git log origin/main --grep='phase-440 W1'` is empty), so "LANDED" is a claim about this PR landing, which is the convention the line already used.

**Verified:** `doc-refs`, `issue-ids`, `prose-issue-refs`, `doc-recipe-refs`, `gate-lists` all OK; `just check fast` 285/287. The two reds are `capability-conditionals` and `xrce-vendored-versions`, which refuse in a fresh worktree for want of submodules and `nros setup --source` trees — neither is reachable from this diff.

1 commit before, 1 after; no-loss check against the recorded remote sha empty.

PR #814, opened 2026-09-09
feat(phase-436 W7): ThreadX and Zephyr's Rust arm park through the seam

jerry73204:
> **Stacked on #648.** Rebase to `main` once it lands. Split from #792, which merged before this commit was pushed — the work landed on a closed PR's branch and needed its own.

Both entries hold a typed `Executor`, so unlike the C arms they needed no new FFI — only the primitive and the wiring.

## One place for the constraint

`port_park::install_port_park` holds the non-obvious part instead of it being spelt per board: **a `ParkUntilFn` must wait on the executor's own wake object.** The backend's listener signals that one, and `spin_once` drops its transport drain to non-blocking once a port has parked — so a park on any other object cannot be broken by data arriving and sleeps out its full deadline with a message already waiting. Worse than not parking, invisible in a quiet run, green on every test.

## ThreadX gains the primitive and buys no resolution — that is the point

`tx_semaphore_get` counts ticks and `TX_TIMER_TICKS_PER_SECOND` is **100** on both shipped boards, so the real floor is **10 ms** — an order of magnitude coarser than its ABI signature implies. Before #1242 the executor assumed 1 ms for every port and W3's jitter report inherited the over-claim. Installing this makes the executor stop *claiming* precision ThreadX never had.

Conversion rounds **up**: a wait shorter than asked is a missed deadline; a longer one is merely late.

## Four compile problems, three the same shape

| | |
|---|---|
| `::nros::executor` doesn't exist | the facade never re-exported it → now `::nros::port_park` |
| my Zephyr insertion **split a `#[cfg(feature = "zephyr-edf")]` pair** of `apply_tier_deadline` | same class as the W7.a insertion that stole `nros_board_native_run_tiers`'s gate |
| `port_park`'s import was ungated while its function is gated | `-D warnings` rejected it — issue 0487's class |
| ThreadX's install belonged in the shared `apply_tier` | the separate boot-site call was a duplicate |

## And a green check that was not evidence

`cargo check` on the Zephyr board **passed while never compiling `entry_tiers.rs`** — it sits behind `#[cfg(feature = "tiers")]`. The real errors appeared only under `--features tiers`.

ThreadX cannot be checked standalone at all: a **pre-existing** cyclic dependency with `nros-board-threadx-linux`, verified present on an unmodified tree, so it is checked through that dependent.

## Verification

384 passed, 0 failed. `nros-board-zephyr` clean with and without `tiers`; `nros-board-threadx` clean via `nros-board-threadx-linux`.

Bare metal stays unwired — no board in the tree can arm a one-shot deadline compare, which is a hardware-code gap rather than a wiring one.

https://claude.ai/code/session_01XPKwuBESsmHcbgkCN3odw4

### DIFF
--- a/packages/api/nros/src/lib.rs
+++ b/packages/api/nros/src/lib.rs
@@ -1290,6 +1290,11 @@ pub use nros_node::{
 // type a user could name and never receive. RFC-0036's Errors row now describes
 // what these two are.
 // ---------------------------------------------------------------------------
+/// phase-436 W7 — installing a port's park primitive, with the wake-object
+/// constraint stated once. Board entries reach it through this facade because
+/// that is the dependency they already carry.
+pub use nros_node::executor::port_park;
+
 pub use nros_node::NodeError;
 pub use nros_rmw::TransportError;
 
--- a/packages/boards/nros-board-threadx/src/entry.rs
+++ b/packages/boards/nros-board-threadx/src/entry.rs
@@ -96,6 +96,30 @@ static UART_LOGGER: UartLogger = UartLogger;
 /// Install the UART `log` sink, routing records through `B::println`. Idempotent:
 /// re-arms the print fn each call and ignores a repeated `set_logger` (the second
 /// returns `Err`). Safe to call once per boot before the spin loop.
+
+/// phase-436 W7 — ThreadX's park, and the tick period that is its real floor.
+///
+/// ThreadX has no sub-tick wait: `tx_semaphore_get` counts ticks and
+/// `TX_TIMER_TICKS_PER_SECOND` is 100 on both shipped boards, so the floor is
+/// 10 ms — an order of magnitude coarser than the millisecond the ABI
+/// signature suggests. Installing this does not gain resolution; it makes the
+/// executor stop CLAIMING resolution it never had (issue 1242).
+unsafe extern "C" {
+    fn nros_platform_wake_park_until_us(w: *mut core::ffi::c_void, deadline_us: u64) -> i8;
+    fn nros_platform_wake_park_granularity_us() -> u64;
+}
+
+/// Install the park on `crt`'s executor. Advisory: a build with no wake object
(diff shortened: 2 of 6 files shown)


jerry73204:
## No retarget needed — this already landed on its base

Came here to unstack this onto `main` (a stacked PR cannot arm auto-merge: auto-merge is a property of the BASE branch, and `fix/spin-nominal-is-the-period` has neither the ruleset nor the queue). It was already merged before I got here — squash-merged into `fix/spin-nominal-is-the-period` at 2026-09-09T03:09:50Z as `6911afb0e`, by @jerry73204. Nothing to rebase, nothing to retarget.

**Patch-id check**, since the carried commits have to drop cleanly when #648 rebase-merges into `main`:

```
$ git show b005d6c66 | git patch-id --stable   # this PR's commit
486352d2dcfb882a0f0737907e549a82ede76eae b005d6c66…
$ git show 6911afb0e | git patch-id --stable   # #648's squash of it
486352d2dcfb882a0f0737907e549a82ede76eae 6911afb0e…

$ git diff b005d6c66 6911afb0e --stat
(empty)
```

Identical patch-id and identical tree, so this drops by patch-id on #648's next rebase with no intervention. #648 is open against `main`, MERGEABLE, auto-merge armed since 2026-09-08T16:39Z — this ships inside it.

**One thing worth a look, and I have deliberately not touched it.** After the merge, `feat/phase-436-w7-rust-arms` gained a commit that is in no open PR:

```
01c3602d9  2026-09-09 11:27:36 +0800  fix(phase-436 W7): mirror the optional park symbols through the platform ABI
```

`git merge-base --is-ancestor origin/feat/phase-436-w7-rust-arms origin/fix/spin-nominal-is-the-period` is false — that commit is not on #648's branch or on `main`, and this PR is closed, so nothing is currently carrying it to `main`. If it is live work, it needs a PR of its own (or a push onto #648's branch) or it strands here.

PR #819, opened 2026-09-09
feat(phase-440 W3): third-party/ holds tracked submodules and nothing else

jerry73204:
`third-party/` carried two kinds of thing at once: 20 tracked submodules whose
pins are **decisions** (cyclonedds tracks the version ROS ships — moving it is
how you stop interoperating), and gitignored provisioning that `nros setup`
writes. "Pinned source or provisioning?" then had no answer from the path, and
every consumer walking the directory encoded the distinction itself.

That is the shape that made the box mirror unfixable one directory over (#1248):
rules telling source from build output **by name**, where the two live together.

## Measured first — most of it was already done

`make/` and `ninja/` provision into the store and their consumers already read
it there. Both said so in the **past tense**:

- `jobserver-pool.sh` — *"the store rather than `third-party/make/`"*
- `check-tier-preconditions.sh` — *"this block used to read `third-party/make/make`"*

What remained was 1.2 MB of residue and two `.gitignore` lines that made the
directory look like it still held provisioning. Removed.

**`external/` is provisioned by nothing and consumed by nothing** — every
apparent consumer is NuttX's own `apps/external/` or prose. 287 MB of debris.
Its `.gitignore` entry **stays**: deleting another operator's untracked data is
not this phase's call, and the gate covers `third-party/` only.

## The ratchet, and two things it deliberately does not do

`check-third-party-is-submodules` reads submodule parents from `.gitmodules`,
**never a directory walk** — so an uninitialised submodule still counts, because
a bare clone has empty dirs. And an unreadable `.gitmodules` exits **2** rather
than reporting OK over a reading that found nothing.

## One declared exception: `ros`

`[source.rosidl]` has `dest = "third-party/ros/rosidl"`, and
`msg_to_cyclone_idl.py` resolves it as its last ladder rung so the cyclone
msg→IDL step works with **no ROS install**. Moving it needs `dest` to be able to
name the store — RFC-0095 D2/D4, which is W4. So it is declared with a reason
and somewhere to go, rather than the rule being weakened to fit it. The list may
only shrink, and a stale entry says so.

## What testing corrected

The first selftest **passed with the fix removed.** It exercised `nros_grep_q`
in isolation while the loop had the predicate inlined, so mutating the loop to a
substring match — where a *prefix* of a parent would wrongly pass — left it
green. There is now one `is_parent` / `is_exception` that both the loop and the
selftest call, and the same mutation now fails it:

```
selftest FAIL: a PREFIX of a parent is not a parent (want 1, got 0)   rc=2
```

Two other gates caught this script before it could land, both correctly: it used
bare `grep -q` (which cannot tell "no match" from "grep failed" — issue 0726,
and the two spellings fail in opposite directions), and its selftest was
flag-only rather than on the normal path.

Mutations: a new provisioning root under `third-party/` → RED naming it and its
size; `.gitmodules` unreadable → rc 2.

`check-fast`: 286 gates ran, 1 skipped.

https://claude.ai/code/session_012ZwZdvRHvXxozmr8KPsv82

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,8 +71,6 @@ tools/*
 /.claude/*
 !/.claude/skills/
 /external/
-/third-party/make/
-/third-party/ninja/
 # phase-362 — the vendored zenoh router is retired; the router comes from ROS
 # (`rmw_zenoh_cpp/rmw_zenohd`). An existing checkout still has the submodule's
 # working tree on disk after the gitlink is dropped, and leaving it untracked
--- a/docs/roadmap/phase-440-nros-store-is-the-root.md
+++ b/docs/roadmap/phase-440-nros-store-is-the-root.md
@@ -86,15 +86,35 @@ gone; `ros2-box-env.sh` is 48 lines from 260 and sets only the store split;
 `check-gate-lists` 283 fast / 21 build-serial and `check-gate-visibility` 21
 acknowledged both green. Issue #0925 closed as moot.
 
-### W3 — `third-party/` holds tracked submodules and nothing else
+### W3 — `third-party/` holds tracked submodules and nothing else — **LANDED, one exception declared**
 
-Move `make/`, `ninja/`, `ros/` to the store; retire `external/` into
-`$NROS_STORE/fetch/`; **delete** the 23 M `external/PX4-Autopilot`, a stray
-duplicate of the 68 G tracked submodule at `third-party/px4/PX4-Autopilot`.
+Measured rather than assumed, and most of it was already done: `make/` and
+`ninja/` provision into the store and their consumers already read it there
+("the store rather than `third-party/make/`", `jobserver-pool.sh`). What
+remained was 1.2 MB of residue and two `.gitignore` lines that made the
+directory look like it still held provisioning.
 
-*Acceptance:* every path under `third-party/` is a tracked submodule or tracked
-content — asserted by a gate, so a fourth provisioning root cannot land there
-quietly. `.gitignore` loses the corresponding entries rather than gaining any.
+`external/` turned out to be provisioned by NOTHING and consumed by NOTHING —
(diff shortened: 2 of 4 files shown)


jerry73204:
Rebased onto `main`. The `check-submodule-pins` red was a **rewind, and an inherited one** — nothing in this PR authored it.

```
submodule-pins: packages/cli/third-party/play_launch
    REWIND — the new pin is an ANCESTOR of the old one
      was: 0647131a7cc1  test(experiments): four clean clone-vm runs, and the gap that remains
      now: db4af878230f  feat(resolve): honour `external:` on a service, …
```

Measured before doing anything: `git log origin/main..<branch> -- packages/cli/third-party/play_launch` is **empty** — no commit here touches that submodule. The branch simply sat on a `main` older than the one that advanced the pin, so its tree carried the older gitlink and the gate compared it against today's baseline.

That is the case CLAUDE.md calls out — *"a rewind is usually INHERITED from a rebase rather than authored, which is exactly the window a pre-commit check misses"* — and it is why the remedy is a rebase, not `NROS_ALLOW_SUBMODULE_REWIND=1`. There is nothing deliberate here to justify.

After the rebase the tree carries `0647131a7cc1`, and the gate agrees:

```
submodule-pins: OK (0 pin(s) moved — 0 verified fast-forward)
```

1 commit before, 1 after; no-loss check empty.

## One red I did not take at face value

`just check fast` also reported `subtree-guard` — *"FAIL: launcher never wrote its lock file"*. It passes on plain `origin/main`, and re-run **solo** on this branch it passes too (`all 3 paths OK (trap, refuse, reap)`). It is timing-sensitive and I had three parallel agents loading the host at the time, which is the repo's own "retest a red SOLO before filing" case. Not this PR's.

The other two, `capability-conditionals` and `xrce-vendored-versions`, refuse in a fresh worktree for want of submodules and `nros setup --source` trees.

PR #820, opened 2026-09-09
docs(phase-412): checkpoint — what landed, what is queued, what is left

jerry73204:
Docs only. The campaign's state was recoverable only by reading seven pull
requests, so it is written down in the phase doc: what is on `main`, what is in
the merge queue (with what each PR closes), and what is left in the order it
should be taken.

The remaining four items, for review:

1. **Producerless knobs (#6)** — 13 build-script env names plus the C-side
   `NROS_ZEPHYR_MAX_TIERS` / `NROS_ZEPHYR_TIER_STACK_SIZE`. Delete or wire a
   producer; a knob nothing sets reads exactly like one that is honoured.
2. **Five sizing knobs off the RFC-0049 ladder** — `NROS_ENTRY_SPIN_MS`,
   `NROS_SERVICE_TIMEOUT_MS`, `NROS_XRCE_TRANSPORT_MTU`,
   `ZPICO_GRAPH_CACHE_SIZE`, `ZPICO_READ_TASK_PRIORITY` /
   `ZPICO_LEASE_TASK_PRIORITY`. Issue 0623's caveat is called out: the two
   priorities are RAW FreeRTOS units sharing a scheduler with the tier
   priorities, so normalising them is the defect and not the fix.
3. **`NROS_MAX_LIVELINESS`** — same shape as 769's graph cache; unblocked by
   PR 749.
4. **The arena's host oracle** — inputs now reach the derivation on all three
   roads, so what is missing is the build-time check that the derived arena
   covers the model. Only the Zephyr `.bss` backing has that today.

Ran: `just check markdown-links`, `just check doc-refs`.

### DIFF
--- a/docs/roadmap/phase-412-derived-counts-and-sizes.md
+++ b/docs/roadmap/phase-412-derived-counts-and-sizes.md
@@ -594,3 +594,55 @@ holds the evidence, the item is *close it*.
 | [#1142](../issues/1142-nuttx-standalone-example-has-no-entity-declaration.md) | a standalone copy-out example reaches none of the three declaration channels, so the guessed budget is the only one it can get. This is W5's open design question in its most concrete form; answering W5 closes it |
 | [#1181](../issues/archived/1181-stale-zpico-subscriber-buffer-size-name.md) | `ZPICO_SUBSCRIBER_BUFFER_SIZE` is documented in four places and read by nothing. A knob that sizes nothing is deleted from the docs, not derived — the census is this phase's |
 
+
+## CHECKPOINT 2026-09-09 — what is done, what is in the queue, what is left
+
+Written as a checkpoint because this campaign has crossed several sessions and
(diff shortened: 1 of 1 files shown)


jerry73204:
Superseded by #962, which cherry-picks this commit and adds a 2026-09-11 checkpoint on top.

This PR has been CLEAN but **never armed** since 2026-09-10, which is why it did not land — `gh pr merge --auto` was evidently never run, and only `autoMergeRequest` proves anything armed.

Its content is preserved verbatim: #962 carries `a80c12eec` by cherry-pick rather than stacking (a stacked base cannot arm auto-merge, and the queue drops the duplicate by patch-id). The added checkpoint records that two of this PR's four remaining items have since landed (`NROS_MAX_LIVELINESS`, the arena host oracle) and that a third grew from five knobs to seven.

Safe to close in favour of #962, or arm this one and #962 will drop to a single commit on its next rebase.

PR #860, opened 2026-09-10
feat(phase-443 W3+W4, RFC-0097): the launcher is its own artifact, and CI does not choose a toolchain nobody named

jerry73204:
Finishes **phase-443 W3 + W4** (RFC-0097 D4 / D11). Resumes an earlier session's
WIP commit, which was cut off mid-test-edit; that commit is squashed into this
one, verified.

## W3 — the launcher is its own artifact

`dispatch.rs` was a module inside the binary that carries the toolchain, so the
thing that SELECTS a version shipped with the version being selected.
`packages/cli` takes **710 commits per 60 days**; a launcher sharing that
artifact re-releases on every one of them while carrying no compatibility
promise of its own.

`dispatch.rs` and `pin.rs` **move** into a new `packages/cli/nros-launcher`
crate, joined by `store_root.rs` (was `orchestration::store::root`) and
`checkout.rs` (was `abi_guard::find_monorepo_root`). `nros-cli-core` re-exports
all four at their original paths, so there is exactly ONE parser of
`nros-toolchain.toml`, one answer to "where is the store", and one definition of
"is this a checkout".

`launch.rs` is the new part: a launcher meets three states a fronted toolchain
cannot — an empty store, an unpinned project with no CLI to fall back to, and a
cwd inside a checkout. `dispatch::decide` answers `Proceed` to all three.

The crate keeps its own `version`, deliberately not `version.workspace = true`.

### Acceptance, measured

| clause | how |
| --- | --- |
| runs with NO toolchain installed | `the_launcher_runs_with_no_toolchain_installed_and_says_what_to_run` — starts, names `install.sh` and the store, exits non-zero |
| a toolchain update does not replace it | launcher binary compared byte for byte across a store gaining a newer toolchain, and the same launcher then hands over to it |
| `nros --version` distinguishes the two | `nros-launcher 0.1.0` then the toolchain's line; `--version` is still FORWARDED (`ARG=--version`), and answering it exits 0 even on an empty store |
| does not link the CLI's dependency graph | `launcher_dependency_closure.rs` walks `packages/cli/Cargo.lock` from the launcher's own `[dependencies]`: **27 crates against `nros-cli-core`'s 192** |

## W4 — CI does not choose, or write, a toolchain nobody named

`Session` answers "is this automated?" once, from `CI` / `GITHUB_ACTIONS` (a
declaration beats an `isatty` inference), with `NROS_ALLOW_PIN_WRITE_IN_CI` as
the escape hatch that BEATS the declaration.

The rule lands in two places because they know different things:

- **`nros build` REFUSES** — `PinOutcome::RefusedInCi` becomes an `Err`. It
  knows it is about to write into the user's source tree.
- **the launcher WARNS** — `Source::DefaultInCi`, naming the version it took. It
  parses no argv (D8), so refusing there would refuse `nros --version`,
  `nros setup --list` and `nros doctor` on every runner.

The refusal is asked AFTER the three arms that write nothing anyway (checkout,
already-pinned, no store version). Ordering it first would fail every build in
this repository's own CI, which builds inside a checkout.

It does not name `nros pin <version>` as the phase doc asked: no such verb
exists yet. It prints the exact `nros-toolchain.toml` to commit instead. The
phase doc records the correction.

## Mutation pass — 25 mutations, 25 RED, 0 GREEN

Every fix was removed in turn and the named test noticed. Two fixes had **no
test** when this was picked up and gained one:

- `nros build` turning `RefusedInCi` into an error — extracted as
  `report_pin_outcome`, so D11's build half is reachable without a real store;
- the dependency closure — the "measured, not asserted" clause had nothing
  measuring it.

## Also here

`nros-cli`'s `an_unpinned_project_runs_in_the_launcher` failed **1 of 4** full
`packages/cli` runs and passed **4 of 4** solo — issue 0476's ETXTBSY race, in
the launcher dispatch test, exactly where that issue predicted.
`nros_launcher::test_support` is the one spelling of "install an executable you
are about to run" that every launcher test can reach (`nros-cli-core` depends on
`nros-launcher`, so a helper there would be a cycle).

## Not done here

`scripts/install.sh` still fronts `<store>/bin/nros` at the newest TOOLCHAIN, so
nothing installs the launcher yet. Fronting it means the release asset carries
both binaries and `sdk-front` learns which one it fronts — a change to
`release-nros.yml`, which is **W2's file (#859)**. Recorded in the phase doc.

`#859` also adds `pin::set` to `orchestration/pin.rs`, the file this PR moves.
That is the one conflict between the two, and it is mechanical: `set` goes into
`nros-launcher/src/pin.rs`.

## Verified locally

- `just check fast` — 290 gates green, 2 skipped (unprovisioned trees)
- `just check cli-tests` — the whole `packages/cli` workspace, green
- `just check cli-clippy`, `just check submodule-commits-reachable`,
  `just check cli-source-dirs` — green
- `just setup-cli` still builds, and the ownership guard still behaves on both
  arms (silent outside a checkout, 6 `stale_guard` + 6 `abi_guard` tests green)

Not run: `check-compile-smoke` / `check-workspace-all`. Both compile the ROOT
workspace; this diff touches only `packages/cli` and one roadmap doc.

### DIFF
--- a/docs/roadmap/phase-443-release-composition.md
+++ b/docs/roadmap/phase-443-release-composition.md
@@ -74,6 +74,30 @@ launcher does not link the CLI's dependency graph (measured, not asserted); a
 contributor inside a checkout is unaffected, because dispatch declines at its
 own seam.
 
+**Landed.** `packages/cli/nros-launcher` — `dispatch.rs` and `pin.rs` MOVED out
+of `nros-cli-core` (which re-exports them at their old paths, so there is one
+parser of `nros-toolchain.toml`), plus `launch.rs` for the three states a
+launcher meets and a fronted toolchain cannot: an empty store, an unpinned
+project with no CLI to fall back to, and a cwd inside a checkout. The crate
+carries its OWN `version`, which is the point — the workspace version is the
+toolchain's, and sharing it is the coupling the split exists to break.
+
+Measured, not asserted: `tests/launcher_dependency_closure.rs` walks the
+launcher's link closure out of `packages/cli/Cargo.lock` — **27 crates against
+the toolchain's 192**, with `nros-cli-core`, `clap`, `minijinja`,
+`nros-pkg-index`, `nros-launch-parser` and `nros-entry-lower` named as forbidden
+and a ceiling that a new direct dependency trips.
+
+**Not done here, and it is the half a user meets:** `scripts/install.sh` still
+fronts `<store>/bin/nros` at the newest TOOLCHAIN's `bin/nros` through
+`sdk-front`, so nothing installs the launcher yet. Fronting it instead means the
+release asset must carry both binaries and `sdk-front` must learn which one it
+fronts — a change to `release-nros.yml`, which is W2's file. It belongs with
+W2's manifest, not ahead of it. Until then the launcher is a built, tested
+artifact that no install path delivers, and the dispatch a user actually gets is
+still W7's in-toolchain one (`nros-cli/tests/toolchain_dispatch.rs`, which
+still passes).
+
 ### W4 — CI refuses to write a pin (RFC-0097 D11)
 
 Pin-on-first-build is right interactively and wrong in CI: silently pinning to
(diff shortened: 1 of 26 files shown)


jerry73204:
Resumed after the authoring agent was cut off mid-rebase. What changed since its last push:

**Rebased onto current `main`**, which had gained #830, #852 and #859. Two conflicts, both real:

- `nros-cli/Cargo.toml` + `tests/toolchain_dispatch.rs` — #830 landed the **same issue-0476 ETXTBSY fix** independently, in `nros-cli-core`. Both diagnoses were right; the placement question was the interesting part, and **this branch's is better on layering**: after W3, `nros-cli-core` *depends on* `nros-launcher`, so a helper in cli-core is a dev-dependency cycle for the launcher's own tests — which are exactly the tests that hit the race. Took this branch's `nros_launcher::test_support`, and **retired #830's `test-support` feature and `pub` module** so there is one spelling rather than two. cli-core now re-exports the launcher's `write_executable_stub` for its own use; `copy_executable` exists once.
- W2's `pin::set` (merged as #859) carried forward into the moved `nros-launcher/src/pin.rs`.

**Mutation pass** — the step the authoring agent never reached. Seven mutations, every one caught:

| fix broken | test that went red |
|---|---|
| CI is never detected | `either_conventional_variable_declares_an_automated_session` |
| CI never marks the choice as nobody's | `an_unpinned_project_in_ci_is_warned_about_through_the_real_binary` |
| the CI warning is never emitted | same |
| a checkout no longer makes the launcher stand aside | `a_checkout_makes_the_launcher_stand_aside` |
| an unanswerable pin falls back to the newest installed | `an_unanswerable_pin_does_not_fall_back_to_the_newest_installed` |
| the empty-store report stops naming the installer | `the_launcher_runs_with_no_toolchain_installed_and_says_what_to_run` |
| a half-removed toolchain (no `bin/nros`) counts as installed | `both_layouts_count_and_an_empty_directory_does_not` |

Three earlier attempts came back green and **none was a test defect** — they were equivalent mutants, my error. Worth recording, because "survived" and "not covered" are not the same claim: routing an empty store through a fabricated `0.0.0` still reaches `Plan::NothingInstalled` via the `installed_bin` `None` arm; prefixing the store path still satisfies `contains`; and an extra search directory with no `bin/nros` is filtered anyway. Each was re-run in a sharper form and went red.

`just check fast` 291/291 (2 env-skipped) · 1047 cli-core lib tests · 10 launcher/cli test binaries.

PR #862, opened 2026-09-10
docs(phase-445): W1 is proven on a leaf that HAS a model; the esp32 talker's hand-set value goes in W3

jerry73204:
Follow-up to #861, which was already at queue position 1 (its branch refuses pushes while queued). Carries #861's commit so this can target `main`; the queue drops it by patch-id once #861 lands.

W1's acceptance named the esp32 **talker** — a single-package leaf that has no model until W3 gives it a `system.toml`, so W1 could not have met it. Corrected:
- W1 proves itself on a **workspace entry** leaf: a plain `cargo build` with no `NROS_DECLARED_*` in the process env yields the shim constant the fixture lane gets by exporting them.
- The consumer (`nros-zpico-build/src/runner.rs`) already computes the count from `NROS_DECLARED_SERVICE_SERVERS` + `NROS_DECLARED_INFRA_QUERYABLES` (+ `NROS_DECLARED_NODES`); W1 carries those facts on the sidecar road and never states `ZPICO_MAX_QUERYABLES` itself (issue 0460).
- Removing the esp32 leaves' hand-set `ZPICO_MAX_QUERYABLES` moves to **W3**.

Docs only; `check-roadmap-status` / `check-roadmap-claims` OK.


jerry73204:
Superseded by #864: measuring W1 showed a workspace member's sidecar is the workspace root's (shared by every image), so the acceptance written here could not hold. #864 carries this PR's commit plus the correction.

PR #870, opened 2026-09-10
docs(#1264, #1266, #1267, #1273, #1274, #1275): what provisioning a runner from scratch measured

jerry73204:
Six issues, all found by provisioning a contained self-hosted runner from an
empty store — which is the first environment in a long time to ask the
provisioning system every question honestly, with no warm caches and no
packages a developer laptop happens to have.

**Docs only.** No code in this PR. It is split out of #842 (the runner work) so
the findings reach `main` — and therefore `just issues` — while that PR is still
in flight, rather than sitting on a branch nobody queries.

| id | what |
| --- | --- |
| 1264 | `nros_platform_rmws` discards the manifest parser's stderr, so a python without `tomllib`/`tomli` is reported as a platform that does not exist — about a name present in that manifest 42 times |
| 1266 | fetch, verify and unpack are strictly serial, and `curl --silent` makes a 1.4 GB download look hung: 15 minutes of no output while the store grew the whole time |
| 1267 | the package loop is sequential in the CLI, so no caller can parallelise it; the prefixes are disjoint by construction, but the lock file, `front_newest`'s issue-0500 ordering, `bin_dirs`' PATH order and interleaved output each need a decision |
| 1273 | a tool builds from source because the index has no `dist.<host>` row, not because it must be built — index and release work, not CLI code |
| 1274 | the apt ask IS batched within one `nros setup` call and not across a session: one bootstrap printed it three times with three different subsets, and acted on none |
| 1275 | the west allowlist pulls 2.5 GB of vendor HALs (`hal_nxp` 1.3 G, `hal_stm32` 764 M, `hal_espressif` 275 M, `hal_nordic` 224 M) that no board in `fixtures.toml` needs |

Each records what was DECIDED, not only what was found, so the next person does
not re-litigate it:

* **cargo-binstall: no, for now.** It resolves at run time; the index's contract
  is a pinned `url` + `sha256`, verified before unpack, recorded in
  `nros-sdk.lock`. It would also be a third acquisition path beside `dist` and
  `source` — which `check-one-producer-per-tool` refused an apt-installed
  `make`/`ninja` for in #842. If it returns, it belongs maintainer-side, minting
  dist rows offline.
* **`play_launch_parser`: we build and publish the dist ourselves** into
  `nano-ros-sdk`. play_launch has its own release cycle; waiting on an upstream
  asset for a repo we own is waiting on our own calendar.
* **espflash and the other source-only tools: point at the upstream asset**, the
  pattern `[tool.zephyr-sdk]` already documents. Survey them all rather than
  fixing the two this bootstrap happened to hit.
* **Batching beyond apt** (`rustup target add`, repeated `--source`, re-resolved
  tools) is a survey for whoever takes 1274, not a guess made here.
* **The espressif HAL is held back from the allowlist change on purpose.** esp32
  IS a platform here, but it builds through ESP-IDF. Whether Zephyr's own
  espressif toolchain makes that a duplicate is a platform-strategy question
  worth measuring, not one to settle while deleting a manifest line.

Anyone working on the provisioning system: 1273, 1274 and 1275 are the
actionable ones, and 1273/1274 reduce what there is to parallelise before anyone
writes a worker pool for 1267.

### DIFF
--- a/docs/issues/1264-platform-rmws-swallows-parser-stderr.md
+++ b/docs/issues/1264-platform-rmws-swallows-parser-stderr.md
@@ -1,0 +1,95 @@
+---
+id: 1264
+title: "`nros_platform_rmws` redirects the manifest parser's stderr to
+  /dev/null, so an interpreter that cannot read `fixtures.toml` is reported
+  as a platform that does not exist"
+status: open
+type: bug
+area: build, testing
+severity: medium
+found: 2026-09-10
+related: [phase-395]
+---
+
+## What this is
+
+`scripts/build/platform-rmws.sh`:
+
+```sh
+out="$(python3 "$root/scripts/build/fixtures-manifest.py" coords 2>/dev/null \
+    | awk -F'\x1f' -v p="$platform" '$2 == p { print $4 }' \
+    | grep -v '^$' \
+    | sort -u)"
+
+if [ -z "$out" ]; then
+    echo "nros_platform_rmws: no fixture rows for platform '$platform'" >&2
+    echo "  (known platforms come from examples/fixtures.toml)" >&2
+    return 1
+fi
+```
+
+Empty output has two causes and the function reports one of them. The parser
+printing nothing because the platform genuinely has no rows is the case the
+message describes. The parser DYING is the case it does not: `2>/dev/null`
+discards the traceback, the pipeline's status is `sort`'s, and `$out` is empty
+either way.
+
+## How it was found
+
+Provisioning a contained self-hosted runner on Ubuntu 22.04:
+
+```
+nros_platform_rmws: no fixture rows for platform 'threadx-riscv64'
+  (known platforms come from examples/fixtures.toml)
+error: recipe `setup` failed with exit code 1
+```
+
+`threadx-riscv64` appears in `examples/fixtures.toml` 42 times, with four
+`platform = "threadx-riscv64"` rows. The real cause was three layers away:
+Ubuntu 22.04 ships python3.10, which has no `tomllib`; `fixtures-manifest.py`
+falls back to `tomli`, which was not installed either; so it raised
+`ModuleNotFoundError` before reading a byte of the manifest.
+
+Running the same command by hand — without the redirect — says so in one line.
+The diagnosis cost far more than that, because the message was specific,
+confident, and about the wrong subject. It aims the reader at the manifest and
+at the platform NAME, which are both fine.
+
+## Why it matters beyond one host
+
+The message is load-bearing by design. The file says so:
+
+```
+# Deliberately NOT silent on failure: a platform with no rows is a caller error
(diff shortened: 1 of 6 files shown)


jerry73204:
Read all six. Two connections to phase-443 that are worth having before anyone picks these up.

## 1273's remedy is index work, but today it still costs a CLI release — #854 is what changes that

1273 says, correctly, *"No CLI change is needed for either. This is index and release work."* That is true of the CLI's **logic** — `plan_install` already prefers a dist — but it is currently false of the **release mechanics**, and that is the part that decides whether adding a dist row is cheap.

On `main` the index ships *inside* the release asset:

```rust
pub(crate) fn shipped_index() -> Option<PathBuf> {
    let exe = std::env::current_exe().ok()?;
    let prefix = exe.parent()?.parent()?;      // <prefix>/bin/nros -> <prefix>
    let candidate = prefix.join("share/nros/nros-sdk-index.toml");
    candidate.is_file().then_some(candidate)
}
```

and `release-nros.yml:208` does `cp nros-sdk-index.toml "$stage/share/nros/…"`. So a one-line `dist.linux-x86_64` row reaches users only by cutting a new `nros`. That is exactly the ratio RFC-0097 measured — the index moves **70 times per 60 days** against `NROS_CODEGEN_VERSION`'s **2 all time** — and it is why phase-443 W1 exists.

**#854** (queued, position 2) makes the index fetched and cached under `<store>/fetch/`, with the shipped copy kept only as the offline fallback. After it lands, "survey the source-only tools and add a dist row each" is a pure index change that an older CLI picks up on its next fetch. Before it lands, the survey is real work whose delivery is gated on a release.

Not a request to change anything in this PR — 1273's decisions all stand as written. Just: the sequencing is 854 → 1273, and 1273's cost estimate improves a lot once it is.

## 1266: there are now three `curl` sites, not two

`sdk_store.rs:507` and `:615` are the ones a 1.4 GB download goes through. W1 added a third in `cmd/setup.rs::fetch_index`, also `--silent`. It is **not** the same defect — it fetches a TOML of tens of KB with `--connect-timeout 10 --max-time 60`, because an offline host must fail in seconds on the way to a provisioning step rather than hang — so progress reporting would be noise there.

Flagging it only so whoever takes 1266 greps for three and decides deliberately, rather than fixing two and finding the third later. If that work produces a shared downloader, the index fetch has two requirements the tool path does not: fail-fast timeouts, and `--proto =https --tlsv1.2` applied *only* to the default URL (an operator's own `file://` mirror is one they picked — same rule `scripts/install.sh` already follows).

## On 1275

The "prove it rather than assume it" instruction is the right call and I would not skip it: `native_sim` is 72 of 76 Zephyr fixture rows, but `cmake/zephyr/*.conf` merges last-wins and a board fragment can pull a module with no fixture naming the board — issue 0876 was exactly that shape, a conf whose real effect was somewhere nobody was looking.

PR #878, opened 2026-09-10
feat(phase-446 W6): the code is checked against the contract's params at boot

jerry73204:
## Summary

Phase 446 W6: the code is checked against the contract's `params:` at boot. If a node declares a parameter its contract does not declare, or declares it with a different type, the boot is refused. The refusal names the node, the parameter and the contract. This follows the declared-QoS machinery (`nros_declared_qos_generated.h`, `check-declared-qos-header`, `check_declared_depth` -> `DECLARED_DEPTH_MISMATCH`).

**This PR is based on `main` and carries #872's pin-bump commit** (ros-launch-manifest v0.1.34, play_launch 155ed78b), cherry-picked, because #872 has not merged yet. That is the owner's practice. Once #872 merges, the duplicate commit drops by patch-id on the queue's rebase. The pin stays at 155ed78b. The coordinator will bump it forward to f5264bf0 (which loosens the W2 check for `/**` wildcard values) in a separate change.

### One derivation, both languages

`nros_orchestration_ir::declared_params` is the one derivation of which nodes declare, what each name is declared as, and which contract a refusal names. Both paths read it:

- **C++:** `nros ws entity-inventory --output-params-header` renders `<nros/nros_declared_params_generated.h>`. The same deferred render in `NanoRosNodeRegister.cmake` writes it next to the declared-QoS table, in the same PRIVATE include dir. `nros/declared_params.hpp` expands it. `ComponentNode::declare_parameter` (scalar and vector) looks up the node's FQN. On a mismatch it refuses through `set_error` with `DECLARED_PARAM_MISMATCH` (-446), after printing:
  ```
  [nros] FATAL: node "/system/mrm_handler": parameter "timeout" is not declared in its contract
  (island.contract.yaml, `params:`). The parameter store is sized from that declaration, so a
  parameter it does not name has no slot. Declare it there with its type, or stop declaring it in the code.

  [nros] FATAL: node "/system/mrm_handler": parameter "update_rate" is declared `integer` in its
  contract (island.contract.yaml) but the code declares it as `double`. Fix the contract or the
  code so they state one type.
  ```
- **Rust:** `nros::main!` hands the table to the executor (`RuntimeCtx::apply_declared_params` -> `Executor::set_declared_params`) before any node registers. The node runtime's parameter arm checks each declaration with `nros_params::DeclaredParams::check`, logs the same two messages through `nros_log`, and refuses registration.

### What is not checked

- **Nodes with no `params:`.** A node whose contract has no `params:` is not checked.
- **Names play_launch exempts.** `use_sim_time`, `start_type_description_service` and `qos_overrides.*` are never checked, so the seeded `use_sim_time` cannot trip the check.
- **The launch seed.** Only what the application declares is checked. Until W3 (#867) lands, the seed's value type is inferred from text, and the resolver has already held the seed to the contract. Checking it again would refuse, for example, a declared `string` whose launch value reads `true`.
- **The contract's host path.** A refusal names the scope's recorded manifest by file name only, so no host path gets baked into firmware. Without a recorded manifest it uses `<stem>.contract.yaml`.

### Coordination

This touches `node_runtime.rs`, `main_macro.rs` and `nros-platform`'s `runtime.rs`, which #867 is also changing. The hunks are additive: a new trait method with a no-op default, one check in the parameter arm, and one token stream emitted next to `#param_services_call`. Whichever PR lands second rebases.

## Test plan

- [x] `declared_params.cpp` (wired into `just check cpp`): every lookup is `constexpr`, so it `static_assert`s each answer the boot check depends on against a rendered fixture, and asserts that the fixture table was found at all. It compiles under c++17, with and without `NROS_CPP_STD`. `declared_params.hpp` also parses under `-std=c++14 -ffreestanding`.
- [x] `the_committed_compile_fixture_is_what_this_emitter_renders` holds the committed fixture header to the renderer. The fixture model is `declared_params.yaml`, not `*_model.yaml`, because check-no-tracked-models bans a tracked SystemModel.
- [x] `just check declared-qos-header`: 20 assertions, including a new one that the params header lands in the component's PRIVATE dir from a real configure with the real CLI.
- [x] Unit tests: `nros-params` (the Rust check, exemptions, namespaces, type codes), `nros-orchestration-ir` (derivation, contract naming), `nros-cli-core` (renderer). `cargo test -p nros-cli-core --lib` 1072 passed. `cargo test -p nros-macros --lib` passed.
- [x] `cargo clippy -D warnings` on nros-params, nros-orchestration-ir, nros-macros, nros-platform, nros-node (`param-services,alloc,sim-time[,rmw-cffi]`), nros (`std,param-services,rmw-cffi,sim-time`) and nros-cli-core. `cargo +nightly fmt` and `clang-format` are clean.
- [x] `just check api-parity`: the six new public names have ledger rows (`extension`).
- [x] `just check fast`: only `capability-conditionals` and `xrce-vendored-versions` fail. They fail in a bare worktree only because the zenoh-pico submodule and the vendored XRCE tree are absent. **This was pushed with `NROS_SKIP_PREPUSH_CHECKS=1` for that reason alone.**
- [ ] Not run: a booted image with a mismatching node. W7 is the end-to-end check on the downstream.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QKY5TGsRzN9SVs1e4c8BoH


jerry73204:
Rebased onto main after the queue ejected this for merge_conflict. Cause: phase-427 W4 (`1f3b88aec`) deleted `ComponentNode` while this was queued, and W6 had put the declared-parameter check on `ComponentNode::declare_parameter`. The check is ported to `Node` (beside `check_declared_depth`), the type trait moved to `node_parameters.hpp` as freestanding-safe specializations (the `-nostdinc++` probe in `check-cpp` has no `<type_traits>`), and `cpp:Node::check_declared_param` has its ledger row. The pin commit `993d3f979` dropped by patch-id — main already has it. Details in the new commit's message.

PR #879, opened 2026-09-10
feat(phase-436 E3): say when the stack-headroom rule is armed but blind

jerry73204:
Based on #871. Until #871 merges, this diff includes its 3 commits; **review only the top two**. The queue drops the carried commits by patch-id once #871 lands.

## E3: say when the stack-headroom rule is armed but blind

`check_stack_headroom_rule` treats a 0 from the port's `stack_unused_bytes` as "not instrumented" and skips. So an armed rule that can never judge reports nothing, exactly like one that judged and found the stack healthy.

It wasn't hypothetical. On ASI's default Zephyr FVP build the entry arms the rule, but the build has `CONFIG_THREAD_STACK_INFO` without `CONFIG_INIT_STACKS`, so Zephyr's accessor returns 0. Every "0 headroom violations" from that build meant *not checked*.

Now `set_min_stack_headroom_bytes` asks the port once. If a bound is set and the port answers 0, it logs one warning naming the fix and records it; `Executor::stack_headroom_blind()` reads it back. The warning appears at boot, once per tier. C entries get it through `nros_cpp_executor_derive_min_stack_headroom`.

## B1 decided: no timer deadline source

This is recorded in Path B. W1 already bounds the park for the executor's own timers, with attribution. A one-shot timer signalling the wake object buys nothing for them: the executor is single-owner, so only the parked thread can add a timer. It isn't more precise on Zephyr either, and it costs allocations, ISR constraints and attribution. The mechanism stays for the bare-metal park (C3), and for deadlines raised mid-park from outside the executor.

## Verified locally

- Test-first. The first RED attempt used `rmw-cffi`, which excludes `executor/tests.rs`, and so passed for tests that couldn't compile. Re-run with `--features std`, it failed on exactly `no method named stack_headroom_blind`, then passed.
- The blind test compares against the host port's real answer. That's the POSIX C port, which answers 0, so the case that matters is the one exercised.
- `just check node-std-tests` rc 0, including the executor header-budget test.
- clippy `-D warnings` with and without `rmw-cffi`; nightly fmt; `check-api-parity` rc 0, with the new ledger row a 4-line insertion; roadmap gates OK.


jerry73204:
Superseded: both commits of this branch are already on `main`, landed 2026-09-10 18:46 UTC — `ca7a46d17` (feat(phase-436 E3): say when the stack-headroom rule is armed but blind) and `926ce96e7` (docs(phase-436): B1 decided (no timer deadline source); E3 done). A rebase of `d53fe8e48` onto `origin/main` drops both by patch-id and leaves 0 commits, which is also why the queue ejected it with merge_conflict. #899, which carried these two commits and was held behind this PR, can now be rebased onto `main`.

PR #882, opened 2026-09-10
fix(ci): the doctor called a correctly provisioned host broken — three probes and three stores

jerry73204:
Follow-up to #842, which is in the merge queue and therefore frozen. Everything
here was found by bootstrapping the contained runner from an empty store and
reading what it said.

**One shape, four times.** A probe exists, is authored, is correct about the
question it asks — and something else decides the answer:

* `runner-doctor`'s Zephyr probe restated `ZEPHYR_WORKSPACE` from
  just/zephyr.just, with a comment saying the derivation "lives in a `just`
  variable, which no shell can read". That stopped being true — the recipe
  shells out to `scripts/lib/zephyr-workspace.sh`, plain shell — so the copy's
  reason was gone and the copy had drifted past the STORE location the resolver
  puts first. It reported MISSING for a host that had just completed
  `west update` and the whole patch set. It asks the resolver now.
* `west is not on PATH`, on a store that had west. `~/.local/bin` and
  `~/.cargo/bin` are volumes precisely because they are installable as the
  runner user, and the image put neither on PATH. Exported in the entrypoint,
  which also hands it to every job — a runner whose PATH differs from the one
  its labels were checked against is a runner whose labels were checked against
  nothing.
* the Zephyr SDK, complete and unpacked, unregistered with cmake because
  `~/.cmake` was not a volume — so `find_package(Zephyr-sdk)` would have failed
  at configure, not at download, which is what the doctor's message warns about
  by name.
* `just setup zephyr` stopping to tell you to run `nros setup native --rmw
  xrce`. Fixed in the verb rather than in the runner's label map: the map would
  have made this runner work and left every contributor the same wall on their
  first setup.

**Three more stores** (`.cmake`, plus `.local` and `.rustup` from #842) and two
more baked prereqs (`qemu-system-misc` for the `qemu-system-riscv64` the
`nros-qemu` label claims; `wget`, now declared in the index because the Zephyr
SDK's own setup.sh requires it by name and exits 91 after a 1.4 GB download).

**The bootstrap asks the doctor even when provisioning failed**, because
"provisioned" and "has it" are different questions and the labels are the
contract. Issue #1276 is the case: `nros setup rv-virt-threadx --rmw cyclonedds`
panics on its seventh package, after the toolchain and QEMU that `nros-qemu`
names are already installed. The failure is reported loudly before and after,
and the status captured explicitly rather than left to `set -e` (issue 1249).

Verified end to end:

```
runner-doctor: OK — all 3 label(s) verified.
√ Runner successfully added
Listening for Jobs

nano-ros-runner  online  busy=true  [self-hosted,Linux,X64,nros-qemu,nros-sdk-zephyr,nros-big]
```

It registered only after passing, and the PATH bug is why that ordering is worth
having: the first attempt refused rather than winning `L3` and failing inside it.

The baseline row for issue 1276 goes when #870 lands and its file arrives.

### DIFF
--- a/just/xrce.just
+++ b/just/xrce.just
@@ -27,6 +27,21 @@ setup:
     # The recipe must not re-implement that decision: build.sh prints
     # "already built" itself when the stamp matches, so the fast path is
     # preserved without a second, weaker copy of the rule.
+    #
+    # PROVISION FIRST. build.sh prefers the prebuilt agent from the nros store
+    # and, finding neither that nor a checked-out submodule, exits 1 with
+    # "Micro-XRCE-DDS Agent not provisioned. Run: nros setup native --rmw xrce".
+    # It is right to refuse — provisioning is `nros setup`'s job and it says so
+    # — but a verb named "install XRCE prerequisites" that stops to tell you to
+    # run the install command is a step, not a verb. So run it: three packages,
+    # all XRCE (the agent plus the two client submodules a Zephyr XRCE build
+    # needs anyway), idempotent, and it prints "present (skip)" when the store
+    # already has them.
+    #
+    # Found by `just setup zephyr` on a fresh host: it calls this recipe last,
+    # after `west update` and the whole patch set, so a clean provisioning run
+    # failed at the final step with an error naming a command nothing had run.
+    nros setup native --rmw xrce
     ./scripts/xrce-agent/build.sh
 
 # Diagnose XRCE install status (read-only).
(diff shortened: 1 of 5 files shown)


jerry73204:
Taken out of the merge queue on purpose: this branch carries all 8 commits of #842 (it is stacked on it) and sat AHEAD of #842, so the queue marked it UNMERGEABLE. Once #842 merges, this gets rebased onto main (the 8 shared commits drop by patch-id) and re-armed.

PR #899, opened 2026-09-10
feat(phase-436 A3): per-window jitter readout for C entries, and what the idle A/B measured

jerry73204:
Based on #881 (E3 + E4). Until #881 merges, this diff includes its 4 commits; **review only the top two**. The queue drops the carried commits by patch-id. Do not queue this while #881 is in the queue: the duplicated patches would make it unmergeable, as happened with #879 and #871.

## A3: a per-window jitter readout, and what the idle A/B on ASI established

**Code:** `nros_cpp_executor_clear_release_jitter_stats(handle)` exposes the existing Rust `clear_release_jitter_stats()` in the A2 accessors' shape. The executor's release-jitter figures are counted since it opened. On ASI's FVP image a ~34 ms wake at boot pinned the maximum for the whole run, and nothing in the steady state could show above it. ASI's readout now clears after each report, so every `rt-probe:` line covers its own window.

**Doc:** A3 now records four things:
- The probe and the CTF trace share CNTVCT on FVP (`k_cycle_get_64` / `_32`). So A3 there is a self-consistent check, not absolute latency.
- **E4 measured** on the same painted image, idle, with and without it: ~26.7 → ~162 wakes/s, timer overruns ~6.5/s → ~0.2/s, `main` CPU 15–19 % → 1 %. Each pre-E4 query scanned 458 992 unused bytes.
- The since-boot maximum was pinned by boot, which is the reason for this PR.
- **A2c (open):** `late_wakes` counts any overshoot of 1 µs or more, while the executor reports `granularity_us=1000`. So a wait-paced loop reads ~99 % late. For example, a 27 µs timer park is rounded up to the 1 ms tick by design.

## Verified locally

- Test-first: the null-handle test failed to compile on exactly the missing function, then passed (3/3 readout tests).
- `just regen-c-headers` → `check-cbindgen-headers` OK: one new declaration, nothing else.
- clippy `-D warnings`, nightly fmt, `check-api-parity` rc 0, roadmap gates OK.
- The E4 A/B figures come from two 150 s idle FVP boots of ASI's `--trace-stats` image, built fresh on `30293e5c8` (main, no E4) and on `248d75f66` (#881, E4).

### DIFF
--- a/docs/roadmap/phase-436-poll-wake-revision-deadline-driven-executor.md
+++ b/docs/roadmap/phase-436-poll-wake-revision-deadline-driven-executor.md
@@ -496,11 +496,56 @@ watched run.
     loop can declare its cadence without over-claiming.
   * **Left for A3:** the ASI side printing these on a cadence.
 
-* **A3 — a loaded FVP cross-check.** Run the control loop under load and
-  compare `release_jitter()` and `last_park()` against an independent CTF
-  capture of the same run. **Exit:** the probe's maximum and the trace's agree
-  within the declared `release_jitter_granularity_us()`, and a disagreement
-  is explained, not averaged away.
+* **A3 — a loaded FVP cross-check. IN PROGRESS: the readout is live on ASI,
+  an idle A/B is measured, and the loaded run is next.** ASI's
+  `controller_pkg::Controller` wrapper prints one `rt-probe:` line per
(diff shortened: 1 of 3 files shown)


jerry73204:
Taken out of the merge queue on purpose: this branch carries both commits of #879, which is ahead of it in the queue, so the pair collided (#879 went UNMERGEABLE). Once #879 merges, this gets rebased onto main (the 2 shared commits drop by patch-id) and re-armed.

PR #900, opened 2026-09-10
fix(#1287): the Zephyr C++ feature string spells alloc with param-services

jerry73204:
A Zephyr C++ image on zenoh that declares `param_services` fails to compile at nros-node's requirement check:

```
compile_error!("`param-services` allocates: add \"alloc\" to this crate's features");
```

## Why
- The zenoh C++ feature string in `zephyr/CMakeLists.txt` does not spell `alloc`; it gets it transitively through nros-rmw-zenoh. The xrce and cyclonedds strings spell it.
- The issue 0745 mirror appends `param-services` to that string.
- The nros-c library built for C++ copies `alloc` and `param-services` into its own features only when the C++ string spells them. So on zenoh it got `param-services` without `alloc`.

## Fix
One line: append `alloc` together with `param-services` (phase-361 W8.e: a capability names its requirement). Issue 1287 is filed in `archived/` with this PR, plus a changelog fragment.

## Evidence
Autoware Safety Island's MR-CANHUBK344 image (zenoh, four C++ component nodes, `param_services`) fails with the error above without this line and builds with it. The island has been carrying this exact line as a local submodule patch.

## Relation to PR 843
Split out of PR 843, which is held. The Rust half of 843's original problem was fixed on main by eb8111ad1. After this lands, 843's `zephyr/CMakeLists.txt` hunk is redundant.

`just check fast`: 293 of 295. The two failures, `capability-conditionals` and `xrce-vendored-versions`, happen only because this bare worktree lacks the zenoh-pico submodule and the vendored XRCE trees. Pushed with `NROS_SKIP_PREPUSH_CHECKS=1` for that reason.

https://claude.ai/code/session_01QKY5TGsRzN9SVs1e4c8BoH

### DIFF
--- a/changelog.d/1287.fix.md
+++ b/changelog.d/1287.fix.md
@@ -1,0 +1,4 @@
+A Zephyr C++ image on zenoh that declares `param_services` builds again. The
+C++ feature string now spells `alloc` next to `param-services`, so the nros-c
+library built for C++ gets both, instead of stopping at "`param-services`
+allocates: add \"alloc\" to this crate's features".
--- a/docs/issues/archived/1287-zephyr-cpp-param-services-without-alloc.md
+++ b/docs/issues/archived/1287-zephyr-cpp-param-services-without-alloc.md
@@ -1,0 +1,48 @@
+---
+id: 1287
+title: "A Zephyr C++ image on zenoh with param_services fails to compile: nros-c
+  is built with param-services and without alloc"
+status: resolved
+type: bug
+area: build, zephyr
+severity: medium
+resolved_in: "fix(#1287): the Zephyr C++ feature string spells alloc with param-services"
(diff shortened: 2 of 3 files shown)


jerry73204:
Closing as superseded: `main` already carries this fix. #843 (merged 2026-09-10 20:18, commit `a39f26682`) changed the same line in `zephyr/CMakeLists.txt` to `string(APPEND _nros_cpp_features ",alloc,param-services")`, with an equivalent comment (it filed the same defect as issue 1260). Its `changelog.d/1260.fix.md` already says "On Zephyr the capability now names `alloc` as well", so a 1287 fragment would announce the same change twice. Rebasing this branch onto main leaves only a comment conflict on that one line — no code difference. Note: issue id 1287 was reserved for this PR and no file for it exists on main.

PR #905, opened 2026-09-11
feat(phase-445 W4b): a single-package example builds from build/<image>/nros-cargo.toml

jerry73204:
Implements phase-445 **W4b** (RFC-0098 D1/D2/D7) and W1's acceptance for single packages. **Not armed until the esp32 QEMU suite passes on this branch** (see W1 below).

**One writer.** #880's `builder/cargo_config.rs` gains path-valued `[env]` rows, a header command and the three `nros-cargo-profile` presets. New `cmd/leaf_settings.rs`: a leaf with a `[package]` manifest + `system.toml` + a cargo-driver board gets `<leaf>/build/<image>/nros-cargo.toml` (board `cargo_config` < derived pools + entity facts < board facts < the leaf config's remaining `[env]` rows — the last layer until W6). `nros sync` writes it last; `nros build` in a leaf writes it then execs `cargo build --manifest-path <leaf>/Cargo.toml --config <leaf>/build/<image>/nros-cargo.toml`. `nros ws leaf-system` prints `NROS_LEAF_SETTINGS=<path>`.

**Measured constraints handled:** relative paths resolve against the file's grandparent (`<leaf>/build/`); the `nros-*` profiles now travel IN the file (no dependence on the repo-root config); the leaf's own `.cargo/config.toml` board-projection `include` repeats the file's rustflags and cargo CONCATENATES arrays (`rust-lld: error: memory.x:19: region 'FLASH' already defined`), so cargo runs from the leaf's PARENT until W6 deletes the leaf `.cargo/`.

**W1.** The model alone left the consumer at its 8-slot default, so the facts were completed: the application's service-server count from the leaf's own inventory (probe or `[[component]] entities`), and the infra fact combined with the service features the manifest enables. `NOT_DERIVED_NEEDS_INFRA_COUNT` → `QUERYABLES_DERIVED_BY_CONSUMER`; the leaf road still never states the count. Measured, same settings file, only the leaf `[env]` line removed:

| image | `ZPICO_MAX_QUERYABLES` | `ZPICO_MAX_SUBSCRIBERS` | `.stack` |
|---|---|---|---|
| esp32 talker | 2 → **1** | 1 → 1 | 94,248 → 98,752 B |
| esp32 listener | 2 → **1** | 1 → 1 | 63,984 → 68,480 B |

[...]

PR #912, opened 2026-09-11
feat(phase-445 W5): every workspace entry is generated, and a workspace with no bringup builds like colcon

jerry73204:
Completes phase-445 **W5** (RFC-0098 D9, RFC-0065 D4/D13). Carries #905 (W4b) so it can target `main`; the queue drops it by patch-id.

**Entries converted** (hand package deleted, deployment moved to the bringup `system.toml`):

| entry | became | build | test |
|---|---|---|---|
| `workspaces/rust/src/esp32_entry` | `[image.esp32]` | rc=0 | `test_esp32_workspace_entry_e2e` PASS (6.4 s) |
| `templates/multi-node-workspace-cpp/src/robot_entry` | `[image.native]` | rc=0 | `cpp_multi_node_entry` 4/4 |
| `templates/c-and-cpp-mixed-workspace/src/robot_entry` | `[image.native]` | rc=0 | `c_mixed_workspace` 2/2 |
| `templates/pure-c-workspace/src/robot_entry` | `[image.native]` | rc=0 | `workspace_metadata` PASS |
| `workspace-shadowing`, `local-msg-package` (roots) | package mode | rc=0 | `workspace_shadowing` PASS |
| 7 Zephyr entry manifests + workspace fixtures | claiming image | compile-check all rows rc=0 | 5 suites PASS |

**Package mode:** no bringup, no image requested, root not a package → every package in dependency order, each with its own driver and `build/<pkg>/`. `NrosRclcppCompat.cmake` moves to workspace scope (stock-ROS packages need its Find-stubs).

**Fallback deleted:** `leaf_system::from_manifest`, `BoardFrom`, `is_fallback`, `board_from`, sync's deprecation notice and `NROS_LEAF_FALLBACK` are gone; `leaf_system::for_entry` is the ONE reader for `nros::main!`, the board projection, `ws board-facts` and `nros build`. `check-leaf-deployment-spelling` no longer skips workspace manifests. Five template root `CMakeLists.txt` deleted; the only roots left are the three SINGLE-PACKAGE templates, where `CMakeLists.txt` sits beside that package's own `package.xml`.

**Two real defects the verification caught, both fixed here:** `nros::main!` fell back to the image id for an unknown declared board (`board = "frobnicator"` built as native — main was strict, my earlier W5 fallback caused it); and `workspace_metadata` still looked for `nros-metadata.json` at the old template-root path.

**Verified** (no stale-check override): native Rust workspace lane 21 images, `workspace-{c,cpp,mixed}-native`, esp32, FreeRTOS, `int32-sink`, native + esp32 talker/listener rows, every compile-check row — all rc=0; 38/39 tests then 3/3 esp32 after rebuilding a fixture; `just ci gate` PASSED. **No Zephyr image built** — the west workspace symlinks the main checkout (issues 1253/1280).

**Where the RFC was wrong, now amended in-tree:** D3/D5 assumed a leaf `system.toml` serves a workspace entry (it cannot — a package with one reads as a second bringup, so the generated entry wrote the very `deploy` key D5 retires); D9's "directory of packages" had no build path without an `[image.*]`; D6 is still false for the eight Rust west entries (issue 1288).

**Filed:** 1288, 1289, and **1295** (high) — every generated Rust workspace entry on CycloneDDS fails `PublisherCreationFailed`; the selection facade never names the `nros` crate's type-descriptor feature. Reproduced on the in-tree `workspace-rust-native-cyclonedds` binary too, and `facade.rs` has no diff against main, so it predates this PR.


jerry73204:
Superseded by #917 and closed. Not a rejection — a rebase this PR could not do for itself.

#905 (W4b) merged at 03:38, and this PR went `UNMERGEABLE` right after. The reason is the collision the two only have TOGETHER: W5 deleted `LeafSystem::is_fallback` and its `board_from` field, while W4b added code and tests that name both. Each PR compiles alone; the pair does not. The queue tests a PR against what is AHEAD of it, so neither run could see it.

#917 carries **all six** of this PR's commits, rebased onto post-#905 `main`, plus the three fixes for that collision and the #1295 facade fix. Verified subject-by-subject: every commit here is in #917. `just ci gate` passes there; it did not on the pair without the fixes.

Nothing from this PR is lost. Review continues on #917.

PR #942, opened 2026-09-11
feat(#1142, phase-412 W5): a standalone CMake leaf sizes its pools from its own declaration

jerry73204:
`examples/qemu-armv7a-nuttx/cpp/action-client` is a standalone copy-out project — `find_package(nano_ros)` plus `nano_ros_add_executable`, no bringup, no SystemModel — and it fell between all three declaration channels:

| channel | why it missed |
|---|---|
| CMake **workspace** | wants a resolved model; `nros_record_entity_facts` returns early without one, by design |
| **Zephyr** | wants the `CONFIG_NROS_MAX_QUERYABLES` derive sentinel |
| **cargo leaf** | wants `nros sync`'s `[env]` sidecar, which no cmake build reads |

So `NROS_DECLARED_SERVICE_SERVERS` and `NROS_DECLARED_INFRA_QUERYABLES` never arrived and `queryable_default_from` took its last arm — `if hosted { 32 } else { 8 }`, a guess by construction. Issue 1028 measured what a wrong guess costs one path over: 142,336 B of `.bss` on an image with zero queryables.

## Measured, on the image the issue names

```
SERVICE_BUFFERS   35,584 B (0x8b00)  ->  4,448 B (0x1160)
image .bss       508,208 B           -> 467,248 B
```

The `.bss` delta exceeds the `SERVICE_BUFFERS` one because the C shim's per-session queryable table is sized from the same knob. `35,584 / 8 = 4,448` is the arithmetic: eight guessed queryables against the one this image declares.

## What was actually missing

RFC-0098 **D8** had already decided *where* such a leaf states its surface — `entities = [...]` on its `system.toml` `[[component]]`, the `EntityDecl::parse` grammar every other road uses. RFC-0098's own "What is NOT decided" left open how a C/C++ leaf's CMake **reads** it. That is this PR:

1. **The declaration.** All twelve `examples/qemu-armv7a-nuttx/{c,cpp}/*` leaves declare, each list read from its own `src/` rather than guessed — the action-client creates one node and one action client on `/fibonacci`, so its application queryable count is a real zero. Spot-verified against source before committing (`c/talker` → publisher + timer; `cpp/service-server` → one `create_service(srv, "/add_two_ints")`).
2. **Reading it at configure time.** `nros ws entity-facts --leaf <dir>` applies the MODEL road's own counting rule and prints the same three `KEY=VALUE` facts; `nros_fold_entity_facts` folds them into the same accumulator; `nros_entity_facts_env` delivers them through the same `corrosion_set_env_vars` carrier. **Nothing new is derived** — a second derivation is how two roads come to size one image two ways.
3. **A new module, `cmake/NanoRosLeafEntityFacts.cmake`**, deliberately not more edits to `NanoRosEntityFacts.cmake`, which a sibling PR was changing at the same time.

The facts stay a DEFAULT, never an override: a stated `ZPICO_MAX_QUERYABLES` still wins, and a stated value below the derived floor is refused rather than silently shrinking the pool. Phase-412's ladder, unchanged.

## Found while building it, filed rather than folded in

**Issue 1301** — the NuttX link lane builds its own cargo command and delivers no BOARD facts, and `check-board-facts-delivery` cannot see it because the gate reads two directories and the lane is in a third. The issue-0196 shape: a gate whose reach is narrower than its rule.

## Verification

- `just check fast` on the rebased commit: **305 gates, rc=0**, 1 environmental skip.
- `just test-unit`: **rc=0**, 1312 tests, 1311 passed — the one reported `failed` is a capability skip the junit rewrite reclassifies (`ZPICO_MAX_SESSIONS=1`).

Those two together are what `just ci gate` runs; run as halves because this shared box kills long processes on memory pressure.

`check-markdown-links` caught a real one on the first run: archiving 1142 left a dead link in phase-412's homed table. Fixed, and the gate is green — *"a markdown link is CLICKED, so it must resolve exactly."*

Rebased onto current main (57 commits). Pushed with `--no-verify` — the hook's `check fast` is the half already run and passed; no submodule pin moves.

https://claude.ai/code/session_01Wsdnh5xJGqPb9sT9BPVHXX


jerry73204:
Taken out of the merge queue on purpose: it collided with #944, which sits ahead of it and also edits `packages/core/nros-orchestration-ir/src/leaf_system.rs` (replaying this branch on top of #944 conflicts there; against `main` alone it rebases clean). Held until #944 merges, then rebased onto main and re-armed.

PR #956, opened 2026-09-11
fix(phase-417 stage 3): four C names that compiled and differed — and the siblings beside them

jerry73204:
phase-417 stage 3 (the loudness pass) for the C executor and guard condition:
four ledgered `gap` rows in `docs/reference/api-parity-ledger/exec.json`, each
an upstream name whose contract differed from rclc's, each with disposition
`adopt`. All four adopted, plus the siblings a sweep of each family found.

**Based on `origin/work/phase-444-gap-ledger` (PR #901), not `main`** — it
carries the api-parity ledger truth pass these rows sit in. This branch
therefore carries #901's commit `dda94a9`; the merge queue rebase-merges, so it
drops by patch-id once #901 lands.

## What changed

| ledger row | was | is |
| --- | --- | --- |
| `c:executor_trigger_one` | `obj` read as `*(size_t *)obj`, used as an INDEX — the documented upstream idiom compiled clean and never fired | `obj` is the ENTITY POINTER, as in rclc |
| `c:executor_spin_some` | `NROS_RET_TIMEOUT` on every idle tick, so a ported `if (spin_some(...) != RCL_RET_OK) error();` fired the first time the transport was quiet | `NROS_RET_OK` for any completed cycle — rclc discards `rcl_wait`'s TIMEOUT |
| `c:executor_spin_period` + `c:executor_set_timeout` | both period spins passed `period_ns` as the wait, so `rclc_executor_set_timeout` was obeyed by `rclc_executor_spin` and silently ignored by the other two | the wait is `timeout_ns`; the period is the invocation rate and is spent on the sleep |
| `c:guard_condition_fini` | `NROS_RET_NOT_INIT` for a zero-initialised or already-finalised handle | idempotent — `NROS_RET_OK`, NULL still `INVALID_ARGUMENT` |

Adopting `trigger_one` needed the trigger ABI to carry entity identity at all,
because a `const bool *ready` array carries none: the predicate now receives
`nros_executor_trigger_handle_t[]` (`{ entity, data_available }`), the C
executor records one entity pointer per handle slot through ONE helper at all
ten `..._add_*` sites, and `rclc_executor_set_trigger` installs a bridge that
joins the internal readiness snapshot to that table. The index form survives as
`nros_executor_trigger_index` — the gap row is what asked for a different name.
The new element deliberately does NOT take rclc's `rclc_executor_handle_t`
name, so a ported custom trigger reaching for `handles[i].subscription` fails
to COMPILE.

## Siblings, found by sweeping rather than reported

* `rclc_executor_trigger_all` returned `false` for an empty handle array;
  upstream's loop does not execute and returns `true`.
* `rclc_executor_fini`'s `== UNINITIALIZED` guard let a SECOND call reach
  `drop_in_place` on storage the first call had zeroed.
* The `fini` class was swept with
  `grep -rn 'extern "C" fn [a-z_]*_fini' packages/api/nros-c/src/` and each
  member measured against its OWN upstream source: `rcl_node_fini` and
  `rclc_executor_fini` are idempotent upstream (both say so in a comment) and
  are fixed; `rcl_clock_fini` and `rclc_support_fini` are NOT (upstream returns
  `RCL_RET_INVALID_ARGUMENT` for an uninitialised clock) and were left alone;
  `rcl_timer_fini` is stage 3's other assignment and is untouched.

The `c:guard_condition_fini` row's explicit EVIDENCE BOUND
("implementation-derived and not verified here") is closed by reading
`rcl/src/rcl/guard_condition.c` @ humble.

## Tests

Seven new/rewritten unit tests. Mutation-checked: with all seven fixes reverted
in place and the tests kept, `cargo test -p nros-c --features rmw-cffi,std
--lib` reported exactly those seven failing, 106 passing.
`test_guard_condition_fini_not_init` was deleted — it asserted the divergence
rather than recording it.

## Ledger

Five gap rows deleted (closed); `c:executor_trigger_handle_t` and
`c:executor_trigger_index` added as `extension`.
`python3 scripts/api-parity.py --check` passes; `--self-test` reports 0
failures.

## Local checks

* `cargo test -p nros-c --features rmw-cffi,std --lib` — **113 passed, 0
  failed**.
* `just check c` — **exit 0**, "All C checks passed!" (includes the
  `timer_clock_source` runtime fixture, which drives a real executor through
  `rclc_executor_spin_some`).
* `python3 scripts/api-parity.py --check` — **exit 0**, "every divergence
  carries a ledger entry"; `--self-test` — "0 checks failed".
* NuttX snapshot probe: `gcc -std=c11 -fsyntax-only -DNROS_PLATFORM_NUTTX
  -Ipackages/api/nros-c/include` on a TU including `<nros/nros_generated.h>` —
  exit 0, `_handle_entities[64]` in the preprocessed output (issue 1115's
  two-file rule, checked rather than assumed).
* `just ci gate` — **FAILED at check::fast, 2 of 298 gates**, both for THIS
  WORKTREE'S MISSING PROVISIONING, not for this diff:
  * `check-capability-conditionals`: "…zenoh-pico/include/zenoh-pico/system/
    common/platform.h is missing — run `git submodule update --init`".
    `git submodule status` shows the submodule uncheckout (`-dd071b8d…`).
  * `check-xrce-vendored-versions`: "no vendored tree checked out — nothing
    verified. Run `nros setup --source micro-cdr --source
    micro-xrce-dds-client`."
  The lane stops at the first failure, so `check::build`, `check::api-parity`,
  `test-unit` and `test-lane-contracts` did NOT run from it.
* `just check build` run separately — **NOT COMPLETED**. It sat ~70 minutes with
  its cargo invocations at 0 CPU time, blocked on the shared cargo
  build-directory lock (several agents share this checkout), and was terminated
  (`error: recipe 'build' was terminated on line 223 by signal 15`). It is
  `schedule`/`workflow_dispatch`-only in CI and needs the same absent
  provisioning. **`test-unit` and `test-lane-contracts` were therefore not run
  locally** — the merge queue runs them.

https://claude.ai/code/session_01S7tie7AFCdqoRkxphQcqRx


jerry73204:
Taken out of the merge queue on purpose: it conflicts with #952, its phase-417 stage 3 sibling, which sits ahead of it in the queue. Both append a section to `docs/roadmap/phase-417-ros2-api-adoption.md` (#952 +45 lines, this one +33), so replayed on top of #952 the queue cannot apply it — against `main` alone it rebases clean.

Held until #952 merges, then rebased onto main (the two sections are an append-vs-append, so the resolution is a union) and re-armed.

PR #963, opened 2026-09-11
feat(phase-442 W0-W2): a timer's size stops following a compiler flag, the callback capacity is measured, and the third vocabulary goes

jerry73204:
Phase-442 W0, W1 and W2 — the first three work items of RFC-0096's one
freestanding `rclcpp` API. Three commits, each independently green.

## W1 — a timer's size stops following a compiler flag (closes issue 1225)

`sizeof(nros::Timer)` was 24 or 32 depending on `NROS_CPP_STD`, and four other
subjects followed it: `nros::GuardCondition`, the timer-array holder
`nros::NodeWithTimers<N>` at 8 bytes per slot, and the `rclcpp::Timer` /
`rclcpp::TimerBase` aliases, which are the spellings a ported file writes. One
member between all five, an `#ifdef NROS_CPP_STD`
`std::unique_ptr<std::function<void()>> closure_`.

A capability probe may gate a method; it may never change a layout. Two
translation units of one image disagreeing about a capability is a supported
state here — `examples/px4/cpp/bridge/.../CMakeLists.txt:123` sets
`-DNROS_CPP_STD=1` on one module of a larger image, deliberately — so this was
issue 0135's defect live in a configuration that ships.

The member is unconditional now and spelled `void*`, pointing at a block whose
head is `detail::HostedBlockBase` carrying its own `void (*destroy)(void*)`.
That indirection is what lets `~Timer()` — one inline destructor compiled in
both configurations — free a hosted closure without naming a hosted type.
`rclcpp::Node::hosted_` already had exactly this shape as a private type in
`node.hpp`; rather than write it a second and third time it moves to
`nros/hosted_block.hpp`, and `node.hpp`'s two hand-written destroy sequences
become one shared helper call.

Measured on all arms of `check-cpp-capability-layout` (hosted, hosted without
`-DNROS_CPP_STD`, `-nostdinc++` freestanding against the ThreadX shim):

| subject | before | after |
| --- | --- | --- |
| `::nros::Timer` | 24 / 32 | 32 everywhere |
| `::nros::GuardCondition` | 32 / 40 | 40 everywhere |
| `::nros::NodeWithTimers<4>` | 328 / 360 | 360 everywhere |
| `::rclcpp::Timer`, `::rclcpp::TimerBase` | with `nros::Timer` | 32 everywhere |

The five `diverges` rows leave `.config/cpp-capability-layout-baseline.txt` in
the same commit, which is the acceptance rather than the paperwork: that ratchet
fails on a row naming a subject that no longer diverges, and its selftest proves
that arm.

The 8 bytes are transitional — W3 replaces the block with a fixed-capacity
inplace callable and the pointer goes with it. W1 lands first because it ends a
shipping exposure and depends on nothing.

## W0 — the two knobs, measured

**Capture sizes.** Census over 308 files: `[this]` (7 sites), `[state]`,
`[this, state]`, `[this, &state]`, `[obj, method]`, plus ten `[&]` that are
`std::thread` bodies and local string helpers in the Cyclone backend, not one of
them an rclcpp callback. Sized with `sizeof(decltype(lambda))` on all three
toolchains:

| capture | hosted x86_64 | arm-none-eabi cortex-m3 | riscv64 |
| --- | --- | --- | --- |
| `[]` | 1 | 1 | 1 |
| `[this]` / `[state]` | 8 | 4 | 8 |
| `[this, state]` | 16 | 8 | 16 |
| `[obj, method]` | 24 | 12 | 24 |

Every measured capture is a whole number of pointers, and the widest is three —
`diagnostic_updater`'s, wide because a pointer-to-member-function is two words
on the Itanium ABI. That decides the knob's units: a byte count would be correct
on one word size and wrong on the other.

**`NROS_CPP_CALLBACK_CAPACITY`, default `4 * sizeof(void*)`** — the widest
capture plus one pointer of headroom. `.bss` measured on eight registered
callbacks at six capacities; per callback the cost is `align8(Cap) + 8`. Two
things the table settles that arithmetic would not: `.text` is flat across every
candidate, so capacity is a `.bss` knob and choosing generously does not grow
code; and a capacity below two pointers buys nothing on 32-bit, since the
storage must be `alignas(long long)`. At the default, eight callbacks cost 192
bytes on 32-bit and 320 on 64-bit, against a shipping Zephyr image's
`NROS_CPP_EXECUTOR_STORAGE_SIZE 89352`. The over-budget case was measured too:
`static assertion failed: callback capture too large -- raise
NROS_CPP_CALLBACK_CAPACITY`.

**`X::SharedPtr` needs no control block.** The corpus stores handles as members
and passes one by value (`diagnostic_updater::Updater`'s constructor), with no
`weak_ptr`, `use_count()` or `reset()` outside our own headers. The decisive
measurement is that shared ownership is already fiction at the one site where it
looked real: `Node::shared_from_this()` returns
`std::shared_ptr<Node>(std::shared_ptr<void>(), this)`, the aliasing constructor
with an empty owner. What the corpus compiles against today is already a
non-owning handle wearing `shared_ptr`'s spelling.

RFC-0096's open questions 1 and 3 close with these numbers. Question 2 — whether
NuttX's missing `-ffreestanding` is deliberate — is untouched.

W0 also found a trap W4 should not fall into: hand-declaring the placement
`operator new` with `unsigned long` is accepted on x86_64 and riscv64 and
rejected on cortex-m3, so the obvious workaround is right on two of our three
toolchains and silently wrong on the embedded one.

## W2 — delete `std_compat.hpp`

275 lines, 71 `std::` occurrences, the whole file behind `#ifdef NROS_CPP_STD`,
included from one place behind the same guard. No source outside it calls any of
its eleven free functions; the one build that defines `NROS_CPP_STD` does so for
`<nros/bridge.hpp>`, which is independently guarded and untouched. It was a
third orphaned vocabulary — free functions shadowing the `Node` methods a ported
file calls, predating the phase-427 merge — and RFC-0096 removes the layer it
belonged to.

Two things the delete surfaced, both worth a reviewer's time.

**A return-type divergence the deleted overload was masking.**
`nros::create_publisher<M>(node, topic, qos)` returns `ResultOf<Publisher<M>>`
where `rclcpp::create_publisher` returns `std::shared_ptr<Publisher<MessageT>>`.
Both halves are forced — no `shared_ptr` on a freestanding target, and no
throwing, so the value must carry the error — and same name plus same arity
means a ported file that discards the result loses the failure silently, which
is what `same:ret` exists to flag. It was out of the gate's reach only because
the name carried two overloads of different arity. Deleting the dead one did not
create the divergence; it stopped hiding it. A ledger row states it now.

**`check-api-parity` stayed green over ten ledger rows describing entities that
no longer exist.** The ledger ratchets in one direction only: a difference with
no row fails, a row with no difference costs nothing. Both sibling ratchets in
this area explicitly forbid that shape —
`.config/cpp-capability-layout-baseline.txt` fails on a row whose subject no
longer diverges and proves it in a selftest. Filed as **issue 1323**. The ten
rows are removed here and four surviving rows amended, because a ledger row is
not only an exemption, it is a written claim about what a porting user gets:
nine said "it is the `NROS_CPP_STD` overload in `std_compat.hpp`" about a header
that no longer exists.

One of the ten was my error and the gate caught it: `cpp:create_subscription`
lives in the `differs` bucket, which neither `--show all` nor `--show same`
prints, so a probe built from those views called it dead. `check-api-parity`
failed instantly and the row went back — which is now issue 1323's first design
note, that the reverse walk must read every bucket the extractor computes rather
than the ones a view renders.

`check-ledger-orphan-refs` is the half of this that already exists and works: it
found all ten stale `std_compat.hpp` citations, including the four in rows whose
key survives.

## Verification

`just check fast` — 298 of 298 green.
`just check cpp-capability-layout`, `cpp-freestanding-includes`, `api-parity`,
`ledger-orphan-refs` each green individually. The umbrella header parses on both
arms, `g++ -std=c++17` with and without `-DNROS_CPP_STD=1`.

https://claude.ai/code/session_011A4bG1pAtNwG5BDJBFBtfb


jerry73204:
Taken out of the merge queue on purpose: it collides with #947, which sits ahead of it in the queue. Both edit `docs/issues/archived/1225-capability-probe-changes-sizeof-timer-and-guard-condition.md` — #947 homes the issue to a phase, this branch records the W1 capability-probe outcome. Against `main` alone this branch rebases clean (3 commits, unchanged), so nothing is wrong with it. Held until #947 merges, then rebased (union of both edits to that file) and re-armed.

PR #973, opened 2026-09-11
refactor(phase-442 W6+W7): 27 mechanical `std` uses go, and the freestanding API measures as already `std`-free

jerry73204:
Phase-442 W6 and W7, plus the two RFC-0096 decisions W8 needs before it can
start.

**Carries #963, #964 and #965's commits** (W0–W5), because W6 uses W3's
`nros::tr`. Those drop by patch-id on the queue's rebase once the three land —
the pattern CLAUDE.md records for #699, #550 and #626. Review the last four
commits; the first six are already under review elsewhere.

## W6 — 27 `std` uses that never needed to be `std`

The removable bucket, separated from W7 and W8 because none of it is a design
question: every site already had a freestanding spelling in this tree or in the
language, and no consumer loses a capability.

- 13 `std::move` → `nros::tr::forward_rvalue`
- 3 `std::forward` → `nros::tr::relay`
- 14 `std::size_t` / `std::uint8_t` / `std::int32_t` / `std::uint32_t` → the
  bare spellings the rest of this API already writes. `transport.hpp` is now
  entirely `std`-free — it had nothing but std-spelled C types in a struct of
  function pointers.
- `result.hpp`'s `<utility>` include, verified unused by removing it and
  recompiling rather than by reading.

**`relay` is not `forward_rvalue`, and that is why there are two.** The obvious
economy — one helper for both — is wrong in the direction that does not announce
itself: forwarding preserves an lvalue as an lvalue, a move hands the callee an
rvalue it may gut, and a forwarding wrapper that moves silently steals from its
caller. Both exist, with the distinction written on the one that is easy to
misuse.

**Deliberately left:** the hosted diagnostics (`fprintf`, `fopen`, `getenv`,
`ostringstream`, `abort`, 13 sites). They are genuinely body-only and already
carry a freestanding `#else` — `result.hpp`'s `NROS_TRY_LOG` is the pattern. A
`std` call inside a hosted-only body changes no signature and no layout.

## W7 — met by measurement, and the measurement is the deliverable

W7 asked for 27 entities to be replaced with in-tree equivalents. They already
have them. After W6, the remaining 134 `std::` sites split:

| | sites |
| --- | --- |
| behind `NROS_CPP_STD` / `NROS_CPP_HAS_*` / `NROS_CPP_NODE_HOSTED` | **129** |
| ungated | **5** |

All five ungated are legitimate: two are string *literals* inside a diagnostic
message (which is why this is a comment-stripping scan and not a grep), two are
`std::abort()`, and one is `std::initializer_list` — whose header is
freestanding-guaranteed because it is core language support for braced
initialisation, not a library convenience.

**So the freestanding API is already `std`-free.** The census's predicted gap — a
`std::vector` overload with no `Seq` sibling — does not exist; checked pair by
pair in `parameter.hpp` (:274/:284, :322/:350, :372/:378, and :308 for the
string form). Its named holder, `ComponentNode`, was deleted by phase-427 W7.

That reframes W8: not "replace 27 types" but "collapse 129 gated sites into one
ungated API".

## RFC-0096 D8 — a name parameter is deduced, so `std::string` stays drop-in

A decision the RFC was missing, surfaced by W6. Every hosted `create_*` takes
`const std::string&` behind `NROS_CPP_NODE_HOSTED`, and W8's acceptance is zero
`NROS_CPP_HAS_*` — but D5's list of what is *not* drop-in does not include
`std::string` arguments. Deleting those overloads would have added a fourth item
to that list silently.

The parameter is deduced, and a two-overload `detail::as_c_str` converts it. The
header names no `std` type; a caller passing a literal, a `std::string`, our
`FixedString` or our `HeapString` all bind. This is W5's `NodeOptions::arguments()`
move generalised, and there it was measured to be *better* for a porting user
too. Measured here on three configurations:

| | |
| --- | --- |
| hosted `g++ -std=c++17`, caller passes `std::string` | rc=0 |
| ThreadX shim `-nostdinc++`, no `<string>` reachable | rc=0 |
| `arm-none-eabi -std=c++14 -ffreestanding` (32-bit) | rc=0 |

One cost, stated: a wrong argument type diagnoses one frame inside the template.
`as_c_str` is an overload set rather than a trait to keep it to one frame; W8
should add a `static_assert` so the first line names the parameter.

## RFC-0096 D9 — W8's first decision is entity storage, and it is not made yet

D2 says what `X::SharedPtr` *is*. It never said what the handle points *at*.

Today `create_publisher<M>(topic, qos)` does `std::make_shared` and pushes into
`owned_entities` — eight `push_back` sites, no `erase`, one drain in `~Node()`.
The census called that address stability wearing a `shared_ptr` face, and it was
right. But address stability is a **requirement**, not a fiction: the executor
arena holds a raw dispatch pointer and has no unregister path, so the entity must
outlive the caller's handle. Removing `make_shared` removes the mechanism, not
the requirement.

Why it is not a detail:

| entity | `sizeof` |
| --- | --- |
| `Timer` | 32 |
| `Service<int>` | 560 |
| `Publisher<int>` | 824 |
| `Subscription<int>` | 888 |
| `Client<int>` | **4 672** |

A uniform worst-case pool is wrong by an order of magnitude — one unused
`Client` slot costs more than twenty timers. D9 states three candidates;
the one that fits this repository is entity counts **derived** the way
phase-412 already derives `NROS_CPP_EXECUTOR_STORAGE_SIZE`, and it needs a
`just mem-report` before/after rather than a decision at the keyboard.

**W8 should not start before this is answered.** The remaining 129 sites are
substitution once it is, and substitution on top of an unanswered lifetime
question is how a use-after-free ships.

## Verification

`just check fast` — 309 gates green, 1 skipped. All three compile arms green
(hosted, hosted `-DNROS_CPP_STD=1`, `-nostdinc++` freestanding against the
ThreadX shim).

https://claude.ai/code/session_011A4bG1pAtNwG5BDJBFBtfb

### DIFF
--- a/docs/design/0096-cpp-freestanding-core-and-porting-layer.md
+++ b/docs/design/0096-cpp-freestanding-core-and-porting-layer.md
@@ -372,6 +372,72 @@ making it unconditional removes all five divergent subjects at zero cost and
 ends the px4 mixed-layout exposure immediately. Independent of everything else;
 land it first.
 
+### D8 — A name parameter is DEDUCED, so `std::string` stays drop-in (amendment, 2026-09-12)
+
+This decision was missing, and implementing W6 is what surfaced it. Stated here
+because W7 and W8 cannot proceed without it.
+
+**The gap.** Every hosted `create_*` in `nros.hpp` takes `const std::string&`,
+which is upstream's spelling, and the whole family sits behind
+`NROS_CPP_NODE_HOSTED`. W8's acceptance is zero `NROS_CPP_HAS_*` in the tree, so
+those overloads cannot survive as gated methods. But D5 — "what is NOT drop-in,
+enumerated" — does not list `std::string` arguments, and deleting the overloads
+would silently add a fourth item to that list: every ported call site holding a
+`std::string` name would need `.c_str()`.
+
+Three candidate answers, and only one satisfies both halves of the goal:
+
+| | freestanding-clean | `std::string` call site still compiles |
+| --- | --- | --- |
+| keep the overloads, gated | no | yes |
+| delete them, require `.c_str()` | yes | **no** — a fourth D5 item |
+| **deduce the parameter** | **yes** | **yes** |
+
+**The decision: the parameter is deduced, and one overload set converts it.**
+
+```cpp
+namespace nros { namespace detail {
+inline const char* as_c_str(const char* s) { return s; }
+template <typename S> auto as_c_str(const S& s) -> decltype(s.c_str()) { return s.c_str(); }
+}}
+
+template <typename M, typename S>
+Publisher<M>::SharedPtr create_publisher(const S& topic, const QoS& qos);
+```
+
+The header names no `std` type. A caller passing a string literal, a
+`std::string`, our `FixedString`, our `HeapString`, or anything else with
+`c_str()` all bind. This is the same move W5 already made for
+`NodeOptions::arguments()`, generalised — and there it was measured to be
+strictly better for a porting user as well, because the refusal could speak
+where an overload-resolution failure used to happen first.
+
+**Measured**, not reasoned — one probe, three configurations:
+
+```
+hosted g++ -std=c++17, caller passes std::string          rc=0
+ThreadX shim -nostdinc++, no <string> reachable at all    rc=0
+arm-none-eabi -std=c++14 -ffreestanding (32-bit)          rc=0
+```
+
+**One cost, stated.** An argument that is neither a `const char*` nor a type
+with `c_str()` produces a diagnostic one frame inside the template rather than
+at the call site:
+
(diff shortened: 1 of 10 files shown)


jerry73204:
Taken out of the merge queue on purpose: this branch carries all three of #963's commits (`phase-442 W0`, `W1`, `W2 delete std_compat.hpp` — identical subjects, verified with `git log --format=%s origin/main..` on both), and #963 sits ahead of it in the queue, so the pair collided and this entry went UNMERGEABLE. Against `main` alone this branch rebases clean (7 commits, unchanged), so nothing is wrong with it. Held until #963 merges, then rebased onto main — the three shared commits drop by patch-id — and re-armed.

PR #988, opened 2026-09-11
feat(phase-454 W10): the declared-QoS check was C++-only, and the two other languages need two different roads

jerry73204:
phase-454 W10. Independent of W3–W9; based on `main`.

## The brief was half wrong, and the correction shaped the work

The wave was scoped as "arm the header for Rust and INTERFACE targets". Measured:

- **C already receives the descriptor.** `_nros_declared_qos_arm()` returns early
  only for RUST components and INTERFACE targets. A C component is a STATIC
  library like a C++ one, so it has had `nros_declared_qos_generated.h` on its
  PRIVATE include path since issue 1084 and lacked only a *reader*. **No cmake
  change was needed for C.**
- **Rust cannot receive it at all.** `nano_ros_node_register(LANGUAGE RUST)`
  creates an **empty INTERFACE library** (`NanoRosNodeRegister.cmake:626-651`);
  the sources are compiled by the workspace runtime crate, so there is no
  preprocessor in that lane and no include path to put a header on.

So the two languages needed two different roads, not one shared arming fix.

## C — compile time

New `packages/api/nros-c/include/nros/declared_qos.h` with
`NROS_ASSERT_DECLARED_DEPTH`. Three measurements each ruled out the obvious port
of the C++ version:

- `"abc"[0]` is **not** an integer constant expression on gcc or clang
- a macro parameter of the caller is **not** substituted inside a separately
  defined row macro (`use of undeclared identifier 'q_type'`)
- a `_Static_assert` message **cannot** interpolate an `int`

So the lookup folds `__builtin_strcmp` — verified under `-pedantic`,
`-ffreestanding`, `-fno-builtin`, and on the pinned `arm-none-eabi-gcc 13.2` —
the generator emits a second query-parameterised `NROS_DECLARED_QOS_ROWS_Q` from
the same loop, and the two depths reach the reader through a conflicting
`char[N]` redeclaration (`char[10]` vs `char[1]`) while the topic comes from the
assertion message.

## Rust — registration time

`NROS_ENTITY_DECLARED_DEPTHS` now reaches every lane via
`_nros_declared_depth_table_env`, a sibling of `_nros_qos_depth_env` carrying
only the `resolved` guard. `nros-node/build.rs` renders
`config::DECLARED_QOS_ROWS`; `declared_qos::check` refuses with
`NodeError::DeclaredDepthMismatch` at all **14** subscription-creation sites,
including the C/C++ FFI seams — which is how a **C component** reaches it too,
since `nros_cpp_subscription_register` is what a C configure function calls. Same
`-403` as the C++ boot check.

## The sizing path is untouched, deliberately

The new lane forwards the depth table but **not** the undeclared count, so
`subs_arena`'s guard still returns early and every arena number is byte-for-byte
unchanged.

That split is the point: **sizing must refuse a partial picture; per-endpoint
checking must not.** Every in-tree contract leaves at least one subscription
silent, so reusing the sizing guard would have meant nothing was ever checked.

## Negative controls — run, captured, reverted

1. **C mechanism** — `static assertion failed: "...topic \"/chatter\"
   disagrees..."` plus `conflicting types ... 'char[10]'` / previous declaration
   `'char[1]'`.
2. **C delivery** — stripping `NROS_DECLARED_QOS_ROWS_Q` from the fixture made
   the positive TU red at `NROS_DECLARED_QOS_COMPILE_TIME == 1` *and* the
   mismatch probe compile; both halves of the gate catch a lost table.
3. **Rust** — mutating `check` to return `Ok(())` failed with
   `left: Ok(()) right: Err(DeclaredDepthMismatch)`.

## Gates

- `check-declared-qos-header` grew case F2 (C, against a header a real configure
  just rendered): **25 assertions, up from 16**
- `just check c` grew the positive/expected-failure pair with the diagnostic
  grepped
- **`check-declared-qos-registration` is new** and answers *delivery*: it builds
  nros-node twice and requires the declared-image test to pass with the env var
  and to **not exist** without it — so an always-on cfg cannot make it green over
  nothing
- three `extension` rows added to the api-parity ledger

Green: `check c`, `check declared-qos-header`, `check declared-qos-registration`,
and `ci gate`'s `cli-fresh`, `fast` (314 gates), `api-parity`, `test-unit` (1453
passed), `test-lane-contracts` (17).

## Not from this change

`check::build` has five reds, all agent-worktree artefacts: `rmw-uorb` and
`rmw-xrce` carry a cmake cache generated against the main checkout;
`staticlib-symbols` and `borrowed-e2e` have fixture scripts writing into the main
checkout's `build/` while the test reads the worktree's. `cpp` dies at
`ros2_api_adoption_stage2.cpp:263` on a `Duration`→`chrono::nanoseconds`
conversion that **reproduces with `origin/main`'s nros-c headers substituted in**,
so it predates this branch. Because that lane never reaches the declared-QoS C++
steps, those were compiled by hand against the regenerated fixture: positive
clean, probe rejected naming `declared_depth_agrees<1, 10>`.


jerry73204:
Taken out of the merge queue on purpose: it collides with #993, which sits ahead of it (queue position 2 vs 4). Both are phase-454 siblings and both edit `packages/cli/nros-cli-core/src/entity_inventory.rs`; replayed on #993 alone this branch conflicts there, while against `main` alone it rebases clean (2 commits, unchanged), so nothing is wrong with it. W4 lands before W10. Held until #993 merges, then rebased and re-armed.

PR #1000, opened 2026-09-12
test(phase-444 W6, #1268): the Cyclone parameter cell, and the live run that closes 1268

jerry73204:
Issue 1268 was fixed in the tree on 2026-09-11 and deliberately left **open**, because its first acceptance — `ros2 param list/get/set` reaching every node of a multi-node Cyclone image — needs a ROS 2 peer and the agent that made the fix had none. This is that run, plus the cell that keeps it checked.

## Measured

In the `ros2` distrobox (Ubuntu 22.04, Humble, `rmw_cyclonedds_cpp`), against a two-node Cyclone image built from this tree:

```
param services registered for node: /alpha
param services registered for node: /beta

=== ros2 service list --no-daemon | alpha/beta:
/alpha/describe_parameters        /beta/describe_parameters
/alpha/get_parameters             /beta/get_parameters
/alpha/get_parameter_types        /beta/get_parameter_types
/alpha/list_parameters            /beta/list_parameters
/alpha/set_parameters             /beta/set_parameters
/alpha/set_parameters_atomically  /beta/set_parameters_atomically

=== ros2 param list:            === ros2 param get:
/alpha:                           /alpha rate -> Integer value is: 10
  rate                            /beta  rate -> Integer value is: 20
  stepped                       === ros2 param set /beta rate 21
/beta:                            Set parameter successful
  rate                            /beta  rate -> 21   /alpha rate -> 10 (unmoved)
```

All twelve services — six per node, the exact set the issue reported missing — are on the wire; each node answers with its OWN value for the name they share; a write to one leaves the other alone.

## The fixture could not be pointed at Cyclone at all

`bins/param-two-node-talker` registered its backend with a hardcoded `nros_rmw_zenoh::register()`, pinning it to the one backend on which a parameter service cannot fail on type grounds — a service server **is** a queryable there. That is exactly why the zenoh cell was green for the whole life of this defect. The bin takes the RMW feature axis `graph-probe` already has, and `examples/fixtures.toml` grows the second row `param-two-node-talker-cyclone`; `build_native_param_two_node_talker_rmw` selects between them the way `build_graph_probe_rmw` does.

## The cell

`native-params-per-node-rust-cyclone` in `interop::CELLS`, case `ros2_param_cli_addresses_each_node_of_a_cyclone_image`, runner `just native test-ros2-params-per-node`, verdict recorded `pass`. The case asserts `ros2 service list` as well as the three `param` verbs, because the missing SERVICES are the form the issue was reported in and asserting only the consequence would pass on a build that exposes them and cannot answer them. The binary now hosts two cells, so it gains a `CASE_CELLS` map — without one a Cyclone red would retract the zenoh cell's recorded pass.

Also recorded: **`native-params-per-node-rust-zenoh`**, which had never produced a verdict since phase-426 W6 declared it, passes — both of its cases ran live in the same run.

## Ran

* `just native test-ros2-params-per-node` → 4 passed, 0 skipped
* `check-interop-verdicts --self-test`, `check-interop-cell-runners`, `matrix_fixture_coverage` (10/10)
* `just ci gate` → one failure, `check-third-party-is-submodules`, on untracked provisioning directories (`third-party/{make,ninja,ros}`) that predate this branch in the box tree and are not in the diff. That is also why the push used `--no-verify`; `check submodule-pins` and `check issue-ids` were run by hand first.

Closes #1268.

https://claude.ai/code/session_01S7tie7AFCdqoRkxphQcqRx


jerry73204:
Held for ordering behind #947, and deliberately left unarmed until that one merges.

Both branches add a `examples/fixtures.toml` row with the same id, `param-two-node-talker-cyclone`, and both extend `packages/testing/nros-tests/tests/params_per_node_interop.rs`, `src/interop.rs` and `src/fixtures/binaries/mod.rs` with the phase-444 W6 / issue-1268 acceptance. #947 carries that work across three commits (612c6e2e5, 11c989760, f0d877141); this PR carries a single condensed commit (b848fbd14) that also adds the `.config/interop-verdicts.toml` rows, the changelog fragment and the `just/native.just` wiring.

#947 is the older PR and is already armed, so it goes first. Whichever of the two entered the queue second would have ejected the other on `invalid_merge_commit`, which is why this one came out at 2026-09-12T00:40 on `merge_conflict`.

Once #947 merges: rebase this branch onto `main`, drop the hunks main then carries in equivalent form (keep the verdict rows, the changelog fragment and the `native.just` wiring, which are only here), push with a lease, and arm auto-merge.

This PR is separately behind `main` on `packages/testing/nros-tests/src/fixtures/binaries/mod.rs` and `packages/testing/nros-tests/tests/params_per_node_interop.rs`, so the rebase is needed regardless.

jerry73204:
Hold released and rebased — #947 merged, so the phase-444 W6 / issue-1268 mechanism and acceptance are now on `main`.

Rebased onto `main` (1 commit before, 1 after). Seven files conflicted, because main now carries #947's version of the same work. Resolved by dropping this branch's copy wherever main carries it in equivalent form, and keeping only what is unique here:

**Dropped** (took main's): the archived issue doc (main's resolution section is fuller and carries the `resolved:` frontmatter), `param-two-node-talker/{Cargo.toml,src/main.rs}`, `src/fixtures/binaries/mod.rs`, `src/interop.rs`, `tests/params_per_node_interop.rs`. The auto-merge had also produced a **duplicate** `[[fixture]]` row with the id `param-two-node-talker-cyclone`; main's block is the one that survives.

**Kept** (not on main): the `.config/interop-verdicts.toml` row recording the live run of `native-params-per-node-rust-cyclone`, `changelog.d/1268.fix.md`, and the `just/native.just` note that the two halves of this binary now have different preconditions (zenoh wants `rmw_zenoh_cpp` + a router, Cyclone wants `rmw_cyclonedds_cpp` and none).

One correction the rebase forced: the verdict row named `ros2_param_cli_addresses_each_node_of_a_cyclone_image`, but the test that landed on main is `ros2_param_cli_addresses_each_node_on_cyclonedds`, bound to that cell in `interop::CELLS`. Renamed to match, so the row names a test that exists.

Net against main is 30 insertions over 4 files. `check fixture-variant-features` OK (10 selectors), `check fast` 320 of 322 — the two reds being `capability-conditionals` and `xrce-vendored-versions`, which need submodule sources this worktree does not have, so the push used `--no-verify`.

New head `df28f1a41`; auto-merge armed.

PR #1004, opened 2026-09-12
fix(build, #1280): an inherited absolute path outranked the checkout being built

jerry73204:
## The census is 24, not 19

Measured live with `env | grep nano-ros` inside an agent worktree, not read off
the issue:

| source | n | what the issue's list missed |
| --- | --- | --- |
| `just/sdk-env.just`, `justfile_directory()`-rooted | 21 | **`NROS_LAN9118_LWIP_DIR`, `PX4_AUTOPILOT_DIR`** |
| `just/sdk-env.just`, derived | 1 | `IDF_PATH` |
| `activate.sh` / `activate.fish` | 2 | **`NROS_REPO_DIR`, `nano_ros_ROOT`** |

`FREERTOS_PORT` is the one `sdk-env` export naming no path. **All 24 pointed at
the main checkout from inside the worktree.**

## The rule, not the list

The issue's own addendum says a list cannot be the fix — and its list was already
short by five. Three-valued, and the middle row is why *"is it set?"* cannot be
the discriminator:

- **outside any checkout → keep** (a real out-of-tree SDK; this is what env-first
  exists for)
- **a different checkout → re-root, and say so**
- **this checkout → keep**

Ownership is a lexical walk for `packages/core/nros-core/Cargo.toml`, never
`.git` — a linked worktree's `.git` is a *file* (#1336) — and the walk still
attributes a directory that does not exist yet. That last part matters: in a
worktree these SDK trees are usually **absent**, and that absence is exactly what
made the inherited path look like it was working.

Three spellings (`scripts/lib/checkout-paths.sh`, a `replace()` prefix rewrite in
`sdk-env.just` fed by one ~24 ms detector subprocess, `nros_build_paths::reroot_foreign`)
because the three build systems cannot call each other — the `riscv64` precedent.
The gate pins them to each other. `NROS_REPO_DIR` / `nano_ros_ROOT` stop being
`env(...)` at all. Two or more foreign checkouts cannot be served by one prefix,
so that case **refuses** rather than guessing.

## One rung deliberately left alone

`NROS_BUILD_ROOT` is **not** re-rooted. Its Rust mirror `nros_tests::build_root`
reads it un-re-rooted and cannot be taught the rule without a dep edge;
re-rooting only the shell side would split writer from reader about where a
fixture lives — **a fresh instance of the bug being fixed**. The gate asserts the
pass-through, with the argument, so a later "finish the job" edit has to meet it.

## Measured acceptance

1. `just nuttx build` names the **worktree's** tree and declines, instead of
   reconfiguring the main checkout's NuttX kernel.
2. `#error` in the worktree's `nros-platform-freertos/src/platform.c`, built for
   `thumbv7m-none-eabi` under the inherited env: **0 hits before, 4 after**. The
   same probe against a genuine out-of-tree copy: **0 both ways** — row 1 does not
   regress.
3. With `NROS_REPO_DIR` exported to the main checkout, `nros_build_root` →
   `<worktree>/build`; `just check rmw-uorb` builds and passes in
   `<worktree>/build/uorb-check`.
4. `NROS_BUILD_ROOT=/mnt/fast/nros-build` unchanged; an out-of-tree `NUTTX_DIR`
   survives even beside an inherited foreign `NROS_C_INCLUDE`.
5. Normal case (no foreign checkout): detector silent, `just --evaluate` correct,
   20 ms total.

## Two things reproducing it corrected

- **`cargo.sh`'s `nros_sizes_probe_dir` validated an inherited `NROS_REPO_DIR` by
  probing for `build-root.sh`** — which *accepts* the value it should refuse,
  since another checkout has that file too. It only ever caught malformed values.
- **`test-unit` carried `--exclude nros-build-paths`**, so the five unit tests
  added here would have run in **no merge-gating lane**. The exclusion was written
  for the *embedded clippy* lane (issue 208: a `std`-only crate vs `thumbv7em`)
  and copied onto a host lane along with the list. Removed from `test-unit` only;
  the embedded lane still derives it from the crate's own `host-only = true`.

## Gate

`just check inherited-checkout-paths`, fast lane. Covers `sdk-env.just` (24
today, 1 declared non-path with its reason), one marker across all three
spellings, and **behaviour** — the shell rule, `nros_build_root`, and `just
--evaluate` each driven against synthetic checkouts for all three rows, because a
source read would pass an implementation that never touches the filesystem.

Its self-test measures mutations as a **delta against the live file**, so a real
violation cannot read as "the self-test is broken". Five negative controls, all
run and confirmed red: dropped `replace()`, reversed `replace()`, a new unwrapped
export, a deleted re-root in `nros_build_root`, and an **added** one on
`NROS_BUILD_ROOT`.

## Lane state, honestly

`just check fast`: **313/315**, stable across three runs. The two failures are
this worktree's uninitialised submodules — `zenoh-pico` for
`capability-conditionals`, `micro-xrce-dds-client`/`micro-cdr` for
`xrce-vendored-versions`. Both name the missing file; neither touches anything in
this diff. That is **issue #1345**, filed separately: a gate that cannot reach its
subject should report NOT VERIFIED, not FAIL.

`just ci gate` therefore stops at step 2 of 6 and never reaches `check::build`,
which is running standalone. A first attempt died on a machine-wide full disk —
notably while writing into `<worktree>/build/compile-check-fixtures/`, **which is
the fix working**.


jerry73204:
Additional evidence for the blast radius, from phase-454 W6.a — worth folding into the issue before this closes.

A lane invoked from worktree `agent-ad2ff520fd47659c6` was writing into a **different** agent's worktree. From its build log, 377 instances of:

```
error: couldn't create a temp dir: No space left on device (os error 28) at path
  ".../agent-a8c2c6202f2cf5140/target/nros-relwithdebinfo/build/rustix-.../rmetajKDFvz"
```

That is this issue's symptom in a **third** place, beyond the two acceptance cases: not the `rmw-uorb` / `rmw-xrce` cmake caches and not the FreeRTOS source path, but **cargo's own target dir for the ordinary Rust lanes**.

Two consequences:

1. **The blast radius includes every worktree lane that compiles Rust**, not only the cmake-configured ones. Acceptance row 2 (the `#error` probe) covers a C source path; a cargo `target/` case would be a distinct row worth adding, since it is reached through a different resolution chain.

2. **It explains the disk exhaustion this campaign kept hitting.** N concurrent agents writing into one worktree's `target/` is why `/home` went from 360 GB free to 0 twice in a session, and why the failures landed in a worktree whose own agent was not running the build. That reframes the disk pressure as a symptom of #1280 rather than an independent environment problem.

Not asking for a change to this PR — the fix's rule (`a different checkout → re-root`) already covers the case. Recording it so the acceptance list and the issue's "what this touches" reflect it, and so the disk-exhaustion connection is not rediscovered later as a separate problem.

jerry73204:
Two qualifications from the author's stand-down, both of which narrow what this PR should be read as proving.

## 1. `NROS_BUILD_ROOT` is a named, remaining vector — not just a deferred tidy-up

The commit argues *why* `NROS_BUILD_ROOT` stays pass-through (re-rooting only the shell side would split writer from reader, since `nros_tests::build_root` reads it un-re-rooted). It does **not** state the consequence, which is:

> **an exported `NROS_BUILD_ROOT` still carries cargo writes across checkouts.**

That matters for the cross-worktree `rmeta` error in my earlier comment. Ruling out the alternatives: `CARGO_TARGET_DIR` is **not** one of the 24 — unexported by both `activate.sh` and `sdk-env.just`, unset in practice, and its fallback in `scripts/build/cargo.sh` is `$PWD/target`, worktree-local *by construction*. So that write did not arrive through the cargo default. The remaining vectors are a hand-set `CARGO_TARGET_DIR`, a repo-rooted derivation through `NROS_REPO_DIR` (which this PR re-roots), or `NROS_BUILD_ROOT`.

So the honest framing is: **#1280 is fixed except through the rung we deliberately argued to leave alone.** Worth saying in the issue explicitly rather than leaving it to be rediscovered — and worth deciding whether the follow-up is teaching `nros_tests::build_root` the rule, or accepting the rung with the gap named.

## 2. Do not cite `borrowed-e2e`'s pass as acceptance evidence

It passed, and it is one of the four `check::build` gates the issue names — but **a pass is not the discriminator here**. Before the fix it would have run against the main checkout's `build/` and passed *wrongly*.

The discriminating measurements are the ones in the commit: the `#error` probe (0 hits before, 4 after; 0 both ways out-of-tree), and `nros_build_root` returning `<worktree>/build` under an inherited foreign `NROS_REPO_DIR` with `rmw-uorb` then building in `<worktree>/build/uorb-check`.

## Also

`node-std-tests` and `workspace-features` are **unmeasured, not suspect** — they emitted no error of any kind before the lane aborted and SIGTERM'd them. The queue's run is their first real verdict.

PR #1006, opened 2026-09-12
feat(phase-454 W6.d+e): uORB and cffi size from the descriptor — and a pool the floor gate could not see

jerry73204:
phase-454 W6.d (uORB) + W6.e (cffi). Carries W3 (#984) and W4 (#993); the queue
drops them by patch-id.

## The descriptor gained an `[image]` section

Three counts **no endpoint row can be counted to get** — which is exactly
RFC-0100 D5's cffi row:

```toml
[image]
node_count = 2          # components().len() — derived since phase-412, handed only to the executor
backend_count = 1       # what the image links
subscriber_count = 3    # declared subs + each action client's feedback sub
```

`subscriber_count` is a **fact rather than a row count, on purpose**: an action is
one declared entity costing several session slots, so a build script counting
`action_client` rows and multiplying would be a **third** mirror of a multiplier
that already has two — in a file `check-infra-queryable-counts` does not scan. A
test pins the difference (1 row → 2 slots).

**No `schema_version` bump**, and the reasoning is measured rather than assumed:
the version guards meaning *moving*, and an added section moves nothing. An older
reader already refuses loudly via `deny_unknown_fields` — verified by feeding a
new descriptor to an older `nros-node` build script, which rejected it naming
`nros sync`.

## What derives

- **W6.d (uORB)** — `NrosRmwUorbSizing.cmake`, in the backend's own build, since
  uORB has no Rust crate at all. `REGISTRY_CAPACITY` = distinct topics over
  publisher + subscription rows, deduped by name; `PX4_MAX_CALLBACKS` =
  `subscriber_count`.
- **W6.e (cffi)** — `MAX_BACKENDS`, `MAX_NODES`, `SUBSCRIBER_SLOTS` take the
  descriptor as the rung **under** any stated value, replacing the crate builtin.
  RFC-0049's ladder unchanged. `MESSAGE_INFO_SLOTS` and `SLOT_SIZE` untouched
  (the latter is issue 1322, and D5 says it is not derivable).

## Measured

| image | pool | before | after | delta |
| `esp32-c3-baremetal/rust/listener` | uORB `g_table` (`nm` on the `.o`) | 1,536 B | 24 B | **−1,512 B** |
| `native/rust/talker` | cffi `REGISTRY` (`nm` on a linked binary) | 328 B | 48 B | **−280 B** |

`NODE_TABLE` is `MAX_NODES` × a measured 144-byte `NodeSlot` → 576→144 on any
image registering a node. The probe binary does not keep it live, so that one is
**arithmetic over two measured quantities and is reported as such**, not as a
symbol read.

**No descriptor ⇒ byte-identical**, measured the same way: the same uORB configure
with no descriptor emits no `-D`, and `g_table` is 1,536 B.

**D7 in both directions in one image**: the pub-only talker demands 0
subscribers; cffi's `[T; 0]` takes it untouched, and uORB raises it out loud —
`NROS_RMW_UORB_PX4_MAX_CALLBACKS=0 raised to 1`.

## Two things found beyond the brief

**1. A gate whose reach was narrower than its rule.** `topic_registry.cpp` wrote
`Entry g_table[kCapacity]` through a `constexpr` alias, which **launders the
macro past `check-c-array-pool-floors`' discovery** — the audit reported 24
knob-sized arrays over a tree with 25, and the invisible one is a knob this wave
makes derivable. Extent is the macro now, with a `#if < 1 / #error` guard and a
`PROBE_CONTEXT` in **compile** mode against a real TU. Now 25 arrays / 22
guarded; 22 guards across 10 files fire at 0.

**2. A short-registry hole in my own first cut.** An `ENDPOINT_COUNT` of 0 has
two causes that read identically in cmake — no endpoints, or an inventory that
did not compose. The row-derived capacity is now gated on
`NROS_SIZING_UNDECLARED_ENDPOINTS` being *stated*, because **a short registry is
the one direction with no safe fallback**. Negative control run: a refused
descriptor keeps the builtin 64, loudly.

`_nros_c_array_pool_floor` moved to `cmake/NanoRosPoolFloor.cmake` — a PX4 module
build cannot include a Zephyr file, and one shared helper beats a third copy. The
gate's FLOORED knob set is **per producer** now; a global list made the Zephyr
ZPICO bridge answerable for a pool in a PX4 module.

## Gates

Green: `check fast` 314/315, `cli-tests`, `c-array-pool-floors`,
`c-array-guard-probe`, `pool-inventory --check` (regenerated),
`sizing-descriptor-reader`, `declared-fact-carriers`, `knob-single-reader`,
`knob-delivery`, `kconfig-knob-forwarding`, `config-knob-census`,
`cli-source-dirs`, `cli-fresh`. Tests: `nros-sizing-descriptor` 22/22,
`nros-cli-core` sizing 18/18, `nros-rmw-cffi` green, uORB builds `-Wpedantic`
clean + ctest passes. Clippy clean on all three touched crates.

**Environmental, none from this change:** `check::build` and
`test-lane-contracts` did not finish (~80 min, 24 sibling worktrees, `/home`
oscillating between 0 and ~1.4 GB free). `test-unit` fails on the unprovisioned
`cyclonedds` submodule — every touched crate was tested directly instead. `just
check rmw-uorb` fails with `does not match the source used to generate cache`,
which is **issue 1280's class** (fixed in #1004); the package was verified in a
private build dir — configure, parallel build, ctest, all clean. The remaining
`check fast` red is `xrce-vendored-versions`, needing `nros setup --source`
(issue #1345's class).

## Note for the sibling W6 waves

They will each want to extend the same `[image]`/schema surface, and
`deny_unknown_fields` means a **stale descriptor on disk is a hard error rather
than a silent default** — expect a re-sync, not a merge conflict.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -892,8 +892,14 @@ One-liners; detail in the linked doc. (Many also captured in agent memory.)
   1015's floor — landed in the derivation a day before 1033's fix — silently defeated
   it, and every knob gate stayed green because the number was derived correctly and
   delivered faithfully. Floor at the pool (`c_array_pool_floor` /
-  `_nros_c_array_pool_floor`), keep `#if X < 1 / #error` beside the array as the
+  `_nros_c_array_pool_floor`, the latter now in `cmake/NanoRosPoolFloor.cmake`
+  because phase-454 W6.d gave it a second backend — uORB's two C++ pools, whose
+  build cannot include a Zephyr module file), keep `#if X < 1 / #error` beside the
+  array as the
   backstop that binds a producer neither reaches. Gate: `check-c-array-pool-floors`
+  (its FLOORED knob set is PER PRODUCER since W6.d: a global list made the Zephyr
+  ZPICO bridge answerable for a pool in a PX4 module, and the only way to satisfy
+  that is a written abstention that says nothing true)
   (also refuses an unruled new one; the 1 still unruled is issue 1131 — the
   other 14 were ruled there, 10 guarded and 4 already covered). **Issue 1131
   has since ruled the last one too, so the table is EMPTY and
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2059,6 +2059,7 @@ dependencies = [
  "nros-cc-flags",
  "nros-log",
  "nros-rmw",
+ "nros-sizing-descriptor",
  "nros-zephyr-build",
  "portable-atomic",
 ]
--- a/cmake/NanoRosPoolFloor.cmake
+++ b/cmake/NanoRosPoolFloor.cmake
@@ -1,0 +1,60 @@
+# NanoRosPoolFloor.cmake — the C-array pool floor, in ONE place.
+#
+# Issues 1015 + 1033, and RFC-0100 D7:
+#
+#   > derivation publishes demand, unfloored; the floor lives at the consumer.
+#
+# A derived count is the image's DEMAND, and zero is a legitimate demand.
+# Whether zero is a legal SIZE is a property of the STORAGE, so it is decided at
+# each pool that names a knob — never in the shared derivation, which feeds
+# consumers whose right answers at zero disagree:
+#
+#   zenoh   `queryable_entry_t queryables[ZPICO_MAX_QUERYABLES]` is a fixed C
+#           array. A derived 0 gave a board that transmitted NOTHING in 15
+#           seconds — no panic, no log line, core in WFI, every gate green,
+#           because the number was derived correctly and delivered faithfully.
+#
+#   XRCE    the SAME derived numbers reach `NROS_XRCE_MAX_SUBSCRIBERS` and
+#           `NROS_XRCE_MAX_SERVICE_SERVERS`, where zero is the ANSWER and worth
+#           33,296 / 4,384 bytes of heap a slot.
+#
+#   uORB    `Slot g_pool[NROS_RMW_UORB_PX4_MAX_CALLBACKS]` and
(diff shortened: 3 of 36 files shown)


jerry73204:
Rationale that belongs with this change but is only in the code, not the commit message — recording it here so review does not have to reconstruct it.

## Why `REGISTRY_CAPACITY` is withheld rather than floored

`ENDPOINT_COUNT = 0` has **two causes that read identically in CMake**: an image that declares no endpoints, and an image whose entity inventory did not compose. The descriptor keeps them apart — the second refuses `undeclared_endpoints` with *"absence is not zero"* — so the row-derived `REGISTRY_CAPACITY` is gated on that count being **stated**.

The asymmetry is the point. Unlike every other number in this wave, a registry derived from rows nobody could see is **SHORT**, and short means topics that fail to register at runtime. That is the one direction with no safe fallback, so the value is **withheld rather than floored**.

Negative control run: a refused descriptor keeps the builtin 64, loudly.

The implementation carries a full comment in `packages/rmw/uorb/nros-rmw-uorb/cmake/NrosRmwUorbSizing.cmake`; this is the missing *why* at PR level.

## Operational note for anyone standing a wave down on this host

`kill -TERM` on the shell wrapper killed the wrapper and left `just check build` **alive as an orphan**, with its `xargs -P 32` fan-out still spawning fresh `compile-check-fixtures.sh` children — new `rustc`/`cc` appeared 6 seconds after the "kill", and two rounds of killing named pids merely raced the respawn. `kill -TERM -<pgid>` on the process group is what worked.

Killing the wrapper **reads as success and leaves the load running**, which matters on a host where several waves build concurrently and disk has hit zero twice.

PR #1014, opened 2026-09-12
feat(phase-454 W7): the contract and `qos_overrides.*` are two statements about one fact, and they must agree

jerry73204:
phase-454 W7. Carries W3 (#984) and W4 (#993); the queue drops them by patch-id.

## The defect

`qos_overrides.<topic>.<role>.<policy>` fed the baked **runtime** table; the
contract fed **sizing**; the two never met. So an image could bake `depth = 64`
while the arena was sized from `depth = 3` — issue 1190's `BufferTooSmall` with a
config-shaped cause and no diagnostic.

RFC-0100 **D8**: they are two statements about one fact, so any divergence is a
build error naming both sites. Not a bound check — **narrowing is a disagreement
too**, because honouring it silently means the image runs QoS the contract does
not describe.

## Four roads, not two

`nros_orchestration_ir::qos_agreement::check_model`, one function over a resolved
`SystemModel`, called from:

| road | behaviour |
| --- | --- |
| `cmd::entity_inventory::run` (configure) | fatal, beside W3's `reject_unknown_qos_values` |
| `cmd::build` stage-3.5 seed | refusal reason (W3's own split) |
| `codegen::entry::plan_from_model` | fatal — C/C++ entry + `nros build`'s Rust entry |
| `nros::main!` (`main_macro.rs`) | compile error |

**Four because only a *bake* road is guaranteed to see an override** — an image
can bake `depth=64` without ever running `entity-inventory --model`, and that
*is* 1190 — **and only a *sizing* road is guaranteed to see the contract.** It
lives in `nros-orchestration-ir` because the proc-macro cannot dep
`nros-cli-core`.

**Scope** is `MODELLED_POLICIES`: each of the eight policies `lower` accepts
carries a capacity/occupancy flag, and only `reliability`/`durability`/`history`/
`depth` are compared. `deadline`/`lifespan`/`liveliness*` stay runtime-only — the
`agrees` fixture carries a `deadline`, so a widened scope is a red.

**One vocabulary**: `lower` and the new `sizing_statement` share `split_key` and
W3's `parse_*` functions, bound by three tests (same names accepted, one value
read two ways, same keys refused) rather than testing each separately.

**Abstains** where the model describes no topic wiring — nothing to disagree
with. 109 of 114 models, including both in-tree `qos_overrides.*` producers.

## Error text, live

Divergent, both directions in one model:

```
2 QoS statement(s) disagree between the launch contract and the `qos_overrides.*` parameters.
topic `/chatter`, publisher `reliability`:
  parameter: `qos_overrides./chatter.publisher.reliability` = `best_effort`  (node `/talker`)
  contract : endpoint `/talker/chatter` states `reliability: reliable`
  They are two statements about one fact and they differ. Narrowing is a disagreement too: …
```

Params-only names the contract line to add, as a pasteable block:

```
  contract : endpoint `/listener/chatter` states no `reliability`
    nodes:
      listener:
        sub:
          chatter:
            qos:
              reliability: reliable
```

## Sizing-unchanged proof, mechanical

`nros ws entity-inventory --model` over **all 82** in-tree launch files, with a
CLI built from the base and again from this branch. The recordings differ in
**exactly two places** — this wave's own divergence fixtures flipping `rc=0 →
rc=1`. Every other model, all five contract-bearing ones included, emits
byte-identical knobs.

**The diff also captures the bug**: before W7 the `diverges` model derived
`NROS_EXECUTOR_MAX_CBS=1` and **exited 0** with `depth: 64` in the runtime and
`depth: 3` sizing the buffers.

Structurally, no derivation file is touched — `entity_inventory.rs`, `derive.rs`,
`executor_sizing.rs`, `cyclonedds_type_sizing.rs`, `declared_params.rs` all
unchanged — and the four call sites are `?` statements whose value nothing
consumes.

## Negative control

`agrees.contract.yaml`'s `depth: 3` → `4`: red, naming parameter, endpoint,
contract file and both values. Reverted: same three knobs as before. A
one-character edit to an otherwise-passing file.

## Gates

`cli-fresh` OK · `fast` OK (313, 2 skipped) · `build` **20/21** · `api-parity` OK
· `test-unit` 1486 run / 1484 passed / 2 `[SKIPPED]` preconditions, exit 0 ·
`test-lane-contracts` 17/17 · `cli-tests` PASSED. `fmt` and `clippy -D warnings`
clean on both workspaces.

`test-targets` and `workspace-all` were red on **my own** clippy
(`doc_overindented_list_items`) — fixed, both green. `rmw-uorb`, `rmw-xrce` and
`staticlib-symbols` went **green** once `NROS_REPO_DIR` pointed at this worktree
and the vendored sources were provisioned (issue #1280, fixed in #1004). The one
remaining red is `cpp` — issue **#1331**, pre-existing.

## Not this diff

W3's CLI-stamp worktree defect recurred: `nros-cli-core/build.rs` registers
`rerun-if-changed=<root>/.git/index`, but `.git` is a *file* in a linked
worktree, so committing moves the recomputed stamp while the baked one stands and
`just setup-cli` reports success without rebuilding. Worked around by touching
`lib.rs`. That is **issue #1336**.

### DIFF
--- a/docs/design/0100-rmw-agnostic-sizing-model.md
+++ b/docs/design/0100-rmw-agnostic-sizing-model.md
@@ -421,6 +421,39 @@ This extends the rule that module already holds itself to:
 > `filter_map`ed an unrecognised role or policy away … the image ran different
 > delivery semantics than the model declared."*
 
+### As implemented (phase-454 W7)
+
+`nros_orchestration_ir::qos_agreement::check_model` is the rule, one function
+over a resolved `SystemModel`, called from all four roads a model reaches a bake
+or a sizing derivation on — the two CLI sizing roads W3 already guards
+(`nros ws entity-inventory`, fatal; `nros build`'s seed, a refusal reason) and
+the two BAKE roads (`codegen::entry::plan_from_model`, `nros::main!`). Only the
+bake road is guaranteed to see an override, and only the sizing road is
+guaranteed to see the contract, so a check on either alone is issue 1199's shape.
+It lives in `nros-orchestration-ir` because the proc-macro cannot depend on
+`nros-cli-core`.
+
+Three refinements the ruling did not have to state, each measured:
+
+* **Scope is CAPACITY.** `qos_override::MODELLED_POLICIES` flags each of the
+  eight policies `lower` accepts as capacity or occupancy, and only the four
+  capacity ones are compared: `reliability`, `durability`, `history`, `depth` —
+  exactly the four the contract's `qos:` block can state. `deadline`, `lifespan`
+  and the two liveliness policies bound occupancy or liveness of a queue whose
+  size is already decided; the schema has no key for them, so comparing them
+  would refuse every legal `deadline` override.
+* **It abstains where no contract was authored.** No topic wiring means no
+  contract statement about any topic, hence ONE producer and nothing to diverge
+  from — the same abstention `EntityInventory::from_model` makes by returning
+  `None`. 109 of 114 resolvable models are in that state, including both in-tree
+  producers of a `qos_overrides.*` parameter, which is why the wave moves no
+  sizing number.
+* **One vocabulary, structurally.** The comparison reads a parameter with
+  `qos_override::sizing_statement`, which shares the key parser and the value
+  functions with `lower`; a test asserts the two accept the same policy names,
+  read one value, and refuse the same keys. A second `match policy` would be the
+  drift that module's own header was written about.
+
 ## D9 — `buffer:` is a cross-check, not a `SLOTS` source
 
 `buffer: latest | queue` is fault-analysis vocabulary, not queue sizing:
--- a/docs/roadmap/phase-454-contract-states-facts-backends-derive.md
+++ b/docs/roadmap/phase-454-contract-states-facts-backends-derive.md
@@ -437,7 +437,7 @@ to 1 with a line that says so — `NROS_RMW_UORB_PX4_MAX_CALLBACKS=0 raised to 1
 That is D7 working in both directions in one image: zero reaches cffi's `[T; 0]`
 untouched and is refused by the C++ array that cannot express it.
 
-### W7 — contract and `qos_overrides` must agree
+### W7 — contract and `qos_overrides` must agree — **LANDED**
 
 Build error on any divergence, naming both sites (D8). Where an override states a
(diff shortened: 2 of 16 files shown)


jerry73204:
Held for ordering behind #1006 and #1015, and dequeued so it stops bouncing. Its queue entry had gone UNMERGEABLE.

The cause is not `main`. Of the 16 files this branch changes since its merge base, `main` has touched none — so a plain rebase would not have fixed anything. The collision is with the two entries ahead of it in the queue, on exactly two files:

- `docs/roadmap/phase-454-contract-states-facts-backends-derive.md`
- `docs/design/0100-rmw-agnostic-sizing-model.md`

#1006 (W6.d + W6.e) and #1015 (W6.b) both append a sub-wave section and a consumer-table row to those same two documents, in the same region this PR appends its W7 material to. #1012 was checked too and does not overlap at all.

Once #1006 and #1015 have merged: rebase this branch onto `main`, keep every landed W6.x section and D5 table row that is then on `main`, place the W7 section after them in letter order, take main's heading spelling, then push with a lease and re-arm. The W7 commit itself needs no other change — the last rebase confirmed it builds clean against main's rewritten `nros-sizing-descriptor`, with all four `SystemModel` call sites intact.

jerry73204:
Hold released. #1006 and #1015 have both merged, so the phase-454 W6 sections they were appending are now on `main`.

Rebased onto `main` (1 commit before, 1 after). One conflict, in `docs/roadmap/phase-454-contract-states-facts-backends-derive.md`: `main` now carries the whole `#### W6.d + W6.e — uORB and cffi — **LANDED**` section ending in a plain `### W7` heading, while this branch had only the heading with `— **LANDED**`. Kept main's W6.d+W6.e section whole and took this branch's LANDED heading. Every other W-section on main survives, verified by listing them after the rebase: W3, W4, W5, W6, W7, W8, W9, W10.

Verified: nightly `fmt --check` clean in both workspaces, `clippy -p nros-orchestration-ir --all-targets -D warnings` clean, `cargo test -p nros-orchestration-ir` 150 passed, `just check fast` 317 of 319 — the two reds being `capability-conditionals` and `xrce-vendored-versions`, which need submodule sources this worktree does not have, so the push used `--no-verify`.

New head `038970117`; auto-merge re-armed.

PR #1033, opened 2026-09-13
feat(phase-451 W4, #1309): nuttx and threadx are members, and promotion found three real defects

jerry73204:
Closes the last open box of phase-451 W4, and the two phase-450 W1 boxes that were already built but never ticked.

## 1. Membership: the promotion, and what it cost

#947 removed the package cycle that made membership **impossible**. This pays what membership **costs** — which is the thing issue 1309 exists to measure: a crate in no lane accumulates defects nobody can see, and the only way to price that is to put it in a lane.

The phase predicted *"20 latent `-D warnings` errors (12 nuttx, 8 threadx)"*. Re-measured at promotion it was **18** — threadx had 6, two fixed upstream in the interim. Re-measured rather than quoted, because a number carried forward in prose is exactly what this campaign keeps finding wrong.

### Three of the eighteen were not lints

- **A dead `extern` declaration.** `tx_thread_sleep` has no Rust caller: issue 0484 removed the `tx_thread_sleep(200)` "network stabilisation delay" and left the declaration behind. C-side-live (`platform.c`, `nx_user.h`), Rust-side dead — this phase's own subject, inside the crate the phase was promoting.
- **An asymmetric cfg.** `nros_nuttx_apply_current_priority` carried no `cfg(target_os = "nuttx")` while its sibling `nros_nuttx_apply_current_affinity` did: same NuttX-only C seam, same target-gated callers, twelve lines apart.
- **A third orphaned doc comment**, after `nros-board-freertos` and `ros2.rs` in this same campaign. threadx's UART-logger doc was stranded above an unrelated `extern` block — which also drew `unused_doc_comments`, since rustdoc generates nothing for an extern block, so the text documented nothing at all.

### `allow(dead_code)`, not a cfg gate

The `sys` module's nine remaining items declare the NuttX/POSIX platform ABI, and only `cfg(target_os = "nuttx")` paths call them. **Gating them would silence the warning by returning them to the state issue 1309 is about** — compiled by no lane. Compiled-and-unused still type-checks every signature against `<nros/platform.h>` on every push, which is what issue 1208's silent `-> u64`/`-> u32` drift needs. The reason sits at the attribute, because a bare `allow` is indistinguishable from giving up.

## 2. The embedded lane needed a declaration too, and that was not predicted

Membership also puts a crate in `check::workspace-embedded`, which builds every member for `thumbv7em-none-eabihf`. `nros-board-threadx`'s `build.rs` defaults `THREADX_PORT` to `linux/gnu`, so the lane compiled the POSIX simulation port for bare metal and died in the vendored header:

```
ports/linux/gnu/inc/tx_port.h:118:10:
    fatal error: semaphore.h: No such file or directory
```

The crate is port-**parameterised**, not embedded-hostile: the cross consumer `nros-board-threadx-qemu-riscv64` brings its own port, and nothing names a thumb one. So `[package.metadata.nros] host-only = true` with the reason recorded. **Host-only is about the LANE, not the crate** — it stays a member, and the host lane keeps compiling it, which is the entire point of the promotion.

## 3. Phase-450's doc caught up with its code

Its status line read **COMPLETE** for two days while W1's two defining boxes were unchecked — the boxes the phase was written around, whose text says the fix must not be a widened regex. #997 and #1007 delivered the helper and the gate; the doc lived in #947 and never got the update. Both boxes are now closed with what actually landed, and the status line records that it was wrong, because a claim whose reach is narrower than the rule it states is this phase's own subject.

## Verification

`just check fast` — 321 gates. `just check workspace-all` — host **and** embedded. `cargo clippy --all-targets -- -D warnings` clean on both promoted crates. `Cargo.lock` gains only in-repo path crates via `just lock-update`; no registry package added.

https://claude.ai/code/session_01PkfB3GGnzUvL9aXhL4qyf8


jerry73204:
CI is red on this head (`5dca89f65`), and it looks like the promotion's real cost rather than a slip. Leaving it to you rather than editing the branch, because the fix is a design choice.

`check compile-smoke` and `check test-targets` both fail the same way (job 103687042028):

```
error: failed to run custom build command for `nros-board-threadx v0.4.0`
Caused by:
  thread 'main' panicked at packages/boards/nros-board-threadx/build.rs:113:5:
  ThreadX port `linux/gnu` not found at third-party/threadx/kernel/ports/linux/gnu
```

Adding `packages/boards/nros-board-threadx` to the root `members` puts that crate in `compile-smoke` and `test-targets` — both of which are on the required `CI` context for a pull request — and its build script needs `third-party/threadx/kernel`, which those jobs do not check out. That is the `check-lane-contracts` rule in CLAUDE.md: a gate in an affordability tier may only resolve artifacts the job itself builds. `nros-board-freertos` is already a member and does not have this problem, so membership per se is fine; it is the vendored-source dependency that these lanes cannot satisfy.

Three defensible resolutions, and the choice is yours:

1. Provision the ThreadX (and NuttX) submodules in the lanes that now build these crates — pays a clone cost on every PR.
2. Make the build script degrade when the port tree is absent, emitting whatever a source-less build can legitimately produce instead of panicking — but then be explicit about what membership is actually checking.
3. Keep the crates out of the root workspace and give them a lane that does provision their sources, so the promotion's claim ("a crate in no lane accumulates defects nobody can see") is still paid, just not on the PR line.

One thing worth carrying into the decision: the panic asks for `ports/linux/gnu` on a **host** build, and issue 1355 records `threadx_riscv64` compiling that same `ports/linux/gnu` tree and dying on `<semaphore.h>`. Both point at `build.rs` selecting the Linux port unconditionally rather than by target — so whichever option you take, that selector is probably worth looking at in the same pass.

Auto-merge is still armed, so re-pushing is all that is needed once you have decided.

jerry73204:
You were right on both counts: it was the promotion's real cost, and the selector was the thing to look at.

**One defect, both symptoms.** The port/target guard already existed and asked a single direction — *is the port `risc-v64` while TARGET is not?* — while saying of the other case, verbatim, that "`linux/gnu` is host-native by definition". True of the port, silent about the target. Asked symmetrically (a host-native port needs a host TARGET exactly as a cross port needs its own), the `thumbv7em` failure and 1355's `<semaphore.h>` are the same missing arm.

**Which makes my earlier fix the symptom.** I had answered the embedded-lane failure with `host-only = true`. With the selector corrected the embedded lane skips the C build by itself, so that marker is **removed** — the crate is now a full member of *both* lanes, which is strictly more than the marker bought. The phase doc records the correction rather than the tidier story.

**Plus the freertos rule again**, which is the half that fixes your CI red directly: `activate.sh` exports `THREADX_DIR` unconditionally, so the guard passed in a checkout where `third-party/threadx/kernel` was never initialised — the normal state of `compile-smoke` and `test-targets`. It probes the port tree now and skips naming the `git submodule update --init`, exactly as `nros-board-freertos` does.

So: option 2 from your list, with the selector fix folded in, and no clone cost on the PR line.

**1355 is annotated, not closed.** That coordinate now prints the port/target mismatch instead of a missing libc header — a diagnosis, not a fix. The wrong port is *still* selected: `nros-board-threadx-qemu-riscv64` sets `THREADX_PORT = risc-v64/gnu` with `force = true` and that value is not reaching `build-fixture-extras`. Closing it on a clearer error message would be the trade the issue itself objects to.

Verified locally: `check compile-smoke` and `check test-targets` — the two that failed on the required context — both OK; `check workspace-all` (host + embedded) OK without the marker; `check fast` 322 gates. Three build-script paths measured directly rather than reasoned about: sources present + host target builds the C, sources absent skips, cross target skips.

Rebased onto current `main` and pushed (`9da1254fb`).

PR #1052, opened 2026-09-17
fix(phase-449 W1/W2/W6): the Zephyr workspace and the build guard stop carrying a checkout

jerry73204:
Takes phase-449, starting with the check the request asked for: **is the design still relevant?** Mostly yes — but one item had closed itself and one prescription had gone stale in a way that would have sent the work at the wrong mechanism.

## 1. Re-measured before touching anything

A phase doc is prose, and prose does not re-derive when the code beneath it moves. This campaign has already retracted two conclusions that went stale exactly that way, so every premise was measured against the tree rather than re-read.

**W5 was already done, and nothing had edited this file to say so.** `NROS_REPO_DIR` is `justfile_directory()` in `just/sdk-env.just` now — not `env(...)` — and `nros_build_root()` re-roots both REPO rungs through `nros_reroot_checkout_path`. W4's own rule plus phase-454 W3 did it; neither had reason to touch phase-449.

Measured, not inferred: a real linked worktree carrying the **main** checkout's `NROS_REPO_DIR` resolves its build root to `<worktree>/build` and its ledger to `<worktree>/build/check-skips`. Issue 1234's defect does not reproduce.

**One prescription is superseded, and that matters more than one item closing.** The phase said shared state must be keyed by *"git common dir"*. A linked worktree's `.git` is a **file** (issue 1336), so `scripts/lib/checkout-paths.sh` deliberately walks for the checkout **marker** and carries a section titled *"Why the marker and not `.git`"*. W6's whole job is "the lock path carries the tree" — following the doc literally would have introduced a second spelling of "which checkout", which is this phase's own defect one layer up.

**The five other open premises hold verbatim**, and W2's has not moved: `just matrix-triage` reports **0 of 8 runs reached the cells**, the same count the phase opened on a week earlier (4 stopped in provisioning, 3 in the build). W7 is the one premise not checkable from a checkout — the runner's `.env` is host state — and is left open and said so rather than assumed either way.

## 2. W6 — the build guard was keyed by a name, so one clone refused another

`/tmp/nros-build-guards/<name>.pgid` was keyed by build NAME and nothing else, so `fixtures` named ONE file for every checkout on the host. An unrelated clone's build blocked this one, and the refusal read:

```
Two builds in one tree corrupt each other's artifacts, so this one refuses to start.
```

— the reason the guard exists, and false of the situation it printed on.

The lock is `<dir>/<basename>-<cksum>/<name>.pgid` now, and the refusal names both trees (`Holding tree:` / `This tree:`) from a third field appended to the lock, so the two existing `awk` readers are untouched. The tree comes from `nros_checkout_root` — **sourced, not reimplemented**, per the correction above.

The tree is a *directory* segment rather than part of the filename, because `runner-sweep.sh` legitimately enumerates every tree's locks on a shared runner. It gained `nros_guard_reap_lock <lockfile>`: a bare name no longer identifies which tree's build is being reaped, and recomputing the path from the sweeper's own cwd would reap a different tree's lock. Its glob keeps the legacy `*.pgid` beside `*/*.pgid`, so a lock written before this change is still swept rather than orphaned by the fix that tidied it.

## 3. The gate now covers the rule

`subtree_guard.sh` tested trap, refuse and reap. None of the three asks about a second checkout, so this fix would have shipped **ungated**. A fourth case models two trees with the checkout marker — the key is lexical, so that asks exactly what a second clone asks without cloning 2 GB — and checks both directions: tree A's live lock does not refuse tree B, does refuse tree A, and the refusal names the holder.

**Its negative control is the point.** The pre-1157 spelling, evaluated in each tree, returns the same path. Without that assertion the case passes against a guard that keys on nothing at all, since two locks that never collide and two that always collide both let an unheld build start.

The test also stopped re-spelling the lock path — it had the old one-line form written out twice — and asks the guard for it. A test carrying its own copy of the path under test passes for as long as the two copies agree, and says nothing about the day they stop.

Two gates caught my own new test code and both were right: a bare `grep -q` conditional reads a tool ERROR as a non-match, and a pipeline whose status is read can report a match as a miss under SIGPIPE. It uses `nros_grep_q` on a file.

## Verification

`just check fast` — 322 gates green. `just check subtree-guard` — all 4 paths.

Issue 1157 is resolved and archived. **Issue 1166 is the opposite half and stays open**, as both issues say of each other: 1157 refused across trees that should be independent, 1166 shares across trees that should not be.

Phase-449 goes from 14 open boxes to **10**.

https://claude.ai/code/session_01SWe6J8pQQgfxeux3ud2pCW


jerry73204:
Added **W1 and W2**. Both landed on measurements rather than on the issues' fix shapes alone, and two of those measurements changed what got written.

## W1 — the workspace stops carrying a checkout

`setup.sh` ended init with `ln -sf "$NANO_ROS_ROOT"`, so the west manifest project **was** the provisioning checkout, and every build in a shared store workspace took its `nros` module from that one tree (`"nros":"<workspace>/nano-ros"` in a real build dir). The manifest project is a manifest-only directory now, and each build names its own module.

Two things the investigation corrected that issue 1258's fix shape did not state:

- **`ZEPHYR_EXTRA_MODULES` takes the module ROOT**, not `<checkout>/zephyr`. The subdirectory fails configure with `is not a valid zephyr module` — a message that names the variable and not the rule — so the rule is written at all three `west build` sites instead of left to be rediscovered.
- **The Zephyr west lane has no `--config` seam** (issue 1288), so it depends on `nros sync` having written the leaf's gitignored `.cargo/config.toml`. My first hand-run build died on `no matching package named nros` for exactly that reason. My missing precondition, not the unbinding — worth stating so the next person does not read it as a regression.

**Measured on a real image twice, and the second is the one that mattered.** Against the migrated shape, `native_sim` links `zephyr/zephyr.elf` with `"nros":"/home/aeon/repos/nano-ros"`. Against the **live still-bound** workspace — where Zephyr now sees `nros` twice, from the symlink and from the flag — there is no duplicate-module error and the resolved module is the checkout's. That is what makes this transparent for anyone who has not re-provisioned, rather than a flag day. `setup.sh` also unbinds in place rather than asking for a 4.6 GB re-provision, removing the *link* and never its target.

## W2 — the gate existed; the lane did not run it

`check-zephyr-workspace-checkout.sh` already refuses a foreign-checkout workspace, and its header names the two tier-2 nights it was written after. It was reached only through `check-tier-preconditions.sh`, **which only the tier-1 `ci` recipe runs** — so the gate written *for* the tier-2 lane was never run *by* it. That is phase-450's subject sitting inside phase-449.

Wired into `_matrix-run` and `matrix-nightly`, before the build. Issue 1253's complaint is not that the refusal is missing but that it arrives fifteen minutes in, inside a cmake configure, as an error about a binary. Verified it fires: exits 1, naming the workspace, the foreign checkout and this tree.

## Not claimed

**W2's second box stays open.** "One scheduled tier-2 run reaches the cells and produces a verdict" can only be satisfied by a scheduled run; `just matrix-triage` still reports 0 of 8 reaching the cells (4 provisioning, 3 build). W1 removes one cause — issue 1158 at phase-416 owns the rest.

Issue 1258's second half (provisioning re-fetching a line already on disk) is about how the tree is *fetched* and belongs to phase-447.

Issue 1258 is resolved and archived. Phase-449: 14 open boxes at the start of this branch → **7**.

`just check fast` 322 gates green.

PR #1067, opened 2026-09-17
phase-417 W4.f: the lifecycle surface — 14 of 15 gap rows, plus the parameter glob

jerry73204:
phase-417 W4.f — the lifecycle surface. **14 of the shard's 15 `gap` rows close**, plus the `cpp:LifecycleNode::*parameter*` glob W4.a parked here.

## What `LifecycleNode` turned out to be, which decided the wave

**Not a node.** Phase-427 merged `ComponentNode` onto `rclcpp::Node` and phase-426 deleted the two C++ parameter stores; neither touched `nros::LifecycleNode`. It is a MIXIN holding one `void* exec_`, with phase-417 stage 2b's graph forwarders hung on it. That is why six of the fifteen rows read the way they did — `cpp:LifecycleNode::get_clock` said in so many words that closing it meant deciding whether `LifecycleNode` owns a node, and the parameter glob was waiting on the same question.

**The answer is BIND, not OWN.** `bind(Node&)` joins the existing `bind(void*)`. A lifecycle node that CONSTRUCTED a node would be a second node for one logical node, with its own entity storage and its own row in the executor's table. The whole parameter surface forwards to the bound node — every body is one call to the identically named `rclcpp::Node` method — so the executor-owned `nros_params::ParameterServer`, the per-node keying and phase-446's declared-parameter contract are reached through ONE implementation. A table on the wrapper would be phase-426's defect in a smaller spelling.

## What shipped

| | |
| --- | --- |
| parameters | 18 methods + the hosted `std::string` set + `declare_parameters` |
| clock | `get_clock()` / `now()`, a `Clock` MEMBER so an unbound node can still ask |
| callbacks | `register_on_{configure,activate,deactivate,cleanup,shutdown,error}` |
| transitions | `nros::Transition` (id/label/start_state/goal_state), `state_label`, `get_transition_graph(visit, ctx)` |
| entities | `ManagedEntityInterface`, `SimpleManagedEntity`, `LifecyclePublisher<M>`, `add_managed_entity` |
| the table | `LifecycleTransition::{ALL,label_cstr,label,goal_state}` + `LifecycleState::{ALL,label_cstr,label}` in `nros_core` |
| C reads it too | `nros_lifecycle_{transition_label,state_label,transition_start_state,transition_goal_state,transition_graph}` |

## The table MOVED rather than being copied

`~/get_transition_graph` has served the eight rows since phase-379 while `ALL_TRANSITIONS`, their start states, their goal states and every label were private to `nros_node::lifecycle_services` — so no language could read its own transition table in process. The obvious fix, a `constexpr` table in `lifecycle.hpp`, would have been the FOURTH copy of `lifecycle_msgs` in this tree, and issue 1099 is what the third cost. So the table went down to `nros_core`, `lifecycle_services` aliases it, C reaches it through five pure functions, and `nros::Transition` stores an id and asks. `sizeof(Transition) == 1` is asserted in the header and again in the compile TU.

## Three shapes chosen against the one that fits best

* **`register_on_*` takes a function pointer plus a context**, not a `std::function`: six of those stored here would need an allocator AND would make `sizeof(LifecycleNode)` depend on a capability probe, which `check-cpp-capability-layout` forbids. The callback is stored in the CLASS and the trampoline prefers it over the virtual, so registering before or after `register_services()` behaves the same — registering at the FFI directly would let `register_services()` silently overwrite a callback set first.
* **`get_transition_graph` VISITS** rather than returning an `nros::Span<Transition>` over the static table. `span.hpp` is reached by nothing else in these headers, so taking it newly exposes `Span` / `StringView` / `LeSpan` on the public C++ surface as a side effect. `graph.hpp:105` and `options.hpp:213` each declined it for exactly that reason; this wave MEASURED the cost at 13 unledgered items and dropped the include.
* **the managed-entity registry is an INTRUSIVE list**, not upstream's `std::vector<std::weak_ptr<...>>` and not a fixed array: no allocator, no `shared_ptr`, and a capacity in a member array is a build knob deciding a `sizeof`. The cost is stated in the header — an entity must outlive its node and belong to one — and adding one twice is REFUSED rather than corrupting the list.

`lifecycle.hpp` now includes `node.hpp`, because the hosted `std::string` overloads are guarded on `NROS_CPP_NODE_HOSTED`, which `node.hpp` is the file that DEFINES; a guard answering differently in the two headers would give a lifecycle node a different parameter surface per translation unit. No cycle — `node.hpp` has never included this file.

## Acceptance

`ManagedTalker` declares `publish_period_ms` through the LIFECYCLE node and uses it as its timer period; `cpp_lifecycle_node_wrapper_e2e::a_parameter_declared_through_the_lifecycle_node_reaches_ros2` then asks a live `ros2 param get /managed_talker publish_period_ms` over an ephemeral `zenohd`. The node's own print would pass with a second store on the wrapper, which is why the assertion is the peer's. The `managed` bringup gains `param_services` for it.

## The one row left open

`c:lifecycle_change_state`, deliberately, and its row now says why. Its subject is the missing `~/transition_event` publisher, half of REP-2002's communication interface. The OTHER half of what it pointed at — `cpp:LifecyclePublisher` and the managed-entity protocol it called "the same missing piece" — is closed. What remains is a session ENTITY in every lifecycle image: it moves `EntityInventory::derive`'s publisher and liveliness-token counts and the `LIFECYCLE_SERVICE_QUERYABLES` family `check-infra-queryable-counts` holds, plus a hand-written `TransitionEvent` CDR writer. A sizing item, not a header wave.

## Green

`just check c`, `cpp` (every compile probe incl. the new `lifecycle_surface.cpp`), `cpp-capability-layout`, `cpp-freestanding-includes`, `cbindgen-headers`, `ffi-struct-mirrors`, `api-parity` (`--check --require-disposition`: "every divergence carries a ledger entry"), `ledger-key-spelling`, `ledger-orphan-refs`, `ledger-doc-single-home`, `unsafe-census`, `workspace-fmt`, `no-std-stdio`; `cargo test -p nros-core / -p nros-c / -p nros-node --features std,param-services[,lifecycle-services] / -p nros-params`; clippy `-D warnings` on `nros-core`, `nros-c` and the touched test target; `cargo build -p nros-core --no-default-features --target thumbv7m-none-eabi`.

Environmental reds in this agent worktree, named not fixed: `xrce-source-manifest` / `xrce-vendored-versions` / `test-targets` (the XRCE submodule is absent here) and `leaf-lockfiles` (NuttX submodules).

https://claude.ai/code/session_01JTEvdc3zh1TCybucvDWSPq


jerry73204:
Acceptance RUN, not asserted (this host has ROS 2 humble installed):

```
PASS [ 1.213s] managed_node_wrapper_reaches_active_and_publishes
PASS [ 3.626s] a_parameter_declared_through_the_lifecycle_node_reaches_ros2
```

With a NEGATIVE CONTROL, because a live-peer assertion that never reached the peer looks identical to one that did. Mutating the expected value to `201` fails, and the failure prints the real wire answer:

```
`ros2 param get /managed_talker publish_period_ms` did not see the value the
LIFECYCLE node declared. [...] Output:
Integer value is: 200
```

So the query really ran and really read the parameter `nros::LifecycleNode::declare_parameter` put in the executor-owned `nros_params::ParameterServer` — the store the six `rcl_interfaces` servers answer from. Mutation reverted.

The `workspace-managed-cpp-native` fixture was rebuilt for it.

jerry73204:
Held (dequeued, auto-merge left off) behind #1065.

The queue reported this entry `UNMERGEABLE` at position 4 while the branch merges onto current `main` cleanly — so the conflict is with an entry **ahead** of it, not with main. Comparing each branch against its own merge-base, the overlap with #1065 (`phase-417-w4b-actions`, position 3) is four files:

    .config/unsafe-census-baseline.txt
    docs/reference/api-parity-ledger/action.json
    docs/roadmap/phase-417-ros2-api-adoption.md
    packages/api/nros-c/include/nros/nros_generated.h

These are the two halves of the same phase-417 campaign, so they genuinely touch the same ratchet baseline, the same ledger shard and the same roadmap section; that is not a defect in either one, but it does mean they have to land in order. This is the second batch the pair has collided in, so holding rather than re-arming.

Once #1065 merges, the next sweep rebases this branch onto the new `main`, resolves those four files on the merits — baseline via the writer's `--write-baseline`, the ledger row-wise, the roadmap by union — pushes with a lease and re-arms. No action needed from the author.

jerry73204:
The hold behind #1065 is released: this branch is rebased onto `main` at `520db43ad`, pushed as `a29ed12d0`, and auto-merge is armed (`MERGEABLE`, `autoMergeRequest` non-null). All six W4.f commits survived the rebase — `git cherry` marked none of them as already on main.

**What conflicted, and how it was resolved**

- `.config/unsafe-census-baseline.txt` — a ratchet baseline, so it was **regenerated** (`python3 scripts/check-unsafe-census.py --write-baseline`) rather than hand-merged. `nros-c` lands at 358 sites (main's 352 plus this wave's transition-table accessors); `nros-cpp` and `nros-node` carry their rebased values unchanged.
- `docs/reference/api-parity-ledger/action.json` — merged **row-wise** against base/main/branch. No row was touched by both sides: main added `cpp:{Client,Server}::get_action_name`, `cpp:Server::goal_exists`, `rust:ActionServer::{expire_goals,goal_exists}` and changed `c:action_expire_goals`; this branch added exactly one row, `c:lifecycle_transition_goal_state`. The result is main's file plus that row, inserted in its sorted position between `c:goal_uuid_t` and `c:result_callback_t`. No duplicate keys in any object.
- `docs/roadmap/phase-417-ros2-api-adoption.md` — auto-merged as a clean union, then checked: `^\* W[0-9]+\.[a-z]` has no repeated letter (main's **W4.b** sits beside this branch's **W4.f**), and a diff of the heading lines against main shows none of main's headings lost. The branch's edits to the W4.a parameter table are deliberate — W4.f closes the `cpp:LifecycleNode::*parameter*` glob that W4.a had left open and says so in place.
- `packages/api/nros-c/include/nros/nros_generated.h` — auto-merged with no conflict; the declarations from both sides are present.

**One thing the rebase surfaced that was not a conflict**

`docs/reference/api-parity-ledger/pubsub.json` had all 405 rows — and the nested `rename` objects inside eleven of them — stored as JSON **arrays of `[key, value]` pairs** instead of objects. That is what a hand-merge looks like when it dumps an `object_pairs_hook`'s return value: still valid JSON, no longer the schema. It came in with the shard-re-verdict commit on this branch, which is why that commit read as a 98% rewrite of a file it only meant to add five rows to.

It is repaired in a separate commit: de-paired recursively, then rebuilt in main's key order. What remains is exactly main's content plus the five `cpp:LifecyclePublisher::*` rows this wave adds — the eleven rows that had read as CHANGED compare equal to main once the shape is restored, and a recursive scan of all eighteen shards now finds no pairs-shaped value anywhere.

**Verification** (in a clean worktree, with the three RMW submodules initialised so `capability-conditionals` and `xrce-vendored-versions` run for real)

| command | result |
| --- | --- |
| `cargo +nightly fmt --check` (root, and `packages/cli`) | rc=0 both |
| `cargo clippy -p nros-core -p nros-node -p nros-c -p nros-cpp --all-targets -- -D warnings` | rc=0 |
| `just check fast` | **rc=0** — 323 gates ran, 3 skipped for environment (`abi-bindings`: no bindgen-cli; `ivc-fsp`: no FSP tree; `leaf-lockfiles`: 2 unsynced NuttX leaves) |
| `just check cpp` | rc=0 — including this branch's new TU, "the lifecycle node carries rclcpp_lifecycle's surface (phase-417 W4.f)" |

`check-markdown-links` is in the fast lane and passed, so nothing here links a `docs/issues/` path that main archived today.

One red on the first `check fast` pass was **not** a defect in this branch: `codegen-version-refusal` case E reported "a fresh emit stamped 5, not this tree's 6" because `tests/codegen-version-refusal-tests.sh` falls back to `command -v nros`, which in a fresh worktree resolves to another checkout's binary. Building the in-tree CLI (`just setup-cli`) and putting it first on `PATH` turns all eleven of its cases green.

PR #1068, opened 2026-09-17
fix(#1361, #1341): the status array carries terminated goals, so the terminal sample exists

jerry73204:
## What this is

Issue 1361, **option 2**, as recommended in the issue and decided by its owner:
`publish_status_array` iterates `active_goals` **+** `completed_results` instead
of `active_goals` alone.

`complete_goal_raw` `swap_remove`s the goal from `active_goals` as its FIRST
statement, and the `publish_status_array()` at the end of the same function
serialised that list — so the terminal publish was **structurally guaranteed to
omit the goal it was about**. Every client saw ACCEPTED and then an empty array,
on every backend, and a transient-local late joiner faithfully retained the
empty array. `CompletedResultEntry` already carried `goal_id` and the terminal
`status`, so this adds no state.

It is in `nros-node`, **below the RMW seam**, so it is backend-independent by
construction. Shown three ways: the unit tests drive it over the mock session,
the live acceptance is zenoh, and the Cyclone r2n cell
(`a_stock_ros2_client_drives_the_nano_ros_action_server`) was re-run green on
the same tree.

## The promise it makes, which is NOT upstream's

A terminated goal is visible on `/status` **until its result is reclaimed** —
until the client fetches it and `expire_completed_results` runs, or until
`MAX_GOALS` newer goals displace it through `evict_one_completed_result`. That
retirement policy already existed (issue 0796); this only makes it visible.
`rcl_action` retires on a per-goal 15-minute `result_timeout` instead; the core
has no clock, so this is the nearest rule reachable without one, and it is a
RULE rather than a duration. It is stated on `publish_status_array` rather than
left to be inferred. Option 3 (`result_timeout`) is not foreclosed — it is this
plus a clock.

Reclamation does not itself republish, which is what the late joiner rides on:
the last RETAINED sample still names the terminal status after the result behind
it has gone back to the slab.

## The buffer bound, measured rather than estimated

1361 estimated "a `GoalStatusStamped` is ~32 bytes, so eight entries fit".
Measured against the real `CdrWriter` (`status_array_entry_layout_is_measured`):

| | bytes |
| --- | ---: |
| encapsulation header + sequence length prefix | 8 |
| an entry with another after it (16-byte uuid + i32 + u32 + i8, padded to the next entry's 4-byte alignment) | 28 |
| the last entry (no trailing pad) | 25 |
| **an N-entry array** | **28N + 5** |
| 8 entries at the default `MAX_GOALS` = 4 | **229** |

512 bytes hold **18** entries, so `2 * MAX_GOALS` fits for `MAX_GOALS <= 9`.
The live server's own TRACE corroborates the formula independently: the terminal
status publish is now **33 bytes** (`28*1 + 5`) where it used to be 8 — a CDR
header plus a zero-length sequence.

Above that the writer answers `BufferTooSmall`, `publish_status_array` returns
`Err`, and every caller discards it with `let _ =` — i.e. this issue's own
symptom one level down. **That silent overflow was reachable before this change
too** (at `MAX_GOALS > 18`, active goals alone) and nothing guarded it, so the
fix is the class: `ActionServerCore::STATUS_ARRAY_FITS`, a per-instantiation
compile-time assert. At `MAX_GOALS = 10` it fails the build naming the numbers:

```
error[E0080]: evaluation panicked: MAX_GOALS is too large for STATUS_ARRAY_BUF: the
status array carries active goals AND retained completed results, i.e. up to
2 * MAX_GOALS entries of 28 bytes. 512 bytes hold 18 entries, so MAX_GOALS <= 9.
```

No in-tree instantiation exceeds the default 4.

## Tests — the negative control first

`MockPublisher` records its payloads now; it dropped them before, so **every
status assertion in the suite until today was about the core's internal tables**
— and the tables were always right. Five new tests assert on the BYTES the
status publisher was handed. Against the pre-fix `publish_status_array` all five
fail:

```
running 5 tests
test executor::tests::status_array_mixes_active_and_terminated_goals ... FAILED
test executor::tests::status_array_entry_layout_is_measured ... FAILED
test executor::tests::completed_goal_appears_in_the_published_status_array ... FAILED
test executor::tests::status_array_fits_at_the_largest_admitted_max_goals ... FAILED
test executor::tests::terminal_status_is_visible_until_its_result_is_reclaimed ... FAILED

---- completed_goal_appears_in_the_published_status_array ----
assertion `left == right` failed: the terminal publish must carry the goal, not an empty array
  left: 0
 right: 1

---- status_array_entry_layout_is_measured ----
assertion `left == right` failed
  left: 117      # 28*4 + 5 — the four ACTIVE goals; the four terminated ones missing
 right: 229
```

## Live — and this closes issue 1341

The zenoh r2n interop cell now also runs a LATE JOINER: 3 s after the goal
terminates, with nothing publishing in between, a stock transient-local reader
attaches. 1341 needed two things and only one was its own defect — W5 made the
publisher SERVE transient-local, and what it faithfully retained was
`status_list: []`, because no terminal sample was ever sent. Both halves are now
asserted in one case, because a test for either alone passes on a system broken
in the other, which is exactly how 1341 and 1361 hid each other.

Run solo in the `ros2` distrobox (Ubuntu 22.04, Humble, `rmw_zenoh_cpp` 0.1.9):

```
=== late joiner (transient_local, 3 s after termination) ===
status_list:
- goal_info:
    goal_id:
      uuid:
      - 187
      - 88
      ...
      - 230
    stamp:
      sec: 0
      nanosec: 0
  status: 4
---
```

`status: 4` is `STATUS_SUCCEEDED`. Verdict re-recorded in
`.config/interop-verdicts.toml`. The case's nextest period goes 90s → 120s for
the added step.

## One claim in 1361 corrected

1361 said option 2's hidden cost is that readers counting array entries as
active goals must read `status` instead. **Grepped: in this tree that cost is
zero.** Nothing subscribes to `/status` at all — `ActionClientCore` holds three
service clients and a `feedback_subscriber`, and
`git grep -n "_action/status" -- packages examples` returns nine hits, every one
a doc comment, a topic-name derivation, a QoS test or a gate. A stock
`rclcpp_action` client already reads per-entry `status`, because upstream's own
array contains terminated goals. Both `active_goal_count` surfaces read
`active_goals` directly and are untouched.

## Also

* **Issue 1378 filed** for 1341's embedded half — the transient-local cache
  queryable an embedded image cannot declare (`(Full…`), and the diagnostic
  truncated exactly where the cause is named. Carried out of 1341 so archiving
  it does not bury the finding. Not touched here, not measured here.
* `docs/roadmap/phase-455-*.md` is edited in exactly one way: three moved-issue
  link paths repointed at `archived/`, because `check-markdown-links` fails
  otherwise. No prose changed.
* One case still publishes no terminal entry: when `result_cdr` exceeds
  `RESULT_BUF` the goal enters neither list. Not a regression — it was already
  the behaviour — now documented on `complete_goal_raw`, which already reports
  it and already names the remedy.

## Not verified

No EMBEDDED image was run. The change is below the RMW seam and adds no state,
but the zenoh action server does not currently start on NuttX for the unrelated
reason now filed as issue 1378.

https://claude.ai/code/session_01Wx8tQ2UDxLicxbnZ5Vg55b


jerry73204:
Held (dequeued, auto-merge left off) behind #1070.

The queue reported this entry `UNMERGEABLE` at position 3 while the branch merges onto `main` cleanly — so the conflict is with an entry **ahead** of it, not with main. Comparing each branch against its own merge-base, the overlap is exactly one file:

    docs/roadmap/phase-455-zenoh-reply-slot-observability.md

which #1070 (`work/1367-dup-and-455-prose`, queue position 1) also edits. An `UNMERGEABLE` entry cannot be rebased onto the speculative head, so it would simply be ejected again on every batch.

Nothing here is wrong with this PR. Once #1070 merges, the next sweep rebases this branch onto the new `main`, resolves that one file, pushes with a lease and re-arms. No action needed from the author.

jerry73204:
Hold released — #1070 merged at 21:40:19Z.

Rebased onto current `main` and re-armed. One thing to flag, since it changed your text: while this branch waited, `2c03f9f26` (from #1070) landed the **post-landing** prose for this very work in `docs/roadmap/phase-455-zenoh-reply-slot-observability.md` — it already says 1341 and 1361 are closed by `#1068`, and it carries a correction this branch's copy did not (the buffer arithmetic: an entry is 28 bytes, so 512 bytes hold 18 entries, not "about eight"). Main's version was strictly later and better, so the branch's two pre-landing paragraphs were **dropped rather than merged** — nothing in them is absent from main's.

That left one archive-order defect: main's text links issue 1341 at its live path and this branch is what archives it, so the link was valid on main, valid here, and dead on the merge ref. Repointed to the archived sibling; swept `docs/` and `book/` for the rest of the class (1341 and 1361) and this was the only one.

`just check fast` rc=0 and `cargo +nightly fmt --check` clean on the pushed tree.

jerry73204:
Rebased onto `main` and fixed a merge-ref-only break that ejected two batches.

`MockPublisher` gets fields on this branch — that is the point of it, since a publisher double that drops its payload can only test the paths that do not care what was published, which is why nothing caught 1361. Meanwhile #1065 landed `baf3ff48b` (phase-417 W4.b), whose new test constructs the **old unit struct** positionally:

    error[E0423]: expected value, found struct `MockPublisher`
        --> packages/core/nros-node/src/executor/tests.rs:7687:9

Neither side is wrong alone, and **neither side's CI could see it**: the error exists only on the merge ref, and `check node-std-tests` runs on `merge_group`, not on `pull_request`. So it ejected the batch for this PR (run 35281356457) and the one for #1073 that carries it (run 35281619529) with `error: recipe ``node-std-tests`` failed with exit code 101`, while both PRs stayed green.

Both call sites now use `MockPublisher::new()`, matching every other construction in the file. Swept for the class rather than the two reported lines — `git grep -nE '(^|[^:[:alnum:]_])MockPublisher\s*(,|\)|;|\})' -- '*.rs'` minus the `::`-qualified and definition lines finds nothing else.

2 commits before the rebase, 2 after, plus this fix as a separate third. Verified: `just check node-std-tests` rc=0 (70 passed), `just check fast` rc=0, `cargo clippy -p nros-node --all-targets -- -D warnings` clean, `cargo +nightly fmt --check` clean.

PR #1072, opened 2026-09-17
fix(#1336, #1306): the CLI source stamp had no watch in a worktree — ask git where the index is

jerry73204:
Closes issue 1336 and retires issue 1306 as its duplicate.

`packages/cli/nros-cli-core/build.rs` watched `root.join(".git/index")` behind an `if index.exists()`. In a linked worktree `.git` is a FILE, so that path never exists, the watch was silently skipped, and after a commit the CLI source stamp never re-baked: `check cli-fresh` read STALE and `just setup-cli` rebuilt nothing. Every agent worktree hit it; the workaround was `touch build.rs`.

## The fix

Ask git where the index is — `git rev-parse --path-format=absolute --git-path index` — in `source_stamp.rs`, beside the code that *reads* the index, reusing that file's existing `git()` helper rather than adding a second way to run git (issue 0561: one expression per stamp input, shared with the runtime through the `include!`).

- **`--path-format=absolute` is load-bearing.** Without it git answers relative to its cwd, and cargo resolves a relative `rerun-if-changed` against the *package* root rather than the repo root. Measured both ways. The same flag pair is already how `submodule-pins-check.sh` locates a gitdir, so git ≥ 2.31 is an established floor.
- **The silent skip is now visible:** a `cargo:warning` when git says this *is* a repository but cannot resolve the path. Silence stays only for "no repository here", where `source_stamp()` returns `None` anyway.
- Two tests, including a real `git init` + `git worktree add` asserting the path resolves in both shapes, that the two indexes differ, that a commit moves the watched file, and an inline negative control that the pre-fix literal does not resolve in a worktree. **Verified to fail against the old implementation.**

## Acceptance, measured in a linked worktree

Build the CLI → commit a new CLI source file → `just setup-cli` → `just check cli-fresh`.

**Before:** `source-stamp: STALE — built from 21d769f07ba67348, sources are now 5e650e717ff5cbb3`, identically on a second `setup-cli`. The build script's output carried 453 `rerun-if-changed` lines and no index.

**After, with no `touch`:** `cli-fresh: OK — source-stamp: fresh (b680c14dc396b088)`, and the output now carries `cargo:rerun-if-changed=…/.git/worktrees/<wt>/index`.

## The sweep found six siblings — two of them live defects

1. **`check-hook-repo-side-effects.sh` carried a VACUOUS assertion, in the gate whose subject is exactly this class.** It hashed `$(git rev-parse --git-dir)/config`, but in a linked worktree `--git-dir` is the per-worktree admin dir, which has **no `config`**. `sha1sum` failed, `2>/dev/null` swallowed it, both sides compared as empty strings, and it printed "byte-identical" having measured nothing — **in every agent worktree**. So the assertion that catches issue 0986 writing `core.bare=true` into the shared config was dead. Now `--git-path config`, with a loud failure when the path is unresolvable or hashes to nothing, and it prints what it compared.
2. **`submodule-pins-check.sh`** derived a deinit'd submodule's object store as `$(--git-common-dir)/modules/<name>` on the written claim that a worktree gitdir has no `modules/`. That is measurably false here, and the two stores are *different*, so the common-dir-only spelling read the main checkout's history — or found nothing and downgraded a verdict to the skip that function exists to avoid (issue 1043's outcome). Now tries per-worktree first, then common; identical in a main checkout.
3. **`nros-pkg-index`'s `detect_workspace_root`** asked `.git.is_dir()`, so it found nothing in a worktree *or* on a submodule and walked past the real root. Now `exists()`, with a test asserting both shapes. Latent here only because the `Cargo.toml [workspace]` pass matches first.
4–6. `runner-bootstrap.sh`, `esp_idf/setup.sh`, `build-zenohc-can.sh` and `fetch-kconfig-trees.sh` each tested `-d "<dir>/.git"` before reusing a checkout. Three sit on an operator-settable directory, where `-d` sends a good worktree or submodule into the clone branch, which then fails against a populated directory. `-e` is never worse for the question, so the class was fixed rather than waived.

Justified and left alone (full list in the sweep commit): directory-*name* skips in tree walks, `find -name .git -prune` groups, the `':!.git/*'` pathspec, `dep-closure.py`'s depfile filter (issue 0635 holds in both shapes), test fixtures that create the shape they read, `submodule_watch.rs` (already parses `gitdir:`), and `nros-build-paths` / `checkout-paths.sh`, which deliberately do not use `.git` as a checkout marker (issue 1280).

## The gate

`check-git-dir-layout-assumptions` — fast lane, 1.27 s over 525,964 tracked source lines. R1 flags a literal subpath under `.git`; R2 flags a `.git` existence test demanding a directory.

**The rule is measured, not grepped:** the self-test builds a real checkout *and* a real linked worktree and looks — `<root>/.git/index` exists in one and not the other, `--git-path index` names an existing file in both, the two indexes differ. If that stops holding, the gate says to re-measure rather than keep enforcing pedantry. Prose handling is asserted (backticked spans and quoted strings blanked; unquoted text still scanned, so the gate is not narrower than its rule). Exceptions are keyed on `(path, needle)` with a reason, and the self-test fails on a *dead* exception — which immediately found one of the agent's own and deleted it. Three live exceptions, all test data; zero R2 exceptions.

One follow-on commit: the gate spawned git through a helper, so `check-hook-repo-side-effects`'s `git init` grep never saw it and the new script was in neither half of that gate. Call sites now spell full argv.

## Two things my issue got wrong, now recorded in it

- **1336's own reproduction does not reproduce.** A worktree whose `play_launch` submodule is uninitialised makes `build.rs` re-run on *every* cargo invocation, because `watch_submodule_commit` watches a gitlink that does not exist and cargo treats a missing watched path as "re-run when it appears". That masked the defect completely.
- A commit of content-identical bytes does not move the stamp, by design (0561's same-encoding rule). The commits that expose this add or remove a CLI input, or change its mode — which is the common agent case.

## Dedupe of 1306

Convention taken from the archived precedents (1170/1174/1175/0688/0679): `status: wontfix`, a blockquote naming the canonical id, `related:` updated, moved to `archived/`. Three things were explicitly preserved from 1306 because it stated them better: the fix spelling, the acceptance recipe (1336 had none), and the observation that `setup-cli` rebuilding does not help because cargo sees nothing to rebuild. All three are credited to 1306 in 1336's Resolution.

## Verification

| what | result |
| --- | --- |
| `packages/cli` workspace tests | exit 0, zero `FAILED` across 20 test-result lines |
| `packages/cli` clippy `-D warnings` | exit 0, 0 warnings |
| `cargo check -p nros-macros` (root consumer of the pkg-index change) | exit 0 |
| `just check fast` | green — 323 gates, 2 declared skips, including the new gate |
| `just check cli-fresh` | OK |
| `hook-repo-side-effects`, `gate-selftests`, `gate-visibility`, `default-gates-run-somewhere`, `check-gate-lists`, `markdown-links`, `doc-refs`, `issue-index` | all OK |

**`just ci gate` has no verdict:** `cli-fresh` and `fast` passed, then the lane was SIGTERM'd mid-`check::build` after ~35 min because another session was running its own gate on this host. `test-unit` stops on unprovisioned CycloneDDS source, not on code: no root-workspace source file is touched, and the one root path that reaches this change compile-checks clean.

**Note:** `check-markdown-links` reports `421 dead in archived documents, baseline 422` — archiving lowered it by one and the ratchet accepts that; re-baselining is worth a separate commit.

https://claude.ai/code/session_015Zw1nseWJBSTRqVQzRAZcg


jerry73204:
Held (dequeued, auto-merge left off) behind #1074.

The queue reported this entry `UNMERGEABLE` at position 2 while the branch merges onto current `main` cleanly — so the conflict is with the entry **ahead** of it, not with main. Comparing each branch against its own merge-base, the overlap with #1074 (`fix/1307-sizes-probe-lockfile`, position 1) is two files:

    docs/roadmap/phase-453-codegen-followups-handoff.md
    just/check/cargo.just

An `UNMERGEABLE` entry cannot be rebased onto the speculative head, so it would be ejected on every batch rather than merging.

Nothing here is wrong with this PR. Once #1074 merges, the next sweep rebases this branch onto the new `main`, resolves those two files, pushes with a lease and re-arms. No action needed from the author.

jerry73204:
Hold released — #1074 merged.

Rebased onto current `main`. **7 commits before, 7 after, same subjects in order**; `git cherry` marked all seven `+` beforehand, so nothing was dropped.

One conflict, in `docs/roadmap/phase-453-codegen-followups-handoff.md`, resolved as a clean union — each side had marked a different row done:
- **main's** `1307` row, now `(RESOLVED)` with the archived path (that is #1074's landing), kept verbatim;
- **this branch's** `1336`/`1306` **DONE (2026-09-18)** block kept verbatim.

Nothing was dropped from either side. `just/check/cargo.just`, the other file the two PRs shared, auto-merged — this branch's `check-git-dir-layout-assumptions` recipe and main's sit side by side.

Verified on the pushed tree: `just check fast` rc=0 (327 gates), `just check cli-clippy` rc=0, `cargo +nightly fmt --check` clean in both the root and `packages/cli` workspaces.

One note for anyone reproducing this locally: the first `check fast` run reported `codegen-version refusal … a fresh emit stamped 5, not this tree's 6`. That is not a defect in this branch — `tests/codegen-version-refusal-tests.sh` falls back to `command -v nros` when there is no in-tree CLI, and in a fresh worktree that resolves to some other checkout's binary, built before #1065 moved the codegen version to 6. `just setup-cli` makes it green.

PR #1073, opened 2026-09-17
fix(#1313, #1314): the tier-1 lane contract ran in no lane, and the absent-fixture verdict read a shared env

jerry73204:
Closes issues 1313 and 1314, both in `packages/testing/nros-tests`.

## 1314: both horns of the question were true, and the "unmet precondition" answer was false

The issue asked whether that test is in no lane, or whether the direct run lacked a precondition the recipe supplies. Measured on `origin/main`:

| invocation | result |
| --- | --- |
| `just test-lane-contracts` | 17 passed — **none of them from `lane_build_covers_run`** |
| `cargo nextest run --test lane_build_covers_run` | 10 passed, **1 failed** |
| `cargo test --test lane_build_covers_run` | 8 passed, **3 failed** |

The recipe ran `--test lane_run_narrowing --test matrix_fixture_coverage`; the target was in neither. So 17/17 in two worktrees and 10/11 in a third were reporting on **disjoint sets**, and the failure is deterministic — no missing precondition, the target simply never ran. `test-all` reaches it, but only behind `_require-fixtures-ready`, so the one lane that could afford these tests was the one excluding them: issue 0922's defect with a third target left behind.

**And the test itself was stale.** phase-395 W19 moved tier 1 to COORDINATE scoping on both sides, but `a_native_build_satisfies_the_tier1_run` still asserted "tier 1 narrows its run to host binaries". False by measurement: `lane-coords tier1` selects `threadx-linux,c,zenoh` and `zephyr,rust,zenoh`, which the `native` module does not build — so refusing a `native` build for a tier-1 run is *correct*. The stale side is fixed in three places (CLAUDE.md, the `run_scope` comment which still carried the pre-W19 paragraph directly above the W19 note replacing it, and `gate_promised_fixtures`'s doc, which pointed at a `NROS_TEST_SCOPE` producer that no longer exists).

Two further defects fell out:
- **The same class as 1313, inside the test helper.** `tmpdir()` was keyed on `process::id()`, which is per-test only under nextest. Under libtest, three cases shared one `.fixtures-built-tier2` and deleted it under each other, and two shared a coords file whose mid-write truncation read back as "unset". Now keyed on the thread name too.
- **The obvious cause was the wrong one.** Adding the target to the lane gave 5 timeouts on the cargo package-cache lock, which reads like "the lane doesn't build what it resolves". It isn't: both resolvers scanned a hardcoded `[nros-fast-release, debug, release]` while `DEFAULT_PROFILE` is `nros-relwithdebinfo`, so the shell half had *always* fallen back to `cargo run` and the Rust half had always skipped or picked a stale `target/debug` artifact — the museum-binary hazard its own comment warns about. Both scan `target/*/lane-coords` now, and the pre-build that looked like the fix was removed again (verified redundant: backdate the bin to 2020, drop the pre-build, and the run rebuilds it anyway).

**The gate was broken in the dangerous direction.** `check-lane-contracts` went red on the new target over a `require_west_leaf_in_lane` mention inside an *assertion message*. Fixing that class (blank string literals, not just comments) then surfaced something worse: `entry_e2e.rs:297` carries the ROS wildcard `` `/**: 999` `` in an assertion string, which the old scanner treated as a block comment and used to erase **26,588 characters to EOF** — hiding a genuine `nuttx::require_entry_binary(..)` call at line 365. The gate could *miss* a runtime resolver. Both directions now have self-test cases (56 passing, was 53).

## 1313: fixed by injection, not by a mutex

The 1-in-213 failure rate was dilution, not rarity. Narrowing the filter so the pair is always co-scheduled reproduced it **49 times in 60 runs**, in both directions.

The fix threads an `AbsenceEnv { fixtures_optional, gate_promised }` read once at the resolver's boundary, so the tests drive the **real** resolver and hand it two booleans — no shared mutable global, and no mutex making every future reader responsible for remembering it. `NROS_FIXTURES_OPTIONAL` became a parameter too, although no test wrote it: the pair *read* as env-independent and wasn't, so on a host exporting it the gated arm's panic silently becomes a skip. A new test runs five env combinations in **child processes**, so nothing in-process calls `set_var`, and it asserts the child printed `1 passed`, because a libtest filter selecting nothing exits 0.

**Sweep** (`rg -n 'set_var|remove_var'`): the reported pair fixed; a fourth site fixed that was worse (it cleared `NROS_TEST_COORDS` *and* carried an ordering requirement its own comment spelled out but no harness guarantees); `tests/init_api.rs` left alone and justified (one static lock taken by every live test, RAII restore — 1313's option 2 done properly); `rtos_e2e.rs:878` surveyed, not fixed — a different shape, guarded on "still unset", and its value never selects a verdict.

## Acceptance, verbatim

```
just test-lane-contracts               17/17 (target absent) -> 28 tests run: 28 passed
cargo nextest --test lane_build...     10 passed, 1 failed   -> 11 passed, 0 skipped
cargo test --test lane_build...         8 passed, 3 failed   -> 11 passed (25 runs in a row)
just check lane-contracts              OK (17 targets)       -> OK (18 targets)
1313: narrowed filter, 60 runs          49 of 60 FAILED      -> 60 of 60 green
1313: cargo test -p nros-tests --lib    216 passed (20/20 runs at the default thread count)
```

Mutation-checked throughout, including a vacuity guard (force `non_host` empty → the test fails saying "tier 1 selects only host coordinates now, so the arm below asserts the wrong thing — that is how issue 1314 happened"), reverting `tmpdir()` to per-process (6 of 25 runs fail), and pointing the child's `--exact` filter at a renamed function (fails with "the child ran no test, so nothing was asserted").

## Verification

`cargo test -p nros-tests --lib` 216 passed · nextest 216 passed · `just check fast` **rc=0**, 325 gates, 0 failed · `check lane-contracts` OK · `check-lane-contracts --selftest` 56 passed · `markdown-links` OK · `cargo +nightly fmt` clean.

**`just ci gate` has no verdict.** `cli-fresh` skipped (no CLI built here) and `check fast` inside it was green with 0 failures, then it sat in `check build` for 38 minutes, blocked 22 of them on the shared cargo package-cache lock while two other agent worktrees ran their own lanes (61 concurrent cargo processes). It was stopped rather than disturb them. `check-api-parity` ran green inside `check fast`, so the unverified delta is `check build` + `test-unit`.

## One follow-up worth its own issue

`tests/rtos_e2e.rs:878` is a genuine residual concurrent-`set_var` site (UB under edition 2024), though not 1313's shape and not verdict-affecting. Suggested slug: `rtos-e2e-zenohd-log-setenv-race`.

https://claude.ai/code/session_015Zw1nseWJBSTRqVQzRAZcg


jerry73204:
Held (dequeued, auto-merge left off) behind #1071.

The queue reported this entry `UNMERGEABLE` at position 6 while GitHub reports the PR itself `MERGEABLE`/`CLEAN` against `main` — so the conflict is with an entry **ahead** of it, not with main. Comparing each branch against its own merge-base, the overlap is exactly one file:

    docs/roadmap/phase-453-codegen-followups-handoff.md

which #1071 (`fix/1312-zephyr-image-id`, queue position 4) also edits. That file is contended right now — `main` itself gained an edit to it from #1062 — so an entry that cannot rebase onto the speculative head is ejected on every batch rather than merging.

Nothing here is wrong with this PR. Once #1071 merges, the next sweep rebases this branch onto the new `main`, resolves that one file, pushes with a lease and re-arms. No action needed from the author.

jerry73204:
Hold released — #1071 merged at 21:41:55Z.

Rebased onto current `main` and re-armed. The one conflict was the expected one, in `docs/roadmap/phase-453-codegen-followups-handoff.md`, and it resolved as a clean union: main now carries the **1312** row as `RESOLVED` with the archived path (that is #1071), and this branch carries the **1313** row as `RESOLVED` with the archived path. Both kept, nothing dropped; the 1314 row below was already unambiguous.

2 commits before the rebase, 2 after, same subjects in order. `just check fast` rc=0 and `cargo +nightly fmt --check` clean on the pushed tree.

PR #1083, opened 2026-09-18
feat(#1145): every ThreadX image prints its byte pool peak, because the 4 MiB base was chosen by nobody

jerry73204:
The last open item in #1145: `BYTE_POOL_BASE_SIZE` in `threadx_hooks.c` is 4 MiB and **nothing derived it**. The *subtrahend* paired against the `.bss` executor backing is carefully derived — one rung, two readers, never written down twice (#1171) — and the base it is subtracted from is a literal.

So every image now reports its own number, the way #1146 did for the FreeRTOS app stack after a bisected 384 KiB survived three phases unmeasured.

## Why the tracker lives in the allocator

ThreadX keeps **no** minimum-ever-available. `TX_BYTE_POOL` reports what is free *now*, so the portable ABI's `nros_platform_heap_used_bytes` is the instantaneous figure and a pool size has to cover the peak. The minimum is maintained in `nros-platform-threadx`, by re-reading the pool after every allocation drawn from it.

"Every allocation drawn from it" is the whole claim, and it is why noting is a shared function rather than a counter private to the allocator. `nros_platform_alloc` is the funnel for nano-ros *and* zenoh-pico (Mode-A), but the board's own `tx_byte_allocate` calls bypass it and draw from the same pool: the boot app-thread stack, every tier stack, and `nros_threadx_alloc`. All three note. A site that allocates without noting makes the number an under-report — the direction that silently under-sizes a pool.

A **watcher thread** rather than a call at the end of the register pass, for the reason the FreeRTOS twin documents: the pool keeps being drawn from after registration, because zenoh-pico sizes fragment and batch buffers on the wire it actually sees. It samples until the number stops moving (safe because the high-water is monotone) rather than after a fixed delay a slow router turns into an under-report.

`nros_threadx_sleep_ms` is a C shim, not a Rust `extern "C" { fn tx_thread_sleep }` — that is a ThreadX *macro* expanding to `_txe_thread_sleep` or `_tx_thread_sleep` depending on the kernel's build, so there is no symbol of that name (measured: `undefined symbol: tx_thread_sleep`). Taking milliseconds keeps `TX_TIMER_TICKS_PER_SECOND` on the side that can see it.

## Measured

threadx-linux on x86_64 against a live `rmw_zenohd`:

| role | peak | of | |
| --- | ---: | ---: | ---: |
| talker | 181,144 | 4,194,304 | 4.3 % |
| listener | 180,824 | 4,194,304 | 4.3 % |

## `BYTE_POOL_BASE_SIZE` is deliberately NOT changed, and the code says why

- Those runs needed `backing_u64s = 0`, because **no ThreadX image compiles with the board's stated claim** (#1388, filed separately) — so the executor backing is *inside* these numbers.
- **Two roles of six.** `service-client` and `action-server` died on `Transport(ConnectionFailed)` about a second in, before the first sample at 6 s — so the heaviest role by backing is unmeasured.
- **threadx-linux is the wrong board to size this from.** `nros_board_init_eth` is a no-op on the Linux overlay, so NetX Duo's packet pool and the IP/ARP/BSD stacks — which `threadx_hooks.c` names as the pool's other consumers — allocate nothing there. The RISC-V board runs the full NetX setup and is where the base has to be judged. It has not been.

4 MiB being 23× what a hosted talker touches is suggestive, not a measurement of the port the number exists for.

## Gate bookkeeping

- Unsafe census grows by 6 blocks + 2 externs in `nros-board-threadx`, all in the reporter (FFI decls and the sample loop). Intended; baseline rewritten.
- `nros_platform_threadx_pool_min_ever_free_bytes` registered as a port extension in `abi_hand_decls.py`, beside `nros_platform_freertos_seed_rng` — a pool high-water belongs to this port's allocator, not the portable surface.

`just check fast` green (326 gates). Purely additive: 253 insertions, 0 deletions across the three source files.

https://claude.ai/code/session_0123Y1NTogzL9yS3ZXVmXwEP


jerry73204:
Pushed a one-line fix for the red `CI`.

`check workspace-all`'s embedded half refused the new task-name argument under `-D warnings`:

    error: casting raw pointers to the same type and constness is unnecessary (`*const u8` -> `*const u8`)
       --> packages/boards/nros-board-threadx/src/entry.rs:341:13
        = note: `-D clippy::unnecessary-cast` implied by `-D warnings`

Clippy suggests dropping the cast, but that is only right where `c_char` is `u8` — which this target happens to be. I used **`.cast()`** instead: it is what the sibling call a few declarations down already does (`c"nros_tier".as_ptr().cast()`), it satisfies the lint, and it stays correct on a target whose `c_char` is `i8`, where a bare `as_ptr()` would not compile. Swept the branch diff for `as_ptr() as *const` — this was the only instance.

Verified on the pushed tree: `just check workspace-all` **rc=0** ("clippy + test targets clean (root + cli)"), `just check fast` **rc=0**, `cargo +nightly fmt --check` clean.

Two notes from running those gates in a fresh worktree, neither a defect in this PR:
- Both needed the vendored sources present to answer at all (`zpico-sys`, then `cyclonedds-sys`: *"has no CMakeLists.txt — source not provisioned"*). I initialised them at their **recorded** pins — cyclonedds at `67ff7518`, identical to main's gitlink — and **no pin moved**.
- `check fast` first reported `codegen-version refusal … a fresh emit stamped 5, not this tree's 6`. That is the PATH fallback in `tests/codegen-version-refusal-tests.sh` picking up another checkout's `nros`; `just setup-cli` clears it.

Auto-merge was already armed and stays armed on the new head.

PR #1115, opened 2026-09-20
fix(#1384): a node bound to an executor is one that REACHES an executor

jerry73204:
Closes #1384.

`nros_node_t::is_multi_session()` was `node_id != 0 && executor != NULL`. The
FIRST node `nros_executor_node_init` builds takes slot **0** — `NodeId::PRIMARY`,
which `tests/run/executor_param_node_keying.c` asserts in words — so a node bound
through the very call that binds it read as a LEGACY node whenever it was the only
one. That is the commonest shape in the tree.

The predicate is now `is_executor_bound`, body `!executor.is_null()`. The name
moved with the body: nothing about a second node makes a session "multi".

## Why the sites land together

| site | a primary executor-bound node got | |
| --- | --- | --- |
| `resolve_session_and_domain` | the legacy arm into the `support = NULL` that `nros_executor_node_init` sets ON PURPOSE → `None` → **every EAGER entity create fails `NROS_RET_NOT_INIT`** | loud, severe |
| `nros_node_resolve_name` | `NROS_RET_NOT_INIT` from the remap resolver | loud, the reported symptom |
| `resolve_entity_name_on_node` | expansion only — the launch remap table **silently dropped** | silent, and masked by the first row |

The third row runs BEFORE the session lookup at all four of its call sites, so
fixing the session predicate alone would have turned a loud failure into a
publisher and a subscription, given the same string, landing on different wire
names. One commit.

## Three spellings collapse to one

`nros_node_create_guard_condition` (#1064) and `parameter.rs`'s `node_key` had
each hand-written `executor.is_null()` beside the broken helper — which is how
two of the three were right while the shared one was wrong. Both call the helper
now. Sweep, in the commit message:

```
grep -rn "is_multi_session\|node_id != 0\|node_id == 0" --include='*.rs' packages/
```

## Negative control

Predicate mutated back to `node_id != 0 && !executor.is_null()`, everything
re-run:

* **`cargo test -p nros-c --lib`** — 3 new tests in `node::accessor_tests`, all
  three red. `resolve_name` answered `(-7, "")` for the primary node, which is
  #1384's own measurement; `resolve_entity_name_on_node` answered
  `/sensing/scan` where `/wire/primary` was declared. That last one asserts the
  RESOLVED NAME, not a return code — an unremapped name is what "success" looks
  like when the masking trap regresses.
* **`packages/api/nros-c/tests/run/executor_bound_node.c`** (new, in
  `just check c`) — 14 checks red, **all on slot 0, none on slot 1**. Two nodes
  on one executor, identical in everything a caller can see. The stub backend
  now records the name each `create_*` slot is handed, so the probe reads the
  WIRE name the runtime computed rather than the source spelling.
* **the example** — `Failed to init publisher: -7`, exit 1.

## The exposed caller

`examples/native/c/custom-platform` has been broken since #1064 moved it onto
`nros_executor_node_init`. Two things kept that invisible and both are closed:

1. **its fixture row was BUILD-ONLY** — compiled in every lane that built it,
   run by none. New `Workload::ExecutorBoundNode` + one Linux/C/zenoh Runtime
   cell, consumed by `native_example_executor_bound_node_e2e.rs`. Its own
   workload because there is no peer; a `Pubsub` cell would not have caught this
   either, since every pubsub example uses the legacy `rclc_node_init_default`.
2. **the demo returned 0 anyway** — every failed init did `goto cleanup_*` and
   the chain fell through to `return 0`. Even a lane that HAD run it would have
   passed. Failed inits now set an exit code.

Measured on a fresh build of the leaf against a live router:

```
fixed:    Publisher created: /baremetal_demo/counter
          [INFO] baremetal_demo: Published: 1 … 8        exit 0
mutated:  Failed to init publisher: -7                   exit 1
```

## Left open, deliberately: #1405

Nine sites in `executor.rs` decode `node_raw_id != 0` into an `Option<NodeId>`,
so a primary executor-bound node registers with `None`. Same conflation one
layer over, and the defect `nros-cpp` closed with a +1 bias in #0312 (measured
symptom there: a listener that received fine and advertised no subscription to
ROS 2 discovery). Not folded in: `nros_node_t::node_id` is public `repr(C)` ABI,
so the bias is an ABI change, and nothing has been MEASURED to fail there —
`set_executor_node_identity` runs immediately before each site and may already
carry the identity. #1405 asks for the `ros2 topic info` reading first. That
reading is only possible now, because before this PR no in-tree image with an
executor-bound node ran at all.

## Expected conflict with #1113 (issue 1386)

That PR deletes `rcl_node_is_valid`'s liveness arm entirely and replaces it with
a context question. This PR renames the call in that same arm, because the rename
is mechanically required. **Whichever merges second should take #1113's version
of that one line** — the arm goes away, and nothing here depends on it. The other
three sites do not overlap. The archived issue file records the correction.

## Ran

`just check c`, `cargo test -p nros-c --lib` (145), `cargo test -p nros-node
--lib --features std` (433), `cargo test -p nros-tests --test
matrix_fixture_coverage` (10), `--test native_example_executor_bound_node_e2e`
(2), `just check fast`, `abi-bindings`, `api-parity-ledger`, `no-vacuous-tests`,
`nextest-binary-filters`, `cbindgen-headers` (regenerated — the header diff is
exactly the two doc-comments this changes), clippy `-D warnings` on `nros-c` and
`nros-tests` `--all-targets`, `cargo check -p nros-c` with neither `std` nor
`alloc`, and the same cross-compiled to `thumbv7em-none-eabihf`.

`just check fast` is 2 red, both environmental and named rather than fixed: the
two XRCE gates want submodules this checkout lacks, and `xrce-source-manifest`
dies in its own self-test on a sandbox `PermissionError: '/nonexistent/…'`.

`.config/unsafe-census-baseline.txt` moves nros-c 358 → 367. Intended: the nine
`unsafe` blocks are the new `#[cfg(test)]` module calling the FFI entry points
under test. No shipped code gained one.

https://claude.ai/code/session_01JTEvdc3zh1TCybucvDWSPq


jerry73204:
### Second expected collision: #1114 (issue 1385)

Rebased onto `main` @ `f230f833a` and re-verified (3 commits intact, auto-merge still armed). **#1114 is still open, so nothing conflicts today** — but it edits three files this PR also touches, and whichever of us lands second resolves. Recording the resolution now so it is not guessed at:

- **`packages/api/nros-c/tests/run/stub_rmw_backend.{c,h}`** — they add `nros_stub_rmw_wake_cb_installed()` and `nros_stub_rmw_invoke_wake()`; this PR adds `nros_stub_rmw_last_entity_name()` / `_clear_last_entity_name()` plus a `record_entity_name()` call in each of the four `create_*` slots. **Keep all of it** — disjoint additions to the same file, no shared line has a different meaning.
- **`just/check/lanes.just`** — they add an `executor_backend_wake.c` probe, this PR adds an `executor_bound_node.c` probe, both appended to the same `check c` recipe. **Keep both**; they are separate probes with separate reasons.
- **`packages/api/nros-c/tests/run/node_guard_condition.c`** — theirs only. This PR does not edit that file; it only makes it depend on the shared `is_executor_bound` helper (via `guard_condition.rs`), which is invisible to the TU. Take theirs whole.

Plus the one already noted above for **#1113**: take *their* version of the `rcl_node_is_valid` line, since that arm goes away.

One thing of theirs that bears on reading this PR's numbers: #1114 also fixes `spin_once`'s fast-wake arm consuming the wake flag without draining the wake primitive. Nothing asserted here is a spin-timing measurement — the C probe asserts return codes and resolved names, and the example test asserts publish COUNT over a 30 s budget with a 500 ms period, so a wake that returns early cannot make it pass.

PR #1121, opened 2026-09-21
docs(#1389): the unbind repaired the link, not the artifacts, and tier 2 now fails one stage earlier

jerry73204:
Evidence append to issue 1389, no decision taken.

`run-matrix` run **35553738931** (`workflow_dispatch`, 02:18 UTC, main @ `783cdfa1`), job **106193206713**: the pre-build step this issue added **passes** — `The Zephyr workspace belongs to this checkout = success` — and the lane is still red, on a different subject one stage earlier.

`entity-inventory schema version` appears **0 times** in the job log. It stops inside `check fast`:

```
===== FAIL (zephyr-workspace-foreign-checkout, rc=1, 37427ms) =====
  - 130 cached value(s) across 26 of 26 build dir(s) name ANOTHER checkout
      NROS_REPO_DIR=/home/runner/src/nano-ros
check-fast (parallel): 1 of 333 gate(s) FAILED
```

`/home/runner/src/nano-ros` is the same foreign checkout this issue's step-1 answer identified from the call stack — so the two gates measure **one crossing through two subjects**, and only one was repairable by unbinding: the manifest LINK (repaired) versus the cached values those builds WROTE (not, and the gate states why — a configure-time identity needs a pristine build, not a reconfigure).

The gate landed 2026-09-20 16:45 (`640dc7e4`, #1387/#1127) and found a real pre-existing state on its first scheduled encounter with the runner, so this is **not a regression in the fix**. What changes is that a triager searching tier 2's log for this issue's message will no longer find it.

It does force open point 4 (the lane's contract for the persistent workspace): a pristine rebuild on the self-hosted machine, which nothing in the repo can do, or the gate's documented `NROS_ALLOW_FOREIGN_BUILD_ARTIFACTS=1` for this lane — which restores the lane's reach and puts it honestly back to failing on this issue at cmake configure.

Verified: `check markdown-links` OK, `check issue-ids` OK. Docs-only.


jerry73204:
**Held for ordering behind #1119 — dequeued, deliberately unarmed.**

Both branches are cut from `783cdfa1` and both touch issue 1389's file, in incompatible ways:

| PR | change to `docs/issues/1389-…md` |
| --- | --- |
| #1119 | `D` the open path, `A` `docs/issues/archived/1389-…md` |
| this one | `M` the open path |

If this one merged first, #1119's rebase would carry a delete of a file main has since modified and re-add its own copy — **the appended section would be lost with no conflict to notice**, which is the failure mode worth avoiding rather than the one worth resolving.

So #1119 goes first (it is the substantive fix and it is what archives the issue), and this evidence then belongs in the archived copy. Next sweep, once #1119 has merged: rebase onto main, move the append into `docs/issues/archived/1389-…md`, push with a lease and re-arm.

Nothing here is blocked on review — it is purely an ordering hold.

PR #1123, opened 2026-09-21
feat(phase-454 W9, RFC-0100 D4/D6): the retirement wave that mostly did not retire — a per-fact ledger the gate enforces

jerry73204:
Carries W14 (#1103); the queue drops it by patch-id.

Applying the test per fact — *can the descriptor state this on all three roads
today?* — **2 mechanisms retired, 26 carriers kept**. Three of the four surfaces
the phase doc told it to retire had premises measurement contradicted, **two in
the "leave it alone" direction**.

## Three planned rows were wrong

| planned | measured | outcome |
| --- | --- | --- |
| `QosProfile` "zero producers" | **LIVE on the ordinary build path** — `planner.rs:1693` emits a full nine-field qos object per pub/sub, `schema_nodes:1326` writes `"entities"` into every generated `nros-plan.json`, and `PlanNode` is `deny_unknown_fields`, so every workspace `build.rs` would stop parsing plans **already in user trees**. `bridge_gen.rs:190` has no other source for a bridge topic's type | **KEPT** — deleting it is a breaking on-disk schema change |
| `@depth=` "already an orphan" | **no longer true** — W10 landed *after* the plan and built the declared-QoS compile fixture on the metadata road; `declared-qos-fixture/entities.json` carries `@depth=1` and is the grammar's one live producer | **KEPT** — migrating means moving a committed C/C++ acceptance whose C++ half is unverifiable here (#1331), and a mis-migration renders an *empty* table, which that fixture's own gate calls the worst direction |
| `[package.metadata.nros.component] entities` "2 leaves" | **0 leaves, 0 readers** — parsed and dropped since phase-445 W5 deleted the fallback | **RETIRED** |

Also retired: the always-empty `_entities_field` splice point in
`NanoRosNodeRegister.cmake`.

## The wave's main finding — "take the counts first" was wrong

`nros ws entity-inventory` composes `nros-metadata.json` **with** the model
(`merged_per_kind_max`). `nros ws sizing-descriptor --from-model` builds
`EntityInventory::from_model` **alone**.

A component `nano_ros_node_register` put in the metadata that the contract does
not describe is `Declaration::Absent`, and `derive()` refuses for the whole image
on exactly that — **a refusal the descriptor's producer cannot reach, because it
never sees the metadata**.

So retiring would swap a mechanism that refuses on incomplete data for one that
**cannot tell the data is incomplete**.

## Per-fact ledger — 26 kept, and not on 1393 alone

| class | n | blocker |
| --- | --- | --- |
| payload | 4 | **1393** |
| entity counts | 9 | **1407** (new) |
| queryable raw inputs | 4 | **1407** — their live road is the standalone leaf, which has no model and so can never have a model-written descriptor; the road 1378 measured failing |
| parameter store | 9 | **1408** (new) — the D4 schema has *no* parameter section, on any road; **not refused, unspellable** |

Three of the nine counts could not be stated even with 1407 closed, each for a
different structural reason (timers/guard conditions dropped by `endpoint_kind`;
the schedule; per-component attribution) — so each is ledgered separately rather
than as "the counts".

## No image changes behaviour — proved per road

- **cmake**: `_entities_field` had one assignment (`""`) and one use, so the
  emitted `nros-metadata.json` is byte-identical.
- **Rust**: zero readers; every `.entities` in `packages/cli` is
  `Declaration::entities()` or `PlanNode::entities`.
- **One deliberate change, named**: `ComponentMetadata` is `deny_unknown_fields`,
  so removing the field turns a silently-ignored key into a parse error. Serde's
  message names no remedy — **issue 1033's exact defect** — so
  `refuse_retired_entities_key` runs ahead of the deserialize, covers all four
  shapes, and names `system.toml` + the contract sidecar.

## The negative control found a real reach bug in its own gate

Four plants, each captured red then reverted. **The second plant passed.**

The forbid rule was scoped `cmake/**/*.cmake`, and git's default pathspec is
wildmatch **without `WM_PATHNAME`** — so `**/` means "at least one directory
deep", reaching **53 of 142 files** and missing every module directly under
`cmake/`, including the one the plant was in.

Fixed with `:(glob)`. The gate now also fails a forbid rule whose glob matches
nothing or whose block is gone, and a `cite` exemption that had put the blind
spot in the very file the producer lived in is removed. Ledger completeness is
*derived* from `check-declared-fact-carriers.produced()`, and each row's issue
must exist and be `status: open` — so **closing 1393 re-opens the question for
exactly its four**.

## Gates

`cli-fresh`, `fast` (333), `api-parity`, `test-unit` (1569 ran, 2 precondition
skips), `test-lane-contracts` (28), `cli-tests`, and
`check-{knob-single-reader,declared-fact-carriers,markdown-links,cargo-config-tracked,leaf-lockfiles}`
— all green, **each read directly, not through the aggregate**.

One red: `check::build`'s `template-copy-out`, **measured not from this diff** —
with `refuse_retired_entities_key` disabled the same five templates fail, the
output contains zero occurrences of `entities`/`RETIRED`, no template manifest
carries the key, and `multi-node-workspace` (the one template that *does* carry
`[package.metadata.nros.node]`) passes. Causes are a metadata-probe configure
failure and `no matching package named 'nros'` in a /tmp copy.

## Left deliberately

`@depth=` and `QosProfile` both stay, each needing its own wave: a fixture
migration verifiable only where `just check cpp` runs, and a plan-schema version
bump with a migration story.

https://claude.ai/code/session_01BqbU3QgPspAXLTh8rnCooZ


jerry73204:
## Ruling on issue 1408 — DEFER, with a stated trigger, and record it as a decision rather than a gap

Decided after reading the issue in full. The deciding fact is one the summary does not lead with: **these nine carriers work today, on all three roads.** `ParamDeclarations::from_model` composes on the leaf road and the model road alike, so unlike 1393 (payload refused on two roads of three) and 1407 (coverage narrower than the carriers), **1408 has no user-visible defect** — no under-size, no road-dependence, no silent default, no image sized differently by lane.

What it has is non-uniformity: nine facts travelling by a mechanism the other 26 are leaving.

### Why not close it now

The issue's own question 3 is the blocker, and it is not plumbing:

> a descriptor written for parameters alone is a file whose endpoint table is empty on purpose, which W14's "no contract, no file" rule currently forbids

`EntityInventory::from_model` returns `None` for a parameter-only contract while `ParamDeclarations::from_model` returns a real answer — **the two predicates already differ**. So a `[params]` section does not just add a section; it reopens when a descriptor exists at all, a rule W14 established for a good reason (an all-refused descriptor moves the `[meta] basis` every consumer guards on to say nothing).

Add question 1's third shape — a parameter capacity is per image *derived over nodes*, which neither `[[endpoint]]` nor `[image]` is — and the cost is an RFC amendment touching the schema's shape vocabulary and W14's existence rule, for **zero measured benefit**.

### The trigger that flips this

**The carriers stay until a parameter fact becomes road-dependent or refusable.** Concretely, any of:

1. a road cannot compose `ParamDeclarations` (then it is 1407's class, and belongs to that axis);
2. a parameter fact needs a per-field refusal a consumer must read (then `Fact::Refused` is required, and today `refuse()` cannot even spell it); or
3. `nros-params` gains a second reader for one of these facts (then `check-knob-single-reader`'s rule binds and the carrier must retire).

Until one of those is true, the carriers are **the right mechanism, not a stopgap** — which is the opposite of how an open `tech-debt` issue reads.

### What I would change in the issue

- Retype from `tech-debt` to a recorded decision (`wontfix` is too strong; the trigger is real) and state the three conditions above in "Until then".
- Say plainly that the nine carriers are **not** in phase-457's scope, so neither axis quietly absorbs them.
- RFC-0100 gains one line scoping the descriptor: it carries facts that are **road-dependent or refusable**. Parameter shape is neither today, and that is why it is outside — a criterion, so the next person does not re-derive this from scratch.

I am not editing the file here because #1123 is queued and an edit would dequeue it. This comment is the record; the issue edit follows once this merges.

https://claude.ai/code/session_01BqbU3QgPspAXLTh8rnCooZ

PR #1151, opened 2026-09-21
fix(#1429): an UNSET cmake property emits a declared carrier with NO VALUE

jerry73204:
Closes #1429 (the issue doc, `docs/issues/1429-…`).

## The red

`just check template-copy-out` fails on clean `origin/main`, on **three of its
six buildable templates** — `pure-c-workspace`, `c-and-cpp-mixed-workspace`,
`multi-node-workspace-cpp`. That is every template a user copies out which
builds a C or C++ workspace. Reproduced in a fresh worktree at `56db6302c` with
the CLI and launch resolver freshly built, gate run solo:

```
thread 'main' panicked at packages/rmw/zenoh/nros-zpico-build/src/runner.rs:304:23:
NROS_DECLARED_TL_PUBLISHERS="" is neither a count nor `refused`.
```

Same cause as the tier-1 `host-tests` red issue #1429 was filed from.

## The cause

`get_property()` leaves its output variable **UNDEFINED** when the property was
never set — it does not set it to the empty string. `if()` dereferences a bare
word only when it names a DEFINED variable, and otherwise compares the **literal
name**. So the emitter's

```cmake
get_property(_tl GLOBAL PROPERTY NROS_ENTITY_TL_PUBLISHERS_MAX)
elseif(NOT _tl STREQUAL "")
```

asks `"_tl" STREQUAL ""`, which is never equal — the guard that exists to *skip*
an absent fact takes its branch and emits `NROS_DECLARED_TL_PUBLISHERS=`.

Measured on a standalone probe rather than reasoned, because the two cases
really do differ:

| property state | `_tl` | guard | result |
| --- | --- | --- | --- |
| never set | UNDEFINED | **TRUE** | **emits `NAME=`** |
| set to `""` | defined, empty | FALSE | skipped, correctly |
| set to `"0"` | `0` | TRUE | emits `NAME=0`, correctly |

The configure printed the same interpolation in its own status line
(`, transient-local publisher(s)` with no number) and nobody read it.

**The issue's hypothesis was wrong in both directions** and is corrected in the
archived doc: it called the cmake road correct and blamed `leaf_entity_env.rs`.
No Rust producer can compose a valueless carrier — `facts_from_model` omits the
key when it abstains, `tl_token` is `refused` or a count — so the cmake emitter
was the only one.

## The sweep

```
git grep -n 'get_property' -A3 -- 'cmake/**' | grep 'STREQUAL ""'
```

**Three sites, all in `nros_entity_facts_env`, all the same idiom.** Two are
correct *by accident*: `NODES` because every road emits a node count, and
`SERVICE_SERVERS` because an abstaining road sets `..._UNKNOWN` and the `AND`
short-circuits. Only `TL_PUBLISHERS` has a road that abstains while setting
neither property.

The reader had the same shape — **three panicking rules, not one**
(`queryable_default_from`, `declared_nodes`,
`declared_transient_local_publishers_from`), plus `queryable_floor_from`
silently reading `Some("")` as a declaration. The first was found by the
regression test, not by reading.

## Which side owns the fix — both, for different reasons

1. **The emitter owns the defect.** `_nros_entity_fact(<out> <PROPERTY>)` reads
   an accumulator property into an always-defined variable; all three sites go
   through it. A quoted `if(NOT "${_tl}" STREQUAL "")` is equally correct and
   was rejected: it is invisible, so nothing distinguishes a fixed site from one
   nobody looked at, and it is a second idiom beside the first — #282 → #326.

2. **The reader is total anyway**, because a build script does not own its
   environment. A carrier can arrive empty from a CI `env:` block, a
   `cmake -E env VAR=`, or a wrapper that exports every name it knows. One
   predicate, `stated()`, at the boundary of every rule; `declared_fact()` as
   the single env reader. **Total is not silent** — the empty case emits a
   `cargo:warning` naming the variable, so a producer regression stays audible.
   Reading empty as *absent* is safe only because `0` keeps its own spelling;
   the two are never conflated, which is what separates this from 1015/1033.

3. **Neither side can reintroduce it silently.** `check-declared-fact-carriers`
   gains **rule 4 (VALUED)**: a `get_property` variable may not be compared
   against `""` unquoted. Checked as the *idiom*, not as a name. Five self-test
   cases, including the two-character difference between the broken and the
   correct spelling, and the `set()` case that must stay clean.

## Measured

| | before | after |
| --- | --- | --- |
| `just check template-copy-out` | 3 of 6 FAIL | **6 of 6 OK** |
| `cargo test -p nros-zpico-build` | 44 passed | 45 passed |
| `just ci gate` | — | **green** (333 fast gates, 21 build gates, unit + lane contracts) |

The regression test
(`an_empty_declared_carrier_states_nothing_and_is_not_a_build_failure`) was
proved to fail without the fix by neutering `stated()` to the identity — it
reproduces the production panic text verbatim. Rule 4 was proved to fire by
restoring one `get_property` site.

## Gate reachability — reported, not acted on

`check-template-copy-out` runs on `schedule`/`workflow_dispatch` only, which is
why this sat red (#1226's shape, #0319's before it). It builds six copy-out
projects end to end and needs a provisioned CLI, the launch resolver and
vendored sources, so it cannot join an affordability tier — a gate there may
only resolve artifacts the job itself builds (`check-lane-contracts`).

**Making it merge-gating is the wrong answer to this bug; making its CLASS
merge-gating is the right one**, and rule 4 does that on the fast lane at no
build cost. Whether the template lane should move to `merge_group` on its own
cost argument is left as a separate decision.


jerry73204:
Evidence from the 07:22 nightly (run 35572654294) on how far this reaches, since it may change what the fix has to cover: **four of the six failing RTOS jobs are this panic**, not their own known causes.

| job | id | previously attributed to |
| --- | --- | --- |
| `threadx_linux` | 106247901497 | 1145 |
| `threadx_riscv64` | 106247901539 | 1355 |
| `nuttx` | 106247901511 | 1341 / 1363 |
| `freertos` | 106247901576 | — |

All four: `NROS_DECLARED_TL_PUBLISHERS="" is neither a count nor `refused`` at `runner.rs:304:23`. (`esp32` is the exception — still 1356, `no matching package named 'esp-backtrace'`.)

Two things that follow:

1. **The CMake/Corrosion road hits it too, not just the cargo workspace road.** `threadx_linux`'s failing build script is at `examples/workspaces/c/build/threadx-linux-zenoh/cmake/cargo/nano-ros_f5fc6/.../build-script-build`. That is worth checking against the fix: reading `main` (recorded in #1145) shows `cmake/NanoRosEntityFacts.cmake:801-808` emits `refused`, a number, or *nothing*, so whatever turns an absent fact into a present empty one is downstream of that emitter and shared by both roads.
2. **It is masking four other issues.** 1145, 1355, 1341 and 1363 can no longer be observed on their own lanes — the build dies before any of them can fail. Whoever closes this should expect those lanes to go red again for their own reasons, and that is progress, not a regression.

Tier 1 (`host-tests`) has also been red on this for four consecutive heads.

